PR #56, opened 2019-10-21
make all patches idempotent. Fixes #2844

c0d1ngm0nk3y:
## Description

Make patches idempotent, so that the initContainers can be run several times without failing.

See https://github.com/SUSE/scf/pull/2845

## Test plan

Installed scf via cf-operator and checked that all pods are started.


bosh/releases/pre_render_scripts/diego-cell/cflinuxfs3-rootfs-setup/jobs/patch_pre-start.sh
```diff
@@ -17,3 +17,11 @@ patch --verbose "${target}" <<'EOT'
...
  CA_DIR=$ROOTFS_DIR/usr/local/share/ca-certificates/
 EOT
+)
+
+# Only patch once
+if ! patch --reverse --dry-run -f "${target}" <<<"$PATCH" 2>&1  >/dev/null ; then
+  patch --verbose "${target}" <<<"$PATCH"
+else
+  echo "Patch already applied. Skipping"
+fi
```
f0rmiga: [nit] missing a line break.

c0d1ngm0nk3y: Done. It would be great to get this finally merged. Just let us know if there is still something open.

PR #58, opened 2019-10-22
Support additional cidrs for k8s clusters

kramerul:
## Description

Currently cf on k8s doesn't support clusters created by [kops](https://github.com/kubernetes/kops) or [gardener](https://gardener.cloud/), because they uses different cidrs (100.64.0.0/10). See also [Dedicated space for carrier-grade NAT deployment](https://en.wikipedia.org/wiki/Private_network#Dedicated_space_for_carrier-grade_NAT_deployment)

## Test plan

Manual test

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/singleton-blobstore.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/singleton-blobstore.yaml
@@ -6,6 +6,9 @@
   value: 102400 # 100GB
 
 - type: replace
+  path: /instance_groups/name=singleton-blobstore/jobs/name=blobstore/properties/blobstore/internal_access_rules?
+  value: [ "allow 10.0.0.0/8;","allow 172.16.0.0/12;", "allow 192.168.0.0/16;" , "allow 100.64.0.0/10;"]
+- type: replace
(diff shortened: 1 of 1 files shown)


f0rmiga:
@kramerul These values can be passed through the generic property mechanism with Helm. How do you feel about it?

```yaml
properties:
  singleton-blobstore:
    blobstore:
      blobstore:
        internal_access_rules:
        - "allow 10.0.0.0/8;"
        - "allow 172.16.0.0/12;"
        - "allow 192.168.0.0/16;"
        - "allow 100.64.0.0/10;"
```

kramerul:
@f0rmiga, this is a good idea I haven't thought about.
On the other hand, the root cause is not easy to find for someone how has no deep insights in this software. There is no clear error message like "You need to configure additional CIDRs" when. I don't remember the exact error message, but it took us hours to find out, that we have to add one additional CIDR to the blobstore. 

I think another impact of the missing CIDR was that build packs were missing. It is quite difficult to deduce from missing build packs a missing CIDR.

The additional CIDR 100.64.0.0/10 is well defined like the other 3. 

Therefore, I would argue, that it should be part of an ops file and not part of a document.

f0rmiga:
@kramerul Makes sense.

PR #59, opened 2019-10-22
Bump Eirini to 0.0.21

kramerul:
## Description

Using latest eirini release.

## Test plan

Install kubecf with latest cf-operator. Run `cf push` afterwards.

## Problems

`cf push` is not working. But this is related to https://github.com/cloudfoundry-incubator/cf-operator/issues/648

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/eirini.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/eirini.yaml
@@ -105,7 +105,7 @@
           certs_secret_name: eirini-staging-secret
           cc_internal_api: https://cloud-controller-ng.service.cf.internal:9023
           eirini_address: https://eirini.service.cf.internal:8484
-          downloader_image: "eirini/recipe-downloader:0.2.0"
+          downloader_image: "eirini/recipe-downloader:0.3.0"
           uploader_image: "eirini/recipe-uploader:0.3.0"
           executor_image: "eirini/recipe-executor:0.3.0"
           # TODO: make this configurable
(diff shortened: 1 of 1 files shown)


deploy/helm/scf/assets/operations/instance_groups/eirini.yaml
```diff
@@ -106,6 +117,15 @@
           cc_cert: ((cc_bridge_tps.certificate))
           cc_key: ((cc_bridge_tps.private_key))
           cc_ca: ((service_cf_internal_ca.certificate))
+        quarks:
+            ports:
+            - name: http
+              protocol: TCP
+              internal: 8085
```
phil9909:
```suggestion
```

deploy/helm/scf/assets/operations/instance_groups/eirini.yaml
```diff
@@ -106,6 +117,15 @@
           cc_cert: ((cc_bridge_tps.certificate))
           cc_key: ((cc_bridge_tps.private_key))
           cc_ca: ((service_cf_internal_ca.certificate))
+        quarks:
+            ports:
+            - name: http
+              protocol: TCP
```
phil9909:
```suggestion
```

deploy/helm/scf/assets/operations/instance_groups/eirini.yaml
```diff
@@ -106,6 +117,15 @@
           cc_cert: ((cc_bridge_tps.certificate))
           cc_key: ((cc_bridge_tps.private_key))
           cc_ca: ((service_cf_internal_ca.certificate))
+        quarks:
+            ports:
+            - name: http
```
phil9909: This (the http port) should not be needed anymore.
```suggestion
```

f0rmiga:
@kramerul The bump to 0.0.21 seems good. I tested with cf-operator on commit https://github.com/cloudfoundry-incubator/cf-operator/commit/c193b5b616ef22a3a65489e4ff573109480a5c62, which is part of https://github.com/cloudfoundry-incubator/cf-operator/pull/661, and the failure I get is due to the `opi-task-downloader` init container not being able to connect to blobstore.service.cf.internal, which makes perfect sense given that it is in a different namespace and it's not using our aliases on the CF namespace. I pushed a commit on the PR while it was on the SCF side that fixed this. I'll accept this PR as is and I'll push a fix for that in a separate PR.

PR #67, opened 2019-10-24
Submoduled cf-operator

f0rmiga:
## Description

Pinning cf-operator on a specific commit that is known to work with kubecf.

## Test plan

Verify that using the pinned cf-operator works with kubecf.


f0rmiga:
Instead of submoduling or referencing a pre-built binary, we will reference a published chart (which indirectly references a binary).

PR #70, opened 2019-10-25
Added deployment + smoke-tests + cleanup stages to CI

f0rmiga:
## Description

Refactored Gitlab manifests and scripts a bit to accommodate the deployment + smoke-tests + cleanup steps. Also, removed the global `scf` namespace. Bazel targets that depended on this namespace now expect the `NAMESPACE` environment variable to be passed.

## Test plan

Smoke tests should pass on CI.


.gitlab/pipelines/stages/deploy/deploy_cf_operator.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -o errexit -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/config/config.sh"
+
+# Install cf-operator.
+chart="$(mktemp -t cf_operator_XXXXXXXX.tgz)"
+curl -L -o "${chart}" "${CF_OPERATOR_CHART_URL}"
+bazel run @helm//:helm -- template "${chart}" --name cf-operator --namespace "${KUBECF_NAMESPACE}" | bazel run @kubectl//:kubectl -- apply -f - --namespace "${KUBECF_NAMESPACE}"
```
mook-as: This sounds like it would make sense as a bazel command (as in, `bazel run @operator//:deploy` or something).

.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/config/config.sh"
+
+# Locate the built kubecf chart.
+chart="$(find output/ -name 'scf-*.tgz')"
```
mook-as: Hmm. Not sure I like finding _any_ file with vaguely the right name.  But then I don't like my alternative either:

```bash
bazel aquery //deploy/helm/scf:chart 2>&1 | grep 'Outputs:' | cut -d: -f2- | y2j | jq -r '.[0]'
```

(Also, you might not have `y2j` and `jq` installed)

.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Create a values.yaml for the test.
+values="$(mktemp -t values_XXXXXXXX.yaml)"
+(cat <<EOF
```
mook-as: Just `cat > "${values}" <<EOF`?

.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
```diff
@@ -0,0 +1,27 @@
...
+  clusterIP: "${CLUSTER_IP}"
+
+features:
+  eirini:
+    enabled: ${EIRINI_ENABLED}
+EOF
+) > "${values}"
+
+# Render and apply the kubecf chart.
+bazel run @helm//:helm -- template "${chart}" --name kubecf --namespace "${KUBECF_NAMESPACE}" --values "${values}" | bazel run @kubectl//:kubectl -- apply -f - --namespace "${KUBECF_NAMESPACE}"
```
mook-as: Hmm, we can't just `bazel run //dev/scf:apply` because of the different configs here? Can we do something with `bazel … --define system_domain=${SYSTEM_DOMAIN}`?

.gitlab/pipelines/stages/deploy/wait.sh
```diff
@@ -0,0 +1,20 @@
...
+
+set -o errexit -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/config/config.sh"
+
+# Wait for kubecf to start.
+wait_for_router() {
+  local timeout="1800"
+  until curl -k --fail --head --silent "https://api.${SYSTEM_DOMAIN}/v2/info" || [[ "$timeout" == "0" ]]; do sleep 1; timeout=$((timeout - 1)); done
```
mook-as: This actually waits for `router` _and_ `api`. But it's probably what we actually want.

Except it might be nice to express this in multiple lines ;)

.gitlab/pipelines/stages/tests/smoke_tests.sh
```diff
@@ -0,0 +1,34 @@
...
+
+set -o errexit -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/config/config.sh"
+
+# Trigger smoke-tests.
+bazel run //testing/smoke_tests
+
+# Wait for smoke-tests to start.
```
mook-as: The comment doesn't appear to be documenting the function immediately below it.

.gitlab/pipelines/stages/tests/smoke_tests.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/config/config.sh"
+
+# Trigger smoke-tests.
+bazel run //testing/smoke_tests
```
mook-as: It might be nice if `bazel run //testing/smoke_tests` would also display the logs…

.gitlab/deploy/runner/kubernetes/deploy.sh
```diff
@@ -33,6 +33,21 @@ if [ -z "${runner_registration_token}" ]; then
   exit 1
 fi
 
+kubectl apply -f - <<EOF
```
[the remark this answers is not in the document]
f0rmiga: This is temporary while we don't get kind to work with Gitlab. It'll go away soon.

.gitlab/pipelines/stages/cleanup.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/config/config.sh"
+
+bazel run @kubectl//:kubectl -- delete namespace "${KUBECF_NAMESPACE}" || true
```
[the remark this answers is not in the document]
f0rmiga: That is the intention. But this script will also go away with kind.

.gitlab-ci.yml
```diff
@@ -1,34 +1,83 @@
...
+deploy:cf-operator:
+  stage: deploy
+  only:
+  - external_pull_requests
+  - master
+  script:
+  - .gitlab/pipelines/stages/deploy/deploy_cf_operator.sh
+
+deploy:kubecf:
+  stage: deploy
```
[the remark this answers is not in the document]
f0rmiga: Only if the CRDs are not present, which is not the case. I changed to sequential with the kind implementation I'm currently doing.

.gitlab/pipelines/stages/deploy/deploy_cf_operator.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/config/config.sh"
+
+# Install cf-operator.
+chart="$(mktemp -t cf_operator_XXXXXXXX.tgz)"
+curl -L -o "${chart}" "${CF_OPERATOR_CHART_URL}"
```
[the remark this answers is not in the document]
f0rmiga: `wget `is not currently part of the image. I'll leave a TODO to switch to `wget` once we add it.

.gitlab/pipelines/stages/deploy/deploy_cf_operator.sh
f0rmiga: Indeed. I'll do the move of cf-operator to defs.bzl in a separate PR, then it'll be more appropriate to have a command like this.

.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
f0rmiga: We don't have `y2j` and `jq` available, but we can achieve the same with `awk`, which we do have:
```
bazel aquery //deploy/helm/scf:chart 2> /dev/null | awk 'match($0, /Outputs: \[(.*)\]/, output){ print output[1] }'
```

.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
f0rmiga: We could use the `//dev/scf:apply` target here, but we want to exercise the newly built chart, which will then be published.

.gitlab/pipelines/stages/deploy/wait.sh
f0rmiga: The router always becomes available after the api (at least with kubecf). In the next PR using kind I actually had to change this to waiting for the endpoint to become ready, which is the same as the router becoming ready.

.gitlab-ci.yml
```diff
@@ -1,34 +1,83 @@
-.bazel_docker_image: &bazel_docker_image thulioassis/bazel-docker-image:0.29.1
-.default_branch: &default_branch master
+image: thulioassis/bazel-docker-image:0.29.1
```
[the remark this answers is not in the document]
f0rmiga: I changed to where the images were published, but I couldn't get to use the github ones without logging in. I may move it back to dockerhub.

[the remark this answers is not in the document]
mook-as: Oh, publishing it on docker hub is fine for me; I just want a link to where the `Dockerfile` (and associated files) live, so that it can get rebuilt if needed.

.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
mook-as: Ah yeah, that makes sense. (See previous confusion about `output/` _not_ being a bazel output).

PR #72, opened 2019-10-27
Introduce registry credentials secret template

HeavyWombat:
## Description

Some Docker images may be located in a private Docker registry and therefore require an `imagePullSecret` in order to be pulled.

Analog to SCF v2, introduce template to create a secret containing the configured registry credentials for a hostname.

The location of the registry credential details is based on the [BOSH to Kube](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/from_bosh_to_kube.md#example-deployment-manifest-conversion-details) example manifest. I am happy to change it to another location, if it makes sense to be placed at a different spot.

## Test plan

In the `release` stanza of the `values.yaml`, for example configure a stemcell with a private registry location:
```yaml
releases:
  defaults:
    url: private.registry.io/foobar
    stemcell:
      os: BeOS
      version: 5.0.3-7.0.0_355.gd4df8ff7
  credentials:
    hostname: private.registry.io
    username: token
```

During the Helm based install, provide the secret:
```sh
helm install
  --namespace scf \
  --name scf \
  scf-3.0.0.tgz
  --set "releases.credentials.password="$(vault read /super/secret)"
```

Eventually, this should create a simple secret usable for pulling images:
```
kubectl -n scf get secret registry-credentials
NAME                   TYPE                             DATA   AGE
registry-credentials   kubernetes.io/dockerconfigjson   1      111s
```

### DIFF
--- a/deploy/helm/scf/templates/registry-credentials.yaml
+++ b/deploy/helm/scf/templates/registry-credentials.yaml
@@ -1,0 +1,16 @@
+{{- if .Values.releases.credentials -}}
+---
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/dockerconfigjson
+metadata:
+  name: registry-credentials
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/name: {{ include "scf.fullname" . }}
+    app.kubernetes.io/version: {{ default .Chart.Version .Chart.AppVersion | quote }}
+    helm.sh/chart: {{ include "scf.chart" . }}
+data:
+  .dockerconfigjson: {{ printf "{%q:{%q:{%q:%q,%q:%q,%q:%q}}}" "auths" .Values.releases.credentials.hostname "username" .Values.releases.credentials.username "password" .Values.releases.credentials.password "auth" (printf "%s:%s" .Values.releases.credentials.username .Values.releases.credentials.password | b64enc) | b64enc }}
+{{- end }}


deploy/helm/scf/templates/registry-credentials.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.releases.credentials -}}
```
edwardstudy: should we add `credentials` in values.yaml:
```
releases: 
  credentials: ~
```

f0rmiga:
@HeavyWombat Thanks for the PR!

Adding to what @mook-as already said: the secret is also not being used by any service account or pod spec. We cannot risk referencing all the secrets for private repositories in the default service account. Adding the secrets to every job is even trickier from an implementation perspective. Any ideas?

HeavyWombat:
Thanks for your feedback.

Based on [BOSH to Kube](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/from_bosh_to_kube.md#example-deployment-manifest-conversion-details), there was the implication that each release could have its own private registry credentials setup. I am not sure whether this is an edge case or something that really is required. For my use-case, we would have one private registry.

With the introduction of https://github.com/cloudfoundry-incubator/cf-operator/pull/665, I am tempted to say we can close this one here and move the whole image pull responsibility to the `cf-operator`. This means, whoever deploys the `cf-operator` and uses the `imagePullSecrets` setup in there, has also to make sure to create the respective secret.

viovanov:
Not sure I'm following the discussion. What should happen:
- we need the ability to create a docker image registry credentials secret as described [here](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials)
- we need the operator to implement [this story](https://www.pivotaltracker.com/story/show/169498158) so we can set [`imagePullSecrets`](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret)

mook-as:
@viovanov Unfortunately [#169498158](https://www.pivotaltracker.com/n/projects/2192232/stories/169498158) is incorrectly filed: it is severely lacking in necessary details.  To word it in usual Tracker terms, what is the acceptance criteria? Do we expect one secret, or multiple secrets, or…

viovanov:
@mook-as the operator part has already been implemented so I removed that story

https://github.com/cloudfoundry-incubator/cf-operator/blob/d053b0eca5dfb07f83657fdd9ee85c74a269c7d5/pkg/bosh/converter/bpm_resources.go#L155

viovanov:
@HeavyWombat are you also planning to include ops that set the `env.settings.imagePullSecrets` for each instance group?

HeavyWombat:
@viovanov Yes, this is what I mean with "introduction of https://github.com/cloudfoundry-incubator/cf-operator/pull/665", where it seems this PR is obsolete and can be closed.

Issue #12, opened 2019-09-17
PRs to fix patches in their respective upstream projects


viovanov:
The following patches should result in PRs to upstream projects.

- [x] [./eirini/opi/jobs/patch_opi_yml.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//eirini/opi/jobs/patch_opi_yml.sh)
- [x] [./eirini/opi/bpm/patch_bpm.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//eirini/opi/bpm/patch_bpm.sh)
- [ ] [./eirini/eirini-loggregator-bridge/bpm/patch_bpm.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//eirini/eirini-loggregator-bridge/bpm/patch_bpm.sh)
- [ ] [./bits/bits-service/jobs/patch_bits_config_yml.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//bits/bits-service/jobs/patch_bits_config_yml.sh)
- [ ] [./singleton-blobstore/blobstore/jobs/patch_blobstore_conf.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//singleton-blobstore/blobstore/jobs/patch_blobstore_conf.sh)
- [ ] [./singleton-blobstore/blobstore/jobs/patch_blobstore_pre-start.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//singleton-blobstore/blobstore/jobs/patch_blobstore_pre-start.sh)
- [ ] [./singleton-blobstore/blobstore/ig_resolver/patch_job_mf.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//singleton-blobstore/blobstore/ig_resolver/patch_job_mf.sh)
- [ ] [./database/mysql/jobs/patch_pre-start-setup.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//database/mysql/jobs/patch_pre-start-setup.sh)
- [ ] [./api/dotnet-core-buildpack/jobs/create_bosh_pre_start.sh](https://github.com/SUSE/kubecf/blob/master/bosh/releases/pre_render_scripts//api/dotnet-core-buildpack/jobs/create_bosh_pre_start.sh)
[...]

PR #80, opened 2019-10-30
Add OS switch in UAA pre-start patch

qu1queee:
## Description
In case a different stemcell is used, for example Ubuntu, some patches are no
longer valid for specific BOSH releases. There was an issue that the patch for
the CA certificate location failed to work on Ubuntu based images.

Initially, we tried to enhance the OS specific patches, by getting the operative
system ID, in order to skip the patch in a non SUSE OS. We realized that this wouldn´t
properly work, while the patch is evaluated under the cf-operator img, which is build
based on openSUSE. We could have try to improve the patch by other means, but this already
looks too complicated, and a feature flag will simplify the desired goal.

Introduce a feature flag called `suse_stemcell`, which adds another ops file to
the list of ops files referenced by the BOSHDeployment CRD.

This would enable `kubecf` to add more OS specific patches in the same directory
in the future.

## Motivation and Context
Kubecf should support a CF distribution based on both SUSE and Ubuntu stemcells.

## How Has This Been Tested?
We have an internal pipeline that builds Ubuntu based images(bosh releases), and we have use them, to deploy internally, into an IBM IKS cluster.

## Types of changes

## Checklist:

### DIFF
--- a/bosh/releases/pre_render_scripts/uaa/uaa/jobs/patch_pre-start.sh
+++ b/bosh/releases/pre_render_scripts/uaa/uaa/jobs/patch_pre-start.sh
@@ -6,16 +6,39 @@ target="/var/vcap/all-releases/jobs-src/uaa/uaa/templates/bin/pre-start.erb"
 
 # Patch bin/pre-start.erb for the certificates to work with SUSE.
 PATCH=$(cat <<'EOT'
-24c24
-< rm -f /usr/local/share/ca-certificates/uaa_*
----
-> rm -f /etc/pki/trust/anchors/uaa_*
-26,27c26,27
-<     echo "Adding certificate from manifest to OS certs /usr/local/share/ca-certificates/uaa_<%= i %>.crt"
-<     echo -n '<%= cert %>' >> "/usr/local/share/ca-certificates/uaa_<%= i %>.crt"
----
->     echo "Adding certificate from manifest to OS certs /etc/pki/trust/anchors/uaa_<%= i %>.crt"
->     echo -n '<%= cert %>' >> "/etc/pki/trust/anchors/uaa_<%= i %>.crt"
+--- pre-start.erb	2019-11-05 10:53:19.000000000 +0100
++++ -	2019-11-05 10:59:27.000000000 +0100
+@@ -21,11 +21,29 @@
+ }
+
+ # add certs from manifest to OS certs
+-rm -f /usr/local/share/ca-certificates/uaa_*
++source /etc/os-release
++case "${ID}" in
++  *ubuntu*)
++    rm -f /usr/local/share/ca-certificates/uaa_*
+ <% p('uaa.ca_certs', []).each_with_index do |cert, i| %>
+     echo "Adding certificate from manifest to OS certs /usr/local/share/ca-certificates/uaa_<%= i %>.crt"
+     echo -n '<%= cert %>' >> "/usr/local/share/ca-certificates/uaa_<%= i %>.crt"
+ <% end %>
++  ;;
++
++  *suse*)
++    rm -f /etc/pki/trust/anchors/uaa_*
++<% p('uaa.ca_certs', []).each_with_index do |cert, i| %>
++    echo "Adding certificate from manifest to OS certs /etc/pki/trust/anchors/uaa_<%= i %>.crt"
++    echo -n '<%= cert %>' >> "/etc/pki/trust/anchors/uaa_<%= i %>.crt"
++<% end %>
++  ;;
++
++  *)
++  echo "Unsupported operating system: ${PRETTY_NAME}"
++  exit 42
++  ;;
++esac
+
+ update_ca_certificate
+
 EOT
 )
 


mook-as:
Hi! Thanks for the patch!

I think the more correct way to do this would be to upstream the patches (see #12); in the mean time, though, making things so the person deploying the helm chart (operator? administrator?) seems wrong.  Perhaps we can patch the scripts (as in, `pre-start` and `uaa`) to check `/etc/os-release` instead?  You mentioned that the patch scripts themselves (as in, `patch_*.sh`) are always run on openSUSE in your deployment; but as the files under patch are also plain text, we can probably fix _them_ instead?  This has the benefit that the resulting files are upstream-able.

Does that sound like a reasonable thing to do in this case?

viovanov:
@mook-as this seems like an easy way to _not patch_ things if the stemcell is not SUSE
agreed that we need to upstream all of it, but it's also not that nice to require a patch for something that doesn't need a patch

HeavyWombat:
@mook-as @viovanov I cannot rule out that we did something wrong, but actually we started with using `/etc/os-release` to see whether we can change the patch itself. Problem we faces was that even with a Ubuntu stemcell for uaa, the patch execution is done by the `cf-operator` and therefore our OS checks went detected openSUSE and therefore patched the pre-start scripts.

mook-as:
@HeavyWombat Right, the patch is run within the operator image, so unless you're rebuilding that you'd get openSUSE (in `/etc/os-release`).  However, the patch is changing a shell script and _that_ could be changed to read `/etc/os-release` where it's run (which would be Ubuntu in your case), right?

```bash
patch --verbose "${target}" <<<"EOF"
…
>     echo "Adding certificate from manifest to OS certs …"
>     source /etc/os-release
>     for id in ${ID} ${ID_LIKE:-} ; do
>         case "${id}" in
>             ubuntu|debian)
>                 echo -n '<%= cert %>' >> "/usr/local/share/ca-certificates/uaa_<%= i %>.crt"
>                 break
>                 ;;
>         case suse|opensuse)
>                 echo -n '<%= cert %>' >> "/etc/pki/trust/anchors/uaa_<%= i %>.crt"
>                 break
>                 ;;
>         esac
>     done
EOF
```
Wrote that in this textbox, so it's probably got ridiculous errors…

FWIW, I'm pushing back mostly because I want to avoid knobs where the administrator can set things incorrectly and screw up their deployment.  I'd be more okay with something in our bazel configs that enabled or disabled something, and isn't accessible to the end administrator.

qu1queee:
@mook-as thanks for the input. As we mentioned at the very first comment, you can extend the patch file to detect the OS during runtime(not during patching time, similar to what @HeavyWombat mentioned), but this is already making the patch more complicated(and there is no need to do that, if this will become obsolete with https://github.com/SUSE/kubecf/issues/12). Therefore to simplify things, we consider to avoid running the patches by using a feature flag, which should be more easier to understand/manage for a kubecf administrators. 

If this PR is missing documentation to highlight the usage of that feature flag, let us know, to avoid future incorrect usage from administrators.

Also, why the need on hiding configurations from the end administrator?, for administrators is easier to enable/disable something via helm(while this is what they will consume), rather than doing this via bazel.

mook-as:
@qu1queee I am advocating for hiding the option as the administrator would not be building the images they run on (just consuming a helm chart), and hence they get no say in whether it's going to be Ubuntu or SUSE or whatever.  Hence the only thing they can actually do with the feature flag is to configure it incorrectly :smile: 

If that isn't the case (they are, for some reason, choosing the images to run), please let me know :smile_cat: 

Also, while putting the OS detection in the to-be-patched script makes the patch more complicated, that's work we'd need to do to upstream it anyway?

qu1queee:
@mook-as I´m not sure what does the administrator role implies for you, but is fair to assume that whoever consumes the helm chart can/will potentially build their set of own images with their desire OS. Similar to what happens in BOSH now with the different stemcells https://bosh.cloudfoundry.org/stemcells/. and probably this is why we had documentation around how to do this, see https://github.com/SUSE/kubecf/blob/master/doc/bosh-release-development.md#building-a-docker-image-with-fissile .

HeavyWombat:
@mook-as I had a discussion with Enrique this morning. We totally understand your point and I think our different approach is influenced by the fact that we are actually distributor, and operator of the containerized Cloud Foundry with a preference for more knobs -- knowing that too much points of configurability invite snowflakes and problems along the way.

I will look into rewriting the PR in such a way that everything is eventually covered in the UAA script at its startup time. The main benefit of this is, like you said, that we do not need a setting that could be incorrectly set by accident and cause problems, when there is a way that this is detected automagically and yield the same effect.

HeavyWombat:
Ubuntu stemcell logs:
```
+ /var/vcap/jobs/uaa/bin/pre-start
uaa-pre-start - starting at Tue Nov  5 12:13:19 UTC 2019
Adding certificate from manifest to OS certs /usr/local/share/ca-certificates/uaa_0.crt
trying to run update-ca-certificates...
Clearing symlinks in /etc/ssl/certs...
done.
Updating certificates in /etc/ssl/certs...
Doing .
149 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
Processing certificates for Java cacerts file
Certificate was added to keystore
[uaa-ctl]   Processed certificate 1 of 149
Certificate was added to keystore
[uaa-ctl]   Processed certificate 2 of 149
Certificate was added to keystore
[...]
[uaa-ctl]   Processed certificate 149 of 149
[uaa-ctl] Java keystore creation completed.
[uaa-pre-start] Installing Server SSL certificate
[uaa-pre-start] Installed Server SSL certificate
uaa-pre-start - completed at Tue Nov  5 12:14:01 UTC 2019
```

HeavyWombat:
OpenSUSE stemcell logs:
```
+ '[' -x /var/vcap/jobs/uaa/bin/pre-start ']'
+ /var/vcap/jobs/uaa/bin/pre-start
uaa-pre-start - starting at Tue Nov  5 10:45:35 UTC 2019
Adding certificate from manifest to OS certs /etc/pki/trust/anchors/uaa_0.crt
trying to run update-ca-certificates...
running /usr/lib/ca-certificates/update.d/50java.run ...
creating /var/lib/ca-certificates/java-cacerts ...
running /usr/lib/ca-certificates/update.d/70openssl.run ...
creating /var/lib/ca-certificates/openssl ...
running /usr/lib/ca-certificates/update.d/80etc_ssl.run ...
running /usr/lib/ca-certificates/update.d/99certbundle.run ...
creating /var/lib/ca-certificates/ca-bundle.pem ...
running /usr/lib/ca-certificates/update.d/99x_ubuntu_certs.run ...
Processing certificates for Java cacerts file
Certificate was added to keystore
[uaa-ctl]   Processed certificate 1 of 136
Certificate was added to keystore
[uaa-ctl]   Processed certificate 2 of 136
Certificate was added to keystore
[...]
[uaa-ctl]   Processed certificate 136 of 136
[uaa-ctl] Java keystore creation completed.
[uaa-pre-start] Installing Server SSL certificate
[uaa-pre-start] Installed Server SSL certificate
uaa-pre-start - completed at Tue Nov  5 10:46:13 UTC 2019
```

HeavyWombat:
@mook-as I finished the patch re-work and pasted log output of my two tests with different operating systems.

PR #108, opened 2019-11-05
Upstream patch for opi BPM

f0rmiga:
Ties to https://github.com/cloudfoundry-community/eirini-bosh-release/pull/69.

Part of the effort to upstream patches we currently have on kubecf - https://github.com/SUSE/kubecf/issues/12.


viovanov:
@jimmykarily how hard would it be to get this in the eirini BOSH release?

viovanov:
@f0rmiga if that gets merged, wouldn't we remove the patch altogether?

f0rmiga:
@viovanov Yes. In any case, I wanted to have this merged here so we don't get confused about differences in what we have in the patch and what we have upstream as a PR.

viovanov:
The eirini bosh relese is becoming obsolete because of #651

PR #110, opened 2019-11-05
Fixed stuck CI

f0rmiga:
CI is getting stuck. This PR fixes it by:
1. Using the kubectl binary directly (still managed by Bazel).
2. Killing the k3s cluster after the test is completed.


.gitlab/pipelines/stages/deploy/stop_k8s.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+# shellcheck disable=SC1090
+source "$(bazel info workspace)/.gitlab/pipelines/runtime/config.sh"
+
+sudo kill "$(cat "${K3S_PID_FILE}")"
```
mook-as: Would `killall ${K3S}` or whatever not work?

.gitlab-ci.yml
```diff
@@ -39,6 +39,7 @@ smoke-tests:
   - .gitlab/pipelines/stages/deploy/deploy_kubecf.sh
   - .gitlab/pipelines/stages/deploy/wait_kubecf.sh
   - .gitlab/pipelines/stages/tests/smoke_tests.sh
+  - .gitlab/pipelines/stages/deploy/stop_k8s.sh
```
mook-as: Shouldn't this be an `after_script`?

def.bzl
```diff
@@ -30,20 +30,22 @@ project = struct(
...
-            },
-            {
-                "platform": "windows",
-                "sha256": "3aa2d64f5eb9564622ddabe5f0a6c12d13d9dda90125f5a56ce41779395fa6f5",
-            },
-        ],
+        kubectl = struct(
+            platforms = struct(
+                darwin = {
+                    "url": "https://storage.googleapis.com/kubernetes-release/release/v1.14.6/bin/linux/amd64/kubectl",
```
mook-as: This no longer enforces the same version binary across platforms. Up to you to decide if that's better or worse; I guess it could be either…

f0rmiga: I think this is ok. IMO, it's pretty obvious that we should bump all platforms at once.

.gitlab-ci.yml
f0rmiga: No, because it's running in the same container.

.gitlab/pipelines/stages/deploy/stop_k8s.sh
f0rmiga: `killall: command not found`

We will have to stick to the previous approach.

f0rmiga: I moved to ps | kill.

.gitlab/pipelines/stages/deploy/stop_k8s.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+ps ax | awk '/[k]3s server/{ print $1 }' | xargs --max-line=1 --no-run-if-empty sudo kill
```
mook-as: You may have `pgrep`, actually. Both `pgrep` and `ps` are in the `procps` package.
```suggestion bash
pgrep --full 'k3s server' | xargs --no-run-if-empty sudo kill
```

.gitlab/pipelines/stages/tests/smoke_tests.sh
```diff
@@ -31,4 +33,4 @@ wait_for_smoke_tests_pod || {
 
 # Follow the logs. If the tests fail, the logs command will also fail.
 pod_name="$(smoke_tests_pod_name)"
-bazel run @kubectl//:kubectl -- logs --follow "${pod_name}" --namespace "${KUBECF_NAMESPACE}" smoke-tests-smoke-tests
+"${KUBECTL}" logs --follow "${pod_name}" --namespace "${KUBECF_NAMESPACE}" smoke-tests-smoke-tests
```
mook-as: Hmm didn't we actually tell `kubectl` to follow the one container? How did this even work?
```suggestion
"${KUBECTL}" logs --follow "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container smoke-tests-smoke-tests
```

f0rmiga:
@mook-as 2 different problems. One is the problem that the tests' job never finished. And the problem that it always returns code 0. We are solving the first problem in this PR.

PR #111, opened 2019-11-05
Add test/pr/* to trigger common jobs

f0rmiga:
With this trigger, we can push branches with the `test/` prefix for external PRs and CI will trigger.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,19 +7,20 @@ stages:
 - test
 - publish
 
+.common_refs: &common_refs
+- external_pull_requests
+- /^test\/pr\/.*$/
+- master
+
 lint:
   stage: lint
-  only:
-  - external_pull_requests
-  - master
+  only: *common_refs
   script:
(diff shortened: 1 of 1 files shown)


f0rmiga:
The https://gitlab.com/susecf/kubecf/pipelines/93910338 pipeline was triggered when I pushed this same commit to the `test/pr/0` branch.

.gitlab-ci.yml
```diff
@@ -7,19 +7,20 @@ stages:
 - test
 - publish
 
+.common_refs: &common_refs
```
mook-as: Don't really like the name, but I can't think of any better ones.
`acceptable_brances`? `only_refs` (that one's kind of bad too)?

The [documentation](https://docs.gitlab.com/ce/ci/yaml/#onlyexcept-basic) doesn't suggest anything useful. Though it _does_ say that it'll be deprecated soon…

f0rmiga: I'll think in a better label for this anchor and move to rules in a subsequent task.

PR #112, opened 2019-11-06
Do not create a tls secret if certificate or key is not provided

phil9909:
## Description
Do not create a tls secret for ingress if certifcate or key is not provided.

## Motivation and Context
We use a 3rd party to provide us with a key/certificate automatically.
But it will not overwrite existing secrets.
If `.Values.features.ingress.tls.crt` or `.Values.features.ingress.tls.key` is not set, we could just skip the creation of the secret instead of creating it empty.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:


deploy/helm/kubecf/templates/ingress.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.features.ingress.enabled -}}
+{{- if and .Values.features.ingress.tls.crt .Values.features.ingress.tls.key -}}
```
f0rmiga: If the secret is not created, how do we deal with https://github.com/SUSE/kubecf/pull/112/files#diff-f9706e030015368dbed3dcd52240b733R66-R69?

loewenstein: The secret is going to be created by our cluster provider.

PR #113, opened 2019-11-06
Add labels to ingress via helm

phil9909:
## Description
Allows to add custom labels to ingress via helm parameter `features.ingress.labels` similar to annotations.

## Motivation and Context
We use a 3rd party to provide us with a key/certificate automatically.
This requires a certain label to be present on the ingress.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

## Notice
We implemented it in a different way as `features.ingress.annotations` is implemented, because we think it's more readable that way. If this PR gets merged we are happy to also refactor the implementation for annotations to keep things consistent.


deploy/helm/kubecf/templates/ingress.yaml
```diff
@@ -53,12 +53,27 @@ metadata:
...
+    {{ if not (hasKey .Values.features.ingress.labels "app.kubernetes.io/name") }}
     app.kubernetes.io/name: {{ default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" | quote }}
+    {{- end -}}
+    {{ if not (hasKey .Values.features.ingress.labels "app.kubernetes.io/version") }}
     app.kubernetes.io/version: {{ default .Chart.Version .Chart.AppVersion | quote }}
+    {{- end -}}
+    {{ if not (hasKey .Values.features.ingress.labels "helm.sh/chart") }}
     helm.sh/chart: {{ printf "%s-%s" .Chart.Name (.Chart.Version | replace "+" "_") | quote }}
+    {{- end -}}
+{{ if .Values.features.ingress.labels }}
```
f0rmiga: Since you set a default empty `labels` key, you don't need to do this check here.

phil9909: We have to because `toYaml ` of an empty map will produce `{}` which is invalid syntax because of the other labels.

```yaml
  labels:
     app.kubernetes.io/component: "ingress" 
     helm.sh/chart: "chart-name-version"
     {}
```

PR #115, opened 2019-11-06
Fix smoke-tests exit code

f0rmiga:
Fixes https://github.com/SUSE/kubecf/issues/104.


.gitlab/pipelines/stages/tests/smoke_tests.sh
```diff
@@ -34,3 +34,6 @@ wait_for_smoke_tests_pod || {
 # Follow the logs. If the tests fail, the logs command will also fail.
 pod_name="$(smoke_tests_pod_name)"
 "${KUBECTL}" logs --follow "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container smoke-tests-smoke-tests
+
+exit_code="$("${KUBECTL}" get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output jsonpath='{.status.containerStatuses[?(@.name == "smoke-tests-smoke-tests")].state.terminated.exitCode}')"
+exit "${exit_code}"
```
mook-as:
```suggestion
exit "${exit_code:-1}"
```

In case we failed to get the exit code, in which case we should probably exit with an error.

dev/cf_operator/BUILD.bazel
```diff
@@ -10,7 +10,7 @@ helm_template(
     install_name = "cf-operator",
     namespace = project.cf_operator.namespace,
     set_values = {
-        "operator.watchNamespace": project.namespace,
+        "global.operator.watchNamespace": project.namespace,
```
mook-as: Kind of curious about this… needed to propagate it to a subchart or something? (No need to comment in the code, I think)

f0rmiga: I don't have the exact context for this change on cf-operator.

.gitlab/pipelines/stages/tests/smoke_tests.sh
```diff
@@ -34,3 +34,9 @@ wait_for_smoke_tests_pod || {
 # Follow the logs. If the tests fail, the logs command will also fail.
 pod_name="$(smoke_tests_pod_name)"
 "${KUBECTL}" logs --follow "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --container smoke-tests-smoke-tests
+
+while [ -z "$("${KUBECTL}" get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output jsonpath='{.status.containerStatuses[?(@.name == "smoke-tests-smoke-tests")].state.terminated.exitCode}')" ]; do
```
mook-as: For later refactoring: don't repeat ourselves. Something like:
```bash
jsonpath='{.status.containerStatuses[?(@.name == "smoke-tests-smoke-tests")].state.terminated.exitCode}'
while true; do
  exit_code="$("${KUBECTL}" get "${pod_name}" --namespace "${KUBECF_NAMESPACE}" --output "jsonpath=${jsonpath}")"
  if [[ -n "${exit_code}" ]] ; then
    exit "${exit_code}"
  fi
  sleep 1
done
```

f0rmiga: Fixed in https://github.com/SUSE/kubecf/pull/115/commits/1c03df002070f9e96d11ca725f04759576ea5a5f.

PR #129, opened 2019-11-08
Replace hardcoded stemcell settings with template reference

HeavyWombat:
## Description
Allow for a simple change of the stemcell at one spot in the configuration and avoid duplication in configuration.

**Please note:** I realized that different stemcell versions (to be more precise, different fissile versions) were used, for example in `eirini.yaml`, the configured version was `36.g03b4653-30.80-7.0.0_360.g0ec8d681` whereas in the `values.yaml`, it is `36.g03b4653-30.80-7.0.0_340.g2b599a90`. Having only one spot where you define the stemcell version also means that a common fissile version is enforced. However, it looks like there are missing images for some combinations of fissile and BOSH releases, for example when trying to bump to fissile `7.0.0_360.g0ec8d681`.

## Motivation and Context
When trying to switch to another stemcell, there are multiple sections in `kubecf` that need to be modified.

Therefore, replace hardcoded openSUSE references with the Go template that uses the input from the `values.yaml` to populate the respective fields.

This way, one can simply overwrite the stemcell related settings in the `values.yaml` and the rest should fall into place.

## How Has This Been Tested?
Performed a deployment with the openSUSE stemcell against an IBM Kubernetes service cluster as well as a deployment with an Ubuntu based stemcell. Like mentioned before, I was only able to test with Eirini disabled, because with Eirini it seems that fissile needs to be bumped to `7.0.0_360.g0ec8d681` for all other BOSH releases. For example, it seems that `cfcontainerization/capi:opensuse-42.3-36.g03b4653-30.80-7.0.0_360.g0ec8d681-1.83.0` does not exist at the moment.

n/a

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
@@ -10,10 +10,7 @@
   path: /releases/name=bits-service/sha1
 - type: replace
   path: /releases/name=bits-service/stemcell?
-  value:
-    alias: default
-    os: opensuse-42.3
-    version: 36.g03b4653-30.80-7.0.0_348.gc8fb3864
+  value: {{ include "kubecf.stemcellLookup" (list .Values.releases "bits-service") }}
 
 # Add quarks information to the Bits Service jobs.
 - type: replace
--- a/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
@@ -11,7 +11,7 @@
   path: /releases/name=cf-mysql?
   value:
     name: cf-mysql
-    url: docker.io/cfcontainerization
+    url: {{ .Values.releases.defaults.url | quote }}
     version: 36.19.0
     sha1: ~
 
--- a/deploy/helm/kubecf/assets/operations/instance_groups/eirini.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/eirini.yaml
@@ -136,10 +136,7 @@
     name: eirini
     version: 0.0.22
     url: {{ .Values.releases.defaults.url | quote }}
-    stemcell:
-      alias: default
-      os: opensuse-42.3
-      version: 36.g03b4653-30.80-7.0.0_360.g0ec8d681
+    stemcell: {{ include "kubecf.stemcellLookup" (list .Values.releases "eirini") }}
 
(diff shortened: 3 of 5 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
```diff
@@ -12,8 +12,8 @@
   path: /releases/name=bits-service/stemcell?
```
f0rmiga: I think the whole operation should be conditional here and below. The purpose of the `releases` key on `values.yaml` is to allow passing specifics per release. https://github.com/SUSE/kubecf/blob/master/deploy/helm/kubecf/values.yaml#L15
E.g.

```yaml
releases:
  bits-service:
    stemcell:
      os: opensuse-42.3
      version: 36.g03b4653-30.80-7.0.0_340.g2b599a90
```

So, while we don't have a more generic mechanism for setting those, we can conditionally render this `replace` if `releases.bits-service.stemcell` is set.

```yaml
{{- if .Values.relases.bits-service.stemcell }}
- type: replace
  path: /releases/name=bits-service/stemcell?
  value: {{ .Values.releases.bits-service.stemcell | mustToJson }}
{{- end }}
```
( I didn't test my snippet above - typed directly in the browser)

mook-as: This is _pure bikeshedding_ and you definitely don't have to even think about this, but we can alternatively do (also untested):
```yaml
- type: replace
  path: /releases/name=bits-service/stemcell?
  value: {{ default .Values.releases.defaults.stemcell (index (default .Values.releases.bits-service (dict)) "stemcell") | mustToJson }}
```
And possibly do this with a template helper or something.

… it seemed more sane before I typed it all out, really.

HeavyWombat: Thanks for your feedback! I really appreciate the input. At first, I tried the one-liner as @mook-as suggested, but it quickly became really ugly to read. In my opinion, Go Template is not really pleasant to read in the first place, and the my one-liner attempt even made it harder to read. This is why I opted for a helper function like Mark suggested, too. Let me know what you think about it. I do not have a ton of experience with Go templates, so please tell me if this is not the way to do it or whether there would be a more elegant way.

f0rmiga: Perfect.

PR #133, opened 2019-11-09
UAA: Force TLS between UAA and gorouter

mook-as:
## Description
This makes the traffic between UAA and gorouter be encrypted, so that we can get fully encrypted traffic from the user to UAA (though it's decrypted and re-encrypted a few times in between).

This does _not_ do mTLS.

## Motivation and Context
This is related to (but does not completely overlap) #106

## How Has This Been Tested?
Ran smoke tests locally, and deployed an app (see #106 for my configuration when testing this; ingress + SSL).

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/router.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/router.yaml
@@ -29,6 +29,14 @@
       exec:
         command: [sh, -c, 'ss -nlu sport = 3457 | grep :3457']
 
+# Add UAA CA cert to the router so the router -> UAA traffic can be encrypted
+- type: replace
+  path: /instance_groups/name=router/jobs/name=gorouter/properties/router/ca_certs
+  value: |
+    ((application_ca.certificate))
+    ((service_cf_internal_ca.certificate))
+    ((uaa_ca.certificate))
+
 # Add necessary labels to the router instance group so that the service can select it to create the
 # endpoint.
 - type: replace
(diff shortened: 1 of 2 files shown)


mook-as:
I'm not sure we actually _want_ this before cf-deployment does it.  And it might be worthwhile to work on mTLS instead.

mook-as:
Hmm, double-checking `cf-deployment`, I see that [they have this already](https://github.com/cloudfoundry/cf-deployment/blob/9c38d7510df4b54470248a841a6ba3df6c3e20c6/cf-deployment.yml#L687).  Between cf-deployment v9.1.0 and v9.5.0 (inclusive) UAA doesn't use TLS (comparing the [original revert](https://github.com/cloudfoundry/cf-deployment/commit/708a372e974a280b11f6938044faf2833ece28b4) and the [un-revert](https://github.com/cloudfoundry/cf-deployment/commit/175e576a6cd559cb47cf2ed2c9b90d4d9d56750e).)

Looking at the original PR, it was initially reverted because [it caused downtime during upgrade](https://github.com/cloudfoundry/cf-deployment/pull/747#issuecomment-491028278).  They pushed it out to the next major uaa-release (v10).  So we don't need to actually merge this, just update cf-deployment.

f0rmiga:
cf-deployment is being bumped on https://github.com/SUSE/kubecf/issues/49 - currently in progress.
We can do mTLS after we complete the bump.

Issue #124, opened 2019-11-07
Operator version should be discoverable inside the tarball released by the gitlab pipeline


viovanov:
**Is your feature request related to a problem? Please describe.**
There's no easy way to tell which operator should be used with which kubecf build.

**Describe the solution you'd like**
There should be machine-readable information in the tarball released by the pipeline.

**Describe alternatives you've considered**
- reading the commit id from the pipeline
- cloning the repo at the commit
- parsing the bazel def file to get the version

**Additional context**
Needed to chain pipelines together nicely

viovanov:
ping @f0rmiga

f0rmiga:
@viovanov We could add cf-operator as a subchart.

mook-as:
If we do a subchart, it would have to be deployed in the same namespace, right? (That might be fine, of course…)

mook-as:
Hmm. Given that cf-operator (currently) defines its CRDs in code, we can't really subchart it, because
- By the time the BOSHDeployment is created (by helm), the subchart hasn't finished deploying yet (but it has _started_)
- Which means we will fail to deploy:
> `error: unable to recognize "dev/kubecf/rendered_template.yaml": no matches for kind "BOSHDeployment" in version "fissile.cloudfoundry.org/v1alpha1"`

fargozhu:
#149

PR #149, opened 2019-11-14
Annotate operator chart url in Chart.yaml

mudler:
With this change, the bazel helm package stage will inject a new file in the `Metadata.yaml`, with a new key, `operatorChartUrl` with the project defined operator chart url in https://github.com/SUSE/kubecf/blob/3ebba68fbb03fd72df962d67d168a52419c05198/def.bzl#L9.

In this way we don't have to repackage the chart to inject a value in `Chart.yaml`, which is being generated by helm.

## Motivation and Context
We have a reference between the chart generated and the operator version needed to deploy it  #124 

## How Has This Been Tested?
Locally and with Catapult, see https://github.com/SUSE/catapult/pull/76

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/BUILD.bazel
+++ b/deploy/helm/kubecf/BUILD.bazel
@@ -2,12 +2,23 @@ package(default_visibility = ["//visibility:public"])
 
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
 load("//rules/helm:def.bzl", helm_package = "package")
+load("//:def.bzl", "project")
+load("//deploy/helm/kubecf:defs.bzl", "metadata_file_generator")
+
+metadata_file_generator(
+    name = "metadata",
+    file = "Metadata.yaml",
+    operator_chart = project.cf_operator.chart.url
+)
 
 filegroup(
     name = "chart_files_static",
     srcs = glob(
         ["**/*"],
-        exclude = ["**/BUILD.bazel"],
+        exclude = [
+            "**/BUILD.bazel",
+            "**/defs.bzl",
+        ],
     ),
 )
 
@@ -23,6 +34,9 @@ helm_package(
     name = "chart",
     srcs = [
         ":chart_files_static",
+    ],
+    generated = [
+        ":metadata",
     ],
     tars = [
         ":cf_deployment",
--- a/deploy/helm/kubecf/defs.bzl
+++ b/deploy/helm/kubecf/defs.bzl
@@ -1,0 +1,8 @@
+def metadata_file_generator(name, file, operator_chart, visibility=None):
+  native.genrule(
+    name = name,
+    srcs = [],
+    outs = [file],
+    cmd = "echo 'operatorChartUrl: \"{}\"' > $@".format(operator_chart),
+    visibility = visibility,
+  )
--- a/rules/helm/def.bzl
+++ b/rules/helm/def.bzl
@@ -3,7 +3,7 @@ def _package_impl(ctx):
     output_tgz = ctx.actions.declare_file(output_filename)
     outputs = [output_tgz]
     ctx.actions.run(
-        inputs = [] + ctx.files.srcs + ctx.files.tars,
+        inputs = [] + ctx.files.srcs + ctx.files.tars + ctx.files.generated,
         outputs = outputs,
         tools = [ctx.executable._helm],
         progress_message = "Generating Helm package archive {}".format(output_filename),
@@ -12,6 +12,8 @@ def _package_impl(ctx):
             "PACKAGE_DIR": ctx.attr.package_dir,
             # TODO(f0rmiga): Figure out a way of working with paths that contain spaces.
             "TARS": " ".join([f.path for f in ctx.files.tars]),
+            # TODO(mudler): Support also nested folders and paths with spaces
+            "GENERATED": " ".join([f.path for f in ctx.files.generated]),
             "HELM": ctx.executable._helm.path,
             "CHART_VERSION": ctx.attr.chart_version,
             "APP_VERSION": ctx.attr.app_version,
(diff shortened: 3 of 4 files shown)


rules/helm/package.sh
```diff
@@ -4,15 +4,18 @@ set -o errexit -o nounset
...
 mkdir -p "${build_dir}"
+
 cp -L -R "${PACKAGE_DIR}"/* "${build_dir}"
 
 for t in ${TARS}; do
-  tar xf "${t}" -C "${build_dir}"
+  tar xf "${t}" -C "${build_dir}"  > /dev/null
 done
 
+echo "operatorChartUrl: \"${OPERATOR_URL}\"" >> "${build_dir}"/Metadata.yaml
```
mook-as: Seems slightly off to put the very-much-kubecf-chart stuff into a generic-looking helm chart creation rule.

Given that `Metadata.yaml` doesn't otherwise exist (and doesn't appear to be a standard thing in either [v2] or [v3]), wouldn't it make sense to have a separate rule to create it in `//deploy/helm/kubecf`, and just add it as an extra file on the `//rules/helm` layer?  Especially since it already takes `srcs`.

Bikeshed, feel free to ignore: `Metadata.yaml` vs `metadata.yaml` vs…

[v2]: https://v2.helm.sh/docs/developing_charts/#the-chart-file-structure
[v3]: https://helm.sh/docs/topics/charts/#the-chart-file-structure

rules/helm/package.sh
```diff
@@ -4,15 +4,18 @@ set -o errexit -o nounset
...
+
 "${HELM}" init --client-only > /dev/null
 "${HELM}" package "${build_dir}" \
   --version="${CHART_VERSION}" \
   --app-version="${APP_VERSION}" > /dev/null
 
-mv "${OUTPUT_FILENAME}" "${OUTPUT_TGZ}"
+mv "${OUTPUT_FILENAME}" "${OUTPUT_TGZ}"
```
mook-as: Nit: new line at EOF?

rules/helm/package.sh
mudler: > Seems slightly off to put the very-much-kubecf-chart stuff into a generic-looking helm chart creation rule.
> 
> Given that `Metadata.yaml` doesn't otherwise exist (and doesn't appear to be a standard thing in either [v2](https://v2.helm.sh/docs/developing_charts/#the-chart-file-structure) or [v3](https://helm.sh/docs/topics/charts/#the-chart-file-structure)), wouldn't it make sense to have a separate rule to create it in `//deploy/helm/kubecf`, and just add it as an extra file on the `//rules/helm` layer? Especially since it already takes `srcs`.

Yes `Metadata.yaml` doesn't exists in helm,  there is no other way around to add a custom value to the chart directly from the helm cli as far as I can tell.

I don't know, to me looks like adding a layer of over-complexity, because I would expect all the chart metadata needed for packaging in  the `helm` rule, as we already specify (hardcoded): 
```
   chart_name = "kubecf",
    chart_version = "3.0.0",
    app_version = "2.0",
```
 in the same section. But I will try to have a look

> 
> Bikeshed, feel free to ignore: `Metadata.yaml` vs `metadata.yaml` vs…

I picked capital letter just to be coherent with `Chart.yaml` which is already there, but no strong opinion here.

mudler: Should be fixed in https://github.com/SUSE/kubecf/pull/149/commits/1e7d013029268c6cd42d44b0fd15ee66a270f3b0, I've also added a macro to generate the metadata and added the extra file in the helm layer

rules/helm/package.sh
mudler: Fixed in https://github.com/SUSE/kubecf/pull/149/commits/1e7d013029268c6cd42d44b0fd15ee66a270f3b0 (will squash eventually)

rules/helm/def.bzl
```diff
@@ -12,6 +12,8 @@ def _package_impl(ctx):
             "PACKAGE_DIR": ctx.attr.package_dir,
             # TODO(f0rmiga): Figure out a way of working with paths that contain spaces.
             "TARS": " ".join([f.path for f in ctx.files.tars]),
+            # TODO(mudler): Support also nested folders and paths with spaces
+            "GENERATED": " ".join([f.path for f in ctx.files.generated]),
```
mook-as: We don't need this if we wrap it in a tar, but that's not really necessary.

deploy/helm/kubecf/BUILD.bazel
```diff
@@ -2,12 +2,20 @@ package(default_visibility = ["//visibility:public"])
...
+    file = "Metadata.yaml",
+    operator_chart = project.cf_operator.chart.url
+)
 
 filegroup(
     name = "chart_files_static",
     srcs = glob(
         ["**/*"],
-        exclude = ["**/BUILD.bazel"],
+        exclude = ["**/BUILD.bazel", "**/defs.bzl"],
```
f0rmiga: [nit] Make this multi-line.

deploy/helm/kubecf/BUILD.bazel
```diff
@@ -19,11 +27,15 @@ pkg_tar(
     ],
 )
 
+
```
f0rmiga: [nit] Unnecessary extra line.

deploy/helm/kubecf/defs.bzl
```diff
@@ -0,0 +1,8 @@
+def metadata_file_generator(name, file, operator_chart, visibility=None):
+  native.genrule(
+    name = name,
+    srcs = [],
+    outs = [file],
+    cmd = "echo 'operatorChartUrl: \"%s\"' > $@" % operator_chart,
```
f0rmiga: Prefer `"{}".format(...)` syntax in Python.

mudler: Fixed, squashed and rebased :+1:

PR #153, opened 2019-11-15
Let the user configure the storage class used for components

### LINKED ISSUE #26: Configurable storage classes
Right now, we don't allow users to configure the storage class to be used for persistent volumes.
It should be exposed in `values.yaml` and altered using an ops file.

The instance groups' [`persistent_disk_type`](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/examples/bosh-deployment/boshdeployment-with-persistent-disk.yaml#L31) needs to be set.

jimmykarily:
Resolves #26

Any idea which docs I should update?

## How Has This Been Tested?
Deployed on kind with this in my values file:

```
kube:
  storage_class: custom
```
pvcs were created on the `custom` storageclass (which was not the default)

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
@@ -50,13 +50,14 @@
   value: local
 
 # Attach a persistent disk to bits-service VM to store eirinifs
-# TODO: storage class type should be configurable
-# - type: replace
-#   path: /instance_groups/name=bits/persistent_disk_type?
-#   value: ???
 - type: replace
   path: /instance_groups/name=bits/persistent_disk?
   value: 20480 # 20GB
+{{- if .Values.kube.storage_class }}
+- type: replace
+  path: /instance_groups/name=bits/persistent_disk_type?
+  value: {{ .Values.kube.storage_class }}
+{{- end }}
 
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/bits_service?/enabled?
--- a/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
@@ -16,11 +16,14 @@
     sha1: ~
 
 # Configure the persistent disk in the way that cf-operator can provision.
-- type: remove
-  path: /instance_groups/name=database/persistent_disk_type
 - type: replace
   path: /instance_groups/name=database/persistent_disk?
   value: 20480 # 20GB
+{{- if .Values.kube.storage_class }}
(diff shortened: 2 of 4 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
```diff
@@ -50,13 +50,12 @@
...
 # Attach a persistent disk to bits-service VM to store eirinifs
-# TODO: storage class type should be configurable
-# - type: replace
-#   path: /instance_groups/name=bits/persistent_disk_type?
-#   value: ???
 - type: replace
   path: /instance_groups/name=bits/persistent_disk?
   value: 20480 # 20GB
+- type: replace
+  path: /instance_groups/name=bits/persistent_disk_type?
```
f0rmiga: Since the path exists in the original manifest, there is no need to use `?`. I believe it's true for all the cases in this PR.

I think it also makes sense to use a conditional render for this operation to exist only if `kube.storage_class` is set. The reason behind it is that we don't change the current behaviour of using whatever default storage class the user has.

Also, we need to add this key to the default `values.yaml`. I think there is a good place to document it. The default value can simply be null `~`.

jimmykarily: I think `?` is still needed for bits (https://github.com/cloudfoundry/cf-deployment/blob/master/operations/bits-service/use-bits-service.yml#L86). I will remove from the rest. Adding conditional and default value with docs.

jimmykarily: Is this the default `values.yaml` file you mean? https://github.com/SUSE/kubecf/blob/master/deploy/helm/kubecf/values.yaml . Adding there.

f0rmiga: That's the one!

deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
```diff
@@ -50,13 +50,14 @@
...
 
 # Attach a persistent disk to bits-service VM to store eirinifs
-# TODO: storage class type should be configurable
-# - type: replace
-#   path: /instance_groups/name=bits/persistent_disk_type?
-#   value: ???
 - type: replace
   path: /instance_groups/name=bits/persistent_disk?
   value: 20480 # 20GB
+{{ if .Values.kube.storage_class }}
```
f0rmiga: [very nit] Use `{{- if ...}}` and `{{- end }}` to avoid a blank line in the rendered template.

jimmykarily: done

PR #160, opened 2019-11-18
Switch helm to use external_binary rule

f0rmiga:
## Description

The `helm` external dependency was managed by its own Bazel rule. After refactoring the `external_binary` rule, it now accepts compressed files as input, and therefore `helm` can be managed using this rule.

## Motivation and Context

Reduce code and better external dependency management. The old Helm rule downloaded all the binaries for the various platforms. With the use of `external_binary`, only the file for the host platform is downloaded.

## How Has This Been Tested?

Tested the various external binaries that Bazel manages as well as helm using the new rule.

## Types of changes

## Checklist:


rules/external_binary/def.bzl
```diff
@@ -12,22 +12,43 @@ def _external_binary_impl(ctx):
...
+                name = ctx.attr.name,
+            )
+        ]
+        res = ctx.execute(flatten)
+        if res.stdout != "":
+            print("STDOUT:")
+            print(res.stdout)
+        if res.stderr != "":
+            print("STDERR:")
+            print(res.stderr)
```
mook-as: This _seems_ to work, but I'm not sure this covers all the cases (and I think it needs some style fixups):

```python
    url = info.get("url")
    args = {
        "url": url,
        "sha256": info.get("sha256", ""),
    }
    if any([url.endswith(suffix) for suffix in [".zip", ".tar.gz", ".tgz"]]):
        ctx.download_and_extract(output="{name}/{name}_out".format(name = ctx.attr.name), **args)
        build_contents = """
        package(default_visibility = ["//visibility:public"])
        filegroup(name = "{name}", srcs = glob(["**/{name}*"]))
        """.format(name = ctx.attr.name)
    else:
        args["executable"] = True
        ctx.download(output="{name}/{name}".format(name = ctx.attr.name), **args)
        build_contents = """
        package(default_visibility = ["//visibility:public"])
        exports_files(glob(["**/*"]))
        """.format(name = ctx.attr.name)

    build_contents = '\n'.join([x.lstrip(' ') for x in build_contents.splitlines()])
    ctx.file("{name}/BUILD.bazel".format(name = ctx.attr.name), build_contents)
```

f0rmiga: This didn't work as-is but gave me some light on how to do it. A `filegroup` cannot be executable, but we can have a `genrule` that is executable that takes the `filegroup` as input.

I added a `cmd_bash` and a `cmd_bat` that should cover all platforms that we want to support.

rules/external_binary/def.bzl
```diff
@@ -12,22 +12,40 @@ def _external_binary_impl(ctx):
...
+            srcs = glob(["**/{name}*"]),
+        )
+        genrule(
+            name = "{name}",
+            srcs = [":{name}_filegroup"],
+            outs = ["executable"],
+            cmd_bash = "cp $(location :{name}_filegroup) $@", # Runs on *nix.
+            cmd_bat = "copy $(location :{name}_filegroup) $@", # Runs on Windows.
+            executable = True,
+        )
```
mook-as: We could use [@bazel_skylib//rules:copy_file](https://github.com/bazelbuild/bazel-skylib/blob/1.0.2/docs/copy_file_doc.md) instead, probably.

PR #161, opened 2019-11-18
Add yaml linting

viccuad:
## Description

This PR adds yaml linting to the dev/ scripts and gitlab, which lints for yaml files, excluding `deploy/helm/kubecf/*`, as those are helm templates that should be better linted with `helm lint`.

To achieve it, it adds a bazel target `//dev/linters:yamllint` to install the yamllint python module, and a python wrapper to call it. I haven't found any other way in Bazel to pull a python module and then call it.

It provides the script `dev/linters/yamllint.sh` that calls the python wrapper. And it updates `.gitlab-ci.yaml` to call this last script.

Bazel's pip_install() doesn't install pip, so this PR needs https://github.com/f0rmiga/bazel-docker-image/pull/3.

## How Has This Been Tested?

I have tested it locally by running `dev/linters/yamllint.sh`.

I have tested it against gitlab-ci by editing the `.gitlab-ci.yaml` to `zypper in pip`, and running `gitlab-runner exec docker lint`.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,6 +13,14 @@ external_binary(
     windows = project.shellcheck.platforms.windows,
 )
 
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
+    name = "rules_python",
+    remote = "https://github.com/bazelbuild/rules_python.git",
+    commit = project.rules_python.commit,
+)
+
 http_archive(
     name = "cf_deployment",
     build_file_content = """
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -21,6 +21,19 @@ git_repository(
     commit = project.rules_python.commit,
 )
 
+load("@rules_python//python:pip.bzl", "pip_repositories", "pip3_import")
+
+pip_repositories()
+
+pip3_import(
+    name = "yamllint",
+    requirements = "//dev/linters:requirements.txt",
+)
+
+load("@yamllint//:requirements.bzl", "pip_install")
+
+pip_install()
+
 http_archive(
     name = "cf_deployment",
     build_file_content = """
--- a/dev/linters/BUILD.bazel
+++ b/dev/linters/BUILD.bazel
@@ -1,0 +1,11 @@
+load("@yamllint//:requirements.bzl","all_requirements")
+
+load("@rules_python//python:defs.bzl", "py_binary")
+
+py_binary(
+    name = "yamllint",
+    srcs = ["//dev/linters:run_yamllint.py"],
+    main = "run_yamllint.py",
+    python_version = "PY3",
+    deps = all_requirements,
+)
--- a/dev/linters/run_yamllint.py
+++ b/dev/linters/run_yamllint.py
@@ -1,0 +1,9 @@
+import os
+from yamllint import cli
+
+bazel_working_dir = os.environ['BUILD_WORKING_DIRECTORY']
+
+if bazel_working_dir:
+    os.chdir(bazel_working_dir)
+
+cli.run()
--- a/dev/linters/yamllint.sh
+++ b/dev/linters/yamllint.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+# shellcheck disable=SC2046
+# We want word splitting with find.
+bazel run //dev/linters:yamllint -- \
+      -d "{extends: relaxed, rules: {line-length: {max: 120}}}" \
+      --strict $(find . -type f \
+                      -path "./deploy/helm/kubecf/values.*" \
+                      -or -not -path "./deploy/helm/kubecf/*" \
+                      -name '*.yaml' -or -name '*.yml')
--- a/doc/Contribute.md
+++ b/doc/Contribute.md
@@ -97,18 +97,20 @@ associated documentation, if we have any.
 
 ## Linting
 
-Currently only one linter is available:
+Currently, 2 linters are available:
 
   - `dev/linters/shellcheck.sh`
+  - `dev/linters/yamllint.sh`
 
-Invoke this linter as
+Invoke these linters as
 
 ```sh
 dev/linters/shellcheck.sh
+dev/linters/yamllint.sh
 ```
 
-to run shellcheck on all `.sh` files found in the entire checkout and
-report any issues found.
+to run shellcheck on all `.sh` files found in the entire checkout, or yamllint
+on all `.yaml` or `.yml` files respectively, and report any issues found.
 
 ## Patching
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -17,6 +17,7 @@ lint:
   only: *common_refs
   script:
   - ./dev/linters/shellcheck.sh
+  - ./dev/linters/yamllint.sh
 
 build:chart:
   stage: build
--- a/def.bzl
+++ b/def.bzl
@@ -112,6 +112,9 @@ project = struct(
             },
         ),
     ),
+    rules_python = struct(
+        commit = "94677401bc56ed5d756f50b441a6a5c7f735a6d4"
+    ),
     mysql_chart = struct(
         version = "1.3.3",
         sha256 = "9ef4ce3693eb2a7428598f9dae833ee546eac9c105b4005c6d7375c55e33bdff",
--- a/dev/linters/requirements.txt
+++ b/dev/linters/requirements.txt
@@ -1,0 +1,1 @@
+yamllint
--- a/doc/linters.md
+++ b/doc/linters.md
@@ -9,3 +9,8 @@ The following linters are available:
 
     Runs shellcheck on all `.sh` files found in the entire checkout
     and reports any issues found.
+
+  - `yamllint.sh`:
+
+    Runs yamllint on all `.{yaml,yml}` files found in the entire checkout
+    and reports any issues found.


viccuad:
Linting the helm templates fail, eg:
```
./deploy/helm/kubecf/assets/operations/instance_groups/bits.yaml
  1:3       error    syntax error: expected the node content, but found '-'
```
We can't use `helm lint` as it only accepts `Chart.yml` as input. Another option would be to litter all templates with `# yamllint disable-line`, but that doesn't seem like a better option.

f0rmiga:
https://github.com/SUSE/kubecf/pull/165 Should unblock this PR.

viccuad:
Rebased on top of master to include https://github.com/SUSE/kubecf/pull/165.
Edited PR message to signify that this PR lints yaml files excluding `deploy/helm/kubecf/*`, as those are helm templates that should be better linted with `helm lint`

WORKSPACE
```diff
@@ -13,6 +13,20 @@ external_binary(
     windows = project.shellcheck.platforms.windows,
 )
 
+http_archive(
+    name = "rules_python",
+    url = "https://github.com/bazelbuild/rules_python/releases/download/0.0.1/rules_python-0.0.1.tar.gz",
```
f0rmiga: We are controlling all the repository rules versions and hash sums on the top-level def.bzl. Check the `bazel_skylib` target:

https://github.com/SUSE/kubecf/blob/6e48b3fd04c40d2a64d5faeb33b2d6f73a194d32/WORKSPACE#L73-L77

https://github.com/SUSE/kubecf/blob/6e48b3fd04c40d2a64d5faeb33b2d6f73a194d32/def.bzl#L91-L94

WORKSPACE
```diff
@@ -13,6 +13,20 @@ external_binary(
...
+    sha256 = "aa96a691d3a8177f3215b14b0edc9641787abaaa30363a080165d06ab65e1161",
+)
+load("@rules_python//python:pip.bzl", "pip_repositories","pip_import")
```
f0rmiga: Could you add a line break between each function call, please? Once we add `skylint`, it will ask the same thing. :)

dev/linters/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
+load("@yamllint//:requirements.bzl", "all_requirements")
+load("@rules_python//python:defs.bzl", "py_binary")
+
+py_binary(
+    name = "yamllint",
+    srcs = ["//dev/linters:run_yamllint.py"],
+    data = [
```
f0rmiga: Remove `data` completely, it's not being used.

dev/linters/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+        # "//dev/linters:setup.cfg",
+    ],
+    main = "run_yamllint.py",
+    python_version = "PY2",
+    deps = [
```
f0rmiga: Unnecessary `[]`. Just do `deps = all_requirements`.

dev/linters/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+    ] + all_requirements,
+)
```
f0rmiga: Missing line break at the end of the file.

dev/linters/run_yamllint.py
```diff
@@ -0,0 +1,10 @@
+import os
+from yamllint import cli
+# import yamllint
```
f0rmiga: Remove commented code.

doc/Contribute.md
```diff
@@ -97,18 +97,20 @@ associated documentation, if we have any.
 
 ## Linting
 
-Currently only one linter is available:
+Currently 2 linters are available:
```
f0rmiga: [nit] Missing a comma after `Currently`.

doc/Contribute.md
```diff
@@ -129,7 +131,7 @@ around the deployment manifest the ability to easily toggle various
 preset configurations, for example the use of eirini instead of diego
 as the application scheduler.
 
-### Features
+
```
f0rmiga: Why this change?

dev/linters/yamllint.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+# shellcheck disable=SC2046
+# We want word splitting with find.
+bazel run //dev/linters:yamllint -- \
+      -d "{extends: relaxed, rules: {line-length: {max: 120}}}" \
+      --strict $(find . -not -path "./deploy/helm/kubecf/*" -name '*.yaml' -or -name '*.yml' )
```
f0rmiga: [nit] Missing type file only.
[Super nit] Extra space on the closing parenthesis.

f0rmiga: I think we still want to lint `values.yaml`.

f0rmiga: I'm wondering if it would worth running the `yamllint` on the rendered templates. It would require inverting the build step with the linting step on CI. Something like:

```
bazel run //dev/linters:yamllint -- ... <(helm template --name kubecf ...)
```

Ping @mook-as.

mook-as: Running linters on generated output can often be a bad idea (because we would often get generated output that would be different than how a human would write things from scratch, e.g. with very long lines).

dev/linters/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+load("@rules_python//python:defs.bzl", "py_binary")
+
+py_binary(
+    name = "yamllint",
+    srcs = ["//dev/linters:run_yamllint.py"],
+    data = [
+        # "//dev/linters:setup.cfg",
+    ],
+    main = "run_yamllint.py",
+    python_version = "PY2",
```
mook-as: We should probably be using Python3 throughout instead? I don't think we have any other Python dependencies at the moment, and Python 2 is going to EOL "soon". (And yamllint supports Python 3, as far as I can tell.)

dev/linters/yamllint.sh
f0rmiga: In this case, we should aim to generate human-like manifests. A linter on the generated output would help us with achieving that. It's a different case than running a linter on, let's say, gRPC generated code, although I'd argue that it may be subject to some pre-compile checks. The reason why I believe we should aim to generate human-like manifests is that it makes inspection easy for everyone from devs to final customers.

dev/linters/BUILD.bazel
f0rmiga: +1

Both Python 2 and 3 are available on the Docker image (along with pip). I'll drop Python 2 as soon as it reaches EOL.

dev/linters/yamllint.sh
viccuad: Good catch, I have added `deploy/helm/kubecf/values.*`
We could always lint before and after build.

viccuad:
Many thanks for the review :). Also, I'm quite embarrassed by the simple mistakes I did.

I have fixed the typos and force-pushed with new changes, which I think it's gonna be cleaner:
- Using `git_repository` instead of `http_archive`, needed for python3 support, saving the commit sha in `def.bzl`
- pip_repositories and pip_import need to be evaluated at WORKSPACE-time, added commit comments. The rest can happily go on ./dev/linters/BUILD.bazel.

viccuad:
Rebased on top of master to fix conflicts.

dev/linters/yamllint.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+# shellcheck disable=SC2046
+# We want word splitting with find.
+bazel run //dev/linters:yamllint -- \
```
mook-as: I'm having issues running this; I get:

```
$ bazel run //dev/linters:yamllint
Starting local Bazel server and connecting to it...
DEBUG: Rule 'rules_python' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1573842889 -0500"
INFO: Analyzed target //dev/linters:yamllint (18 packages loaded, 210 targets configured).
INFO: Found 1 target...
INFO: Deleting stale sandbox base …/1902c1ccc22df327a938dca0d6a953d0/sandbox
Target //dev/linters:yamllint up-to-date:
  bazel-bin/dev/linters/yamllint
INFO: Elapsed time: 14.173s, Critical Path: 0.06s
INFO: 0 processes.
INFO: Build completed successfully, 4 total actions
INFO: Build completed successfully, 4 total actions
Traceback (most recent call last):
  File "…/1902c1ccc22df327a938dca0d6a953d0/execroot/kubecf/bazel-out/k8-fastbuild/bin/dev/linters/yamllint.runfiles/kubecf/dev/linters/run_yamllint.py", line 2, in <module>
    from yamllint import cli
  File "…/1902c1ccc22df327a938dca0d6a953d0/execroot/kubecf/bazel-out/k8-fastbuild/bin/dev/linters/yamllint.runfiles/yamllint_pypi__yamllint_1_19_0/yamllint/cli.py", line 25, in <module>
    from yamllint import linter
  File "…/1902c1ccc22df327a938dca0d6a953d0/execroot/kubecf/bazel-out/k8-fastbuild/bin/dev/linters/yamllint.runfiles/yamllint_pypi__yamllint_1_19_0/yamllint/linter.py", line 19, in <module>
    import yaml
  File "…/1902c1ccc22df327a938dca0d6a953d0/execroot/kubecf/bazel-out/k8-fastbuild/bin/dev/linters/yamllint.runfiles/yamllint_pypi__PyYAML_5_1_2/yaml/__init__.py", line 374
    class YAMLObject(metaclass=YAMLObjectMetaclass):
                              ^
SyntaxError: invalid syntax
```

This appears to be because:

```
$ cat bazel-bin/dev/linters/yamllint
#!/usr/bin/env python

…
```

This appears to be https://github.com/bazelbuild/bazel/issues/4815. We seem to need to change how bazel is run to fix this:

```suggestion
bazel run --incompatible_use_python_toolchains //dev/linters:yamllint -- \
```

Looking around, maybe my version of bazel (0.26) is too old?  Ah, yes, installing bazel 0.29 (the newest version in `obs://devel:tools:building`) fixed things.  All good, then.

PR #162, opened 2019-11-18
Use external database

f0rmiga:
## Description

Adds the external database functionality via Helm values.

Resolves https://github.com/SUSE/kubecf/issues/4.

## Motivation and Context

Using an external database is desired on large setups and on setups that consume managed databases.

## How Has This Been Tested?

The local development and testing was conducted using the provided script for deploying a local mysql database on a separate namespace and deploying kubecf with the example values provided in the updated documentation. The kubecf deployment completed successfully and smoke-tests passed.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
@@ -170,6 +170,8 @@
 - type: replace
   path: /instance_groups/name=scheduler/jobs/name=cc_deployment_updater/properties/ccdb/roles/name=cloud_controller/name
   value: *external_cc_database_username
+- type: remove
+  path: /instance_groups/name=scheduler/jobs/name=cc_deployment_updater/properties/ccdb/ca_cert?
 
 - type: replace
   path: /instance_groups/name=diego-api/jobs/name=bbs/properties/diego/bbs/sql/db_driver
--- a/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
@@ -1,3 +1,323 @@
+{{- if .Values.features.external_database.enabled -}}
+
+# Remove the database instance group and release.
+- type: remove
+  path: /instance_groups/name=database
+- type: remove
+  path: /releases/name=pxc
+
+# Remove the database variables.
+- type: remove
+  path: /variables/name=cf_mysql_mysql_admin_password
+- type: remove
+  path: /variables/name=cf_mysql_mysql_cluster_health_password
+- type: remove
+  path: /variables/name=cf_mysql_mysql_galera_healthcheck_endpoint_password
+- type: remove
+  path: /variables/name=cf_mysql_mysql_galera_healthcheck_password
+- type: remove
+  path: /variables/name=cf_mysql_proxy_api_password
+- type: remove
+  path: /variables/name=network_policy_database_password
+- type: remove
+  path: /variables/name=network_connectivity_database_password
+- type: remove
+  path: /variables/name=routing_api_database_password
+- type: remove
+  path: /variables/name=locket_database_password
+- type: remove
+  path: /variables/name=cc_database_password
+- type: remove
+  path: /variables/name=credhub_database_password
+- type: remove
+  path: /variables/name=diego_database_password
+- type: remove
+  path: /variables/name=uaa_database_password
+- type: remove
+  path: /variables/name=pxc_galera_ca
+- type: remove
+  path: /variables/name=pxc_server_ca
+- type: remove
+  path: /variables/name=galera_server_certificate
+- type: remove
+  path: /variables/name=mysql_server_certificate
+
+# Remove database links.
+- type: replace
+  path: /instance_groups/name=uaa/jobs/name=uaa/consumes?
+  value: {database: nil}
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/consumes?
+  value: {database: nil}
+- type: replace
+  path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/consumes?
+  value: {database: nil}
+- type: replace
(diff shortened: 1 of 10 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
```diff
@@ -1,3 +1,323 @@
...
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaadb/roles/name=uaa/name
+  value: {{ .Values.features.external_database.databases.uaa.username | quote }}
+- type: remove
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaa?/ca_certs
+- type: remove
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaadb/tls_enabled?
+
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/ccdb/db_scheme
+  value: &external_cc_database_scheme {{ .Values.features.external_database.type | quote }}
```
mook-as: Would it make sense to replace all of `…/name=cloud_controller_ng/properties/ccdb` as one block?

deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
```diff
@@ -1,3 +1,323 @@
...
+- type: replace
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaadb/roles/name=uaa/password
+  value: {{ .Values.features.external_database.databases.uaa.password | quote }}
+- type: replace
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaadb/roles/name=uaa/name
+  value: {{ .Values.features.external_database.databases.uaa.username | quote }}
+- type: remove
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaa?/ca_certs
+- type: remove
+  path: /instance_groups/name=uaa/jobs/name=uaa/properties/uaadb/tls_enabled?
```
mook-as: This should still be a thing _sometimes_, because the external DB could support SSL?

deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
```diff
@@ -1,3 +1,323 @@
...
+  value: {{ .Values.features.external_database.databases.credhub.username | quote }}
+- type: remove
+  path: /instance_groups/name=credhub/jobs/name=credhub/properties/credhub/data_storage/tls_ca?
+- type: replace
+  path: /instance_groups/name=credhub/jobs/name=credhub/properties/credhub/data_storage/require_tls?
+  value: false
+
+{{- else -}}
+
+# Override the addresses for the jobs under the scheduler instance group.
```
mook-as: Not sure this should be in this file.

I think, ideally, we'd _not_ change this file, but have an _additional_ ops file to remove it all again (applied after this one).

But that's because I want to have a `…/features/` directory to organize all our features-based things, I think.

dev/external_database/deploy_mysql.sh
```diff
@@ -0,0 +1,70 @@
...
+    --namespace "${namespace}"
+
+"{KUBECTL}" wait pod \
+  --for condition=ready \
+  --namespace "${namespace}" \
+  --selector "app=${name}" \
+  --timeout 300s
+
+# Ensure the database is fully functional.
+until echo "SELECT 'Ready!'" | "{KUBECTL}" run mysql-client --rm -i --restart='Never' --image docker.io/mysql --namespace "${namespace}" --command -- \
```
mook-as: You shouldn't need an explicit `docker.io` here. (It's technically `docker.io/library/mysql`, I think?)

Do we need to give an explicit image version (tag) here?

Do we want to import the [`sql-readiness`] stuff here?  And maybe [`database-seeder`]?

[`sql-readiness`]: https://github.com/SUSE/scf/tree/develop/src/scf-release/src/acceptance-tests-brain/sqlreadiness
[`database-seeder`]: https://github.com/SUSE/scf-helper-release/tree/master/src/database-seeder

dev/external_database/deploy_mysql.sh
```diff
@@ -0,0 +1,70 @@
...
+    mysql --host="${name}.${namespace}.svc" --user=root --password="${root_password}"; do
+      sleep 1
+done
+
+"{KUBECTL}" run mysql-client --rm -i --restart='Never' --image docker.io/mysql --namespace "${namespace}" --command -- \
```
mook-as: [Ditto](https://cdn.bulbagarden.net/upload/3/36/132Ditto.png).

deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
f0rmiga: It'd be more complicated to handle the `.../properties/ccdb/databases/tag=cc/name`.

f0rmiga: I assume we don't want to completely replace this.

deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
f0rmiga: Created https://github.com/SUSE/kubecf/issues/191 as a follow-up task.

deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
f0rmiga: I generally dislike having the same condition in multiple places of the project. I.e. a single helm value controlling multiple conditions in multiple different files. It's acceptable in some cases that we cannot do it from a single file. We should discuss if the advantage of having `.../features/` would pay off.

dev/external_database/deploy_mysql.sh
f0rmiga: Given that we don't publish the binaries for sql-readiness and database-seeder, and that the current approach on this PR works, I'll punt this to later once we have more time to this enhancement.

f0rmiga: I removed the `docker.io` registry and pinned the tag.

PR #169, opened 2019-11-20
Update general documentation

qu1queee:
[#169615069](https://www.pivotaltracker.com/n/projects/2192232/stories/169615069)

- Update cf-operator helm chart URL to the current one in use.
- Add note about potential problems when deleting the kubecf ns
  manually, and losing the labelling generated by the cf-operator
  webhooks on start.
- Fix YAML lints in the general.md file

## Motivation and Context
Required for cf-operator v1.0 release shipment https://www.pivotaltracker.com/n/projects/2192232/stories/169615069

## How Has This Been Tested?
Deploying operators in Kind, then deleting the generated `kubecf` namespace, then deploying kubecf charts into `kubecf` namespace. This will eventually affect a service that relies on the label in the `kubecf` ns, leading to a deployment that will never finish, because services do not have endpoints.

None
## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly.

### DIFF
--- a/doc/dev/general.md
+++ b/doc/dev/general.md
@@ -55,13 +55,27 @@ with Helm Tiller pre-installed.
 helm install --name cf-operator \
      --namespace cfo \
      --set "global.operator.watchNamespace=kubecf" \
-     https://s3.amazonaws.com/cf-operators/helm-charts/cf-operator-v0.4.1%2B92.g77e53fda.tgz
+     https://s3.amazonaws.com/cf-operators/helm-charts/cf-operator-v0.4.2-147.gb88e4296.tgz
 ```
 
 In the example above, version 0.4.1 of the operator was used. Look
 into the `cf_operator` section of the top-level `def.bzl` file to find
 the version of the operator validated against the current kubecf
 master.
+
+**Note:**
+> The above `helm install` will generate many controllers spread over multiple pods inside the `cfo` namespace.
+> Most of these controllers run inside the `cf-operator` pod.
+>
+> The `global.operator.watchNamespace=kubecf` path tells the
+controllers to watch for CRD´s instances into the `kubecf` namespace.
+>
+> The cf-operator helm chart will generate the `kubecf` namespace during installation, and eventually one of the
+controllers will use a webhook to label this namespace with the `cf-operator-ns` key.
+>
+> If the `kubecf` namespace is deleted, but the operators are still running, they will no longer
+know which namespace to watch. This can lead to problems, so make sure you also delete the pods
+inside the `cfo` namespace, after deleting the `kubecf` namespace.
 
 Note how the namespace the operator is installed into (`cfo`) differs
 from the namespace the operator is watching for deployments (`kubecf`).
@@ -82,7 +96,7 @@ This again uses helm and a released helm chart.
 
 ### Deployment and Tear-down
 
-```
+```shell
 helm install --name kubecf \
      --namespace kubecf \
      https://scf-v3.s3.amazonaws.com/scf-3.0.0-82165ef3.tgz \
@@ -104,9 +118,9 @@ cf api --skip-ssl-validation "https://api.$(minikube ip).xip.io"
 
 # Copy the admin cluster password.
 admin_pass=$(kubectl get secret \
-	      --namespace kubecf kubecf.var-cf-admin-password \
-	      -o jsonpath='{.data.password}' \
-	      | base64 --decode)
+        --namespace kubecf kubecf.var-cf-admin-password \
+        -o jsonpath='{.data.password}' \
+        | base64 --decode)
 
 # Use the password from the previous step when requested.
 cf auth -u admin -p "${admin_pass}"
@@ -130,8 +144,8 @@ By default, the cluster is exposed through its Kubernetes services.
 
 To use the NGINX ingress instead, it is necessary to:
 
-  - Install and configure the NGINX Ingress Controller.
-  - Configure Kubecf to use the ingress controller.
+- Install and configure the NGINX Ingress Controller.
+- Configure Kubecf to use the ingress controller.
 
 This has to happen before deploying kubecf.
 


doc/dev/general.md
```diff
@@ -104,9 +118,9 @@ cf api --skip-ssl-validation "https://api.$(minikube ip).xip.io"
 
 # Copy the admin cluster password.
 admin_pass=$(kubectl get secret \
-	      --namespace kubecf kubecf.var-cf-admin-password \
```
f0rmiga: Nice catch, linter. :)

doc/dev/general.md
```diff
@@ -49,20 +49,34 @@ but only generic tools, i.e. `kubectl` and `helm`.
 the chosen foundation, and assuming that the cluster does not come
 with Helm Tiller pre-installed.
 
-### Deployment and Tear-down
+### Deployment and Tear-down - cf-operator
```
f0rmiga: Since this is a subsection of `cf-operator`, it doesn't make sense to append it again here. Same for the `kubecf` one.

doc/dev/general.md
```diff
@@ -49,20 +49,34 @@ but only generic tools, i.e. `kubectl` and `helm`.
...
+     https://s3.amazonaws.com/cf-operators/helm-charts/cf-operator-v0.4.2-147.gb88e4296.tgz
 ```
 
 In the example above, version 0.4.1 of the operator was used. Look
 into the `cf_operator` section of the top-level `def.bzl` file to find
 the version of the operator validated against the current kubecf
 master.
 
+**Note:**
+> The above `helm install`, will generate two controllers into `cfo` namespace. The first
```
f0rmiga: There should be no comma before `will`.

This sentence can quickly become out-of-date. How about stating that there will be many controllers divided into multiple pods? cf-operator itself is not a single controller.

doc/dev/general.md
```diff
@@ -49,20 +49,34 @@ but only generic tools, i.e. `kubectl` and `helm`.
...
+controller is the `cf-operator`, the second one is the `quarks-job`.
+>
+> The `global.operator.watchNamespace=kubecf` path, tells both
```
f0rmiga: Same as above - there should be no comma before `tells` (the subject and predicate should not be separated by a comma).

doc/dev/general.md
```diff
@@ -49,20 +49,34 @@ but only generic tools, i.e. `kubectl` and `helm`.
...
+controllers to watch for CRD´s instances into the `kubecf` namespace.
+>
+> The cf-operator will create the `kubecf` namespace, and use a webhook to label the namespace
```
f0rmiga: Unnecessary comma before `and`.

doc/dev/general.md
```diff
@@ -49,20 +49,34 @@ but only generic tools, i.e. `kubectl` and `helm`.
...
+with the `cf-operator-ns` key.
+>
+> If the `kubecf` namespace is deleted, but the operators still running, they will not longer
```
f0rmiga: `not longer` -> `no longer`

doc/dev/general.md
```diff
@@ -49,20 +49,34 @@ but only generic tools, i.e. `kubectl` and `helm`.
...
+know which namespace to watch. This can lead to problems, so make sure you delete also the operators
```
f0rmiga: `delete also` -> `also delete`

f0rmiga: Maybe use `pods` instead of `operators` in this whole phrase?

doc/dev/general.md
edwardstudy: I know what's your meaning, but not sure if it's precise. 

Because the cf-operator does not create watched namespace and its service account should not have that permission to create namespaces. That's helm does this namespace creation.

doc/dev/general.md
```diff
@@ -49,20 +49,34 @@ but only generic tools, i.e. `kubectl` and `helm`.
...
+inside the `cfo` namespace, after deleting the `kubecf` namespace.
```
edwardstudy: I think it's talking about cleanup. About recover, we can restart cf-operator after recreating this namespace. It's also a thing that dev want to do.

doc/dev/general.md
qu1queee: fixed

doc/dev/general.md
qu1queee: fixed

doc/dev/general.md
qu1queee: fixed

doc/dev/general.md
qu1queee: fixed and good catch @edwardstudy

doc/dev/general.md
qu1queee: fixed

doc/dev/general.md
qu1queee: fixed

doc/dev/general.md
qu1queee: not sure what you meant. But this is only for clean-up.

doc/dev/general.md
```diff
@@ -55,14 +55,28 @@ with Helm Tiller pre-installed.
...
 ```
 
 In the example above, version 0.4.1 of the operator was used. Look
 into the `cf_operator` section of the top-level `def.bzl` file to find
 the version of the operator validated against the current kubecf
 master.
 
+**Note:**
+> The above `helm install` will generate many controllers spread over multiple pods inside the `cfo` namespace.
+> Most of this controllers run inside the `cf-operator` pod.
```
f0rmiga: `Most of this` -> `Most of these`

doc/dev/general.md
```diff
@@ -55,14 +55,28 @@ with Helm Tiller pre-installed.
...
+>
+> The `global.operator.watchNamespace=kubecf` path tells the
+controllers to watch for CRD´s instances into the `kubecf` namespace.
+>
+> The cf-operator helm chart will generate the `kubecf` namespace during installation, and eventually one of the
+controllers will use a webhook to label this namespace with the `cf-operator-ns` key.
+>
+> If the `kubecf` namespace is deleted, but the operators still running, they will no longer
```
f0rmiga: `operators still running` -> `operators are still running`

PR #179, opened 2019-11-21
Make scaling adjustable in values

c0d1ngm0nk3y:
Expose scaling information as helm values, so that the `instance_count` for specific `instance_groups` can be adjusted when installing `kubecf`.

## Description

The `sizing` section in the values has now values for each `instance_group`. If such a value is set during install, the `instance_count` will be adapted via op file.

Another option `sizing.single_az` was added to allow using several instances in only one az. The existing property `sizing.HA` has still the same meaning, but combines 2 aspects (`instance_count` 1 AND 1 az).

## Motivation and Context

The installation can be scaled in a custom way. Without this, there are only 2 options
1) `sizing.HA=false` - There is only one `az` and all `instance_count`s are one (minimal installation)
2) `sizing.HA=true` -  The settings for `az`s and  `instance_count`s  are set by the used `cf-deployment`

## How Has This Been Tested?

Tested by manually installing `kubecf` in dfferent combinations using the helm values.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


deploy/helm/kubecf/values.yaml
```diff
@@ -41,6 +41,21 @@ releases:
 
 sizing:
   HA: false
+  single_az: true
+  adapter: ~
```
f0rmiga: Could you move the instance group names to be under the `sizing.instance_groups` key? Then, instead of adding the operation to each instance group file, we can have a single file for dealing with sizing that does:

```
{{- range $name, $instances := .Values.sizing.instance_groups }}
{{- if $instances }}
- type: replace
  path: /instance_groups/name={{ $name }}/instances
  value: {{ $instances }}
{{- end }}
{{- end }}
```

You will also need to rename the snake-case instance group keys here to kebab-case, so the instance group matches its name in the manifest.

Since you are already touching this, could you merge the contents of `single_availability.yaml` and `single_az` to that same `sizing.yaml` file, please?

mook-as: Do we need to worry about SCF compatibility here? Because `sizing.x.count` is the old way.  I guess the original commit doesn't actually implement that either, so we're free to go with whichever makes more sense for kubecf in isolation?

f0rmiga: We already have many differences from SCF. I see an opportunity to improve the values.yaml to make more sense with kubecf, given we will already need some tool or document to instruct the migration and changes. @mook-as feel free to suggest something else that may make more sense.

Maybe:
```yaml
sizing:
  instance_groups:
    nats:
      instances: 3
```

I think it gives us enough room to add subkeys in between as needed in the future.

c0d1ngm0nk3y: > I think it gives us enough room to add subkeys in between as needed in the future.

I think that is a fair point. But is it realistic to have additional subkeys in the section `sizing`? But there is probably no downside with that. At least it is very clear/self documenting.

What about the `single_az: true` - any objections in having this as additional setting?

c0d1ngm0nk3y: > You will also need to rename the snake-case instance group keys here to kebab-case, so the instance group matches its name in the manifest.

That was my first approach, but `-` is not allowed in a property name, is it? I will try it out again.

bkrannich: @c0d1ngm0nk3y: With these parameter names, I'm always thinking do we approach them "in the right direction", so I'm going to throw in `multi_az: false/true` as an alternative.

c0d1ngm0nk3y: To be honest, I haven't put too much thoughts in this. I wanted to start the PR as soon as possible to get some opinions. I think the main reason for putting it that way, was the fact that we actually don't do anything to assure multiple `az`s, but remove them in some cases. So we rely on what is in `cf-deployment`. But

a) this is also true for the property `HA` - `instance_count` is only set to 1 if `false`
b) it is probably fair to assume that `cf-deployment` will always be with multiple `az`s

btw I am not that convinced about the property `sizing.HA`, because I think it tackles two aspects at once (no `az`s and `instance_count` 1). After talking with @loewenstein - how about a setting like `sizing.minimal` which has dominance, but otherwise the values are used and the defaults from `cf-deployment`?

f0rmiga: I think we are doing wrong by controlling AZs as part of the sizing. If we want to do better, I agree to have `multi_az: false` as default in the `values.yaml` - note that it wouldn't be a key under `sizing`. I was tempted by the `sizing.minimal`, but it only makes the naming better rather than solving the problem of having 2 things being controlled by a single property. We can also make the sizing part more clear:

```yaml
multi_az: false
high_availability: false

# Sizing takes precedence over the high_availability property. I.e. setting the instance count
# for an instance group greater than 1 will make it highly available.
sizing:
  adapter:
    instances: ~
  nats:
    instances: ~
```

c0d1ngm0nk3y: Sounds good to me, we have 3 cases

1. The `instance` value is set, so this should be used
2. The `instance` value is not set and `high_availability=false`, so the `instance_count` is changed to 1
3. The `instance` value is not set and `high_availability=true`, so the `instance_count` from `cf-deployment` is used

c0d1ngm0nk3y:
I have uploaded another version.  What puzzles me a bit is the inconsistence in my eyes: On the one we say `multi_az` should not be part of `.Values.sizing`, but on the other we want only one ops file `sizing.yaml` with handles the removal of the azs if needed.

f0rmiga:
> I have uploaded another version. What puzzles me a bit is the inconsistence in my eyes: On the one hand we say `multi_az` should not be part of `.Values.sizing`, but on the other hand we want only one ops file `sizing.yaml` with handles the removal of the azs if needed.

After the latest suggestion for separating AZ and sizing, it makes sense to have all the AZ stuff on its own file.

c0d1ngm0nk3y:
I had to upload another commit fixing some whitespaces (I hate `yaml`), but the manual tests were OK now.

c0d1ngm0nk3y:
> > I have uploaded another version. What puzzles me a bit is the inconsistence in my eyes: On the one hand we say `multi_az` should not be part of `.Values.sizing`, but on the other hand we want only one ops file `sizing.yaml` with handles the removal of the azs if needed.
> 
> After the latest suggestion for separating AZ and sizing, it makes sense to have all the AZ stuff on its own file.

Done

f0rmiga:
I rebased to resolve the merge conflict.

f0rmiga:
I pushed a few commits to get some fixes while I was testing this PR.

PR #204, opened 2019-11-27
Fixed routing-api IP that is used to register itself

f0rmiga:
## Description

Fixes the `routing-api` IP that is used to register itself to provide `api.((system_domain))/routing`.

## Motivation and Context

The `routing-api` job registers itself via code, and it uses the `-ip` command-line flag for the destination IP. This patch will use the `POD_IP` environment variable that we set via the ops-file, and k8s will render it via https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments.

## How Has This Been Tested?

Got a bash session to the router container and called `/var/vcap/jobs/gorouter/bin/retrieve-local-routes` and checked that the `address` for the `api.((system_domain))/routing` was set to the pod IP. Also, the command `cf router-groups` should succeed (previously failed). The CATS `tcp_routing` suite will eventually test this.

## Types of changes

## Checklist:

### DIFF
--- a/bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
+++ b/bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
@@ -5,15 +5,18 @@ set -o errexit -o nounset
 target="/var/vcap/all-releases/jobs-src/routing/routing-api/templates/bpm.yml.erb"
 
 PATCH=$(cat <<'EOT'
-@@ -11,7 +11,7 @@
+@@ -11,8 +11,10 @@
      - -timeFormat
      - rfc3339
      - -ip
 -    - <%= spec.ip %>
-+    - 0.0.0.0
++    - "$(POD_IP)"
      <% if p("routing_api.auth_disabled") == true %>- -devMode <% end %>
++    env:
++      POD_IP: 0.0.0.0 # Set by k8s using an ops-file with cf-operator.
 
      hooks:
+       pre_start: /var/vcap/jobs/routing-api/bin/bpm-pre-start
 EOT
 )
 
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -155,6 +155,12 @@
 - type: replace
   path: /instance_groups/name=api/jobs/name=routing-api/properties/quarks?
   value:
+    envs:
+    - name: POD_IP
+      valueFrom:
+        fieldRef:
+          apiVersion: v1
+          fieldPath: status.podIP
     ports:
     - name: routing-api
       protocol: TCP


bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
```diff
@@ -5,15 +5,18 @@ set -o errexit -o nounset
 target="/var/vcap/all-releases/jobs-src/routing/routing-api/templates/bpm.yml.erb"
 
 PATCH=$(cat <<'EOT'
-@@ -11,7 +11,7 @@
+@@ -11,8 +11,10 @@
      - -timeFormat
      - rfc3339
      - -ip
 -    - <%= spec.ip %>
```
loewenstein: Couldn't we just provide the Pod IP as `spec.ip` and not need a patch file at all?

bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
```diff
@@ -5,15 +5,18 @@ set -o errexit -o nounset
...
-+    - 0.0.0.0
++    - "$(POD_IP)"
      <% if p("routing_api.auth_disabled") == true %>- -devMode <% end %>
++    env:
++      POD_IP: 0.0.0.0 # Set by k8s using an ops-file with cf-operator.
```
loewenstein: Not sure I understand why we add an env to the bpm process.

bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
f0rmiga: No, because `spec.ip` is not available during the bpm rendering, which happens on a pod before the `api` pod is created.

bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
f0rmiga: It's a bug on cf-operator right now. Should go away after it's fixed there.

bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
f0rmiga: Note that the statefulsets depend on the bpm rendering...

bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
```diff
@@ -5,15 +5,18 @@ set -o errexit -o nounset
...
 
 PATCH=$(cat <<'EOT'
-@@ -11,7 +11,7 @@
+@@ -11,8 +11,10 @@
      - -timeFormat
      - rfc3339
      - -ip
 -    - <%= spec.ip %>
-+    - 0.0.0.0
++    - "$(POD_IP)"
```
mook-as: Huh, I wonder what syntax this is. Seems to work, though:
> `/var/vcap/packages/routing-api/bin/routing-api -config /var/vcap/jobs/routing-api/config/routing-api.yml -port 3000 -logLevel info -timeFormat rfc3339 -ip 172.17.0.23`

f0rmiga: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments

k8s sorcery!

f0rmiga:
This fix was required by https://github.com/SUSE/kubecf/issues/206.

bosh/releases/pre_render_scripts/api/routing-api/bpm/patch_bpm.sh
loewenstein: Understood. However, this is a patch we cannot get rid of, right? That's pretty unfortunate.

Couldn't we parse the BPM for `<%= spec.ip %>` (match would have to be more fuzzy I suppose to catch all cases) and replace it by `POD_IP`? I pretty much mean exactly what you are already doing in this PR, we should just not require patching the bpm.yml file.

f0rmiga: I think we can get rid of this patch in the future by opening a PR on routing-release adding a property to control if it uses `spec.ip` or `$(POD_IP)`. I think this is more clear than cf-operator doing some magic on it.

loewenstein: Mmh, why should the Bosh release know about POD_IP? And the "magic" would be pretty much in line with what the cf-operator does, "playing" Bosh director, wouldn't it?

f0rmiga: `$(POD_IP)` works here because it is a BPM argument that becomes a container argument in the k8s pod. It's not valid if `spec.ip` is used to set a BPM environment variable, which I believe I saw in the past. So, cf-operator would need to really do some magic and I don't think this kind of magic is cf-operator's responsibility. There are other releases that support more k8s specifics, and my suggestion aligns with it. Also, it becomes very clear and less error-prone.

PR #210, opened 2019-11-28
Bumps, enhancements and fixes

f0rmiga:
## Description

- Bumped various external dependencies of the project.
- Enhanced and fixed `yamllint` rules, targets and scripts.

## Motivation and Context

To keep up-to-date with the external dependencies.

## How Has This Been Tested?

By running the bumped external dependencies. Most of them are tested by CI.
The changes to `kind` were tested locally.

## Types of changes

## Checklist:


dev/kind/start.sh
```diff
@@ -9,15 +9,13 @@ function cluster_exists {
...
+  "${KIND}" create cluster \
+    --name "${CLUSTER_NAME}" \
+    --image "kindest/node:${KUBERNETES_VERSION}"
 else
   echo "Kind is already started"
 fi
 
-# Use the kubeconfig file generated by kind.
-KUBECONFIG="$("${KIND}" get kubeconfig-path --name "${CLUSTER_NAME}")"
-export KUBECONFIG
```
mook-as: We no longer to use this kubeconfig? Why?

dev/kind/def.bzl
```diff
@@ -41,6 +44,9 @@ attrs = {
     "cluster_name": attr.string(
         mandatory = True,
     ),
+    "kubernetes_version": attr.string(
```
mook-as: The minikube branch uses `k8s_version`; please be consistent (or possibly change the minikube one to match).  I think in the minikube case it's also possible to override this with an environment variable that's set when `bazel` is executed… do the same here?

dev/kind/start.sh
f0rmiga: kind is doing like minikube.

mook-as: I take that to mean "it modifies your default kubeconfig to have a new context &c"? That's fine, then.

f0rmiga: Yep

PR #213, opened 2019-11-28
minikube: allow arbitrary extra options

mook-as:
## Description
This lets us pass in extra arguments to minikube (but does not hard code which, since that does not make sense outside of the local context).

## Motivation and Context
I want to set up a local docker registry mirror; this _can_ be done in minikube, but requires extra arguments.  

## How Has This Been Tested?
I now have extra options to use a docker registry mirror on minikube (using @f0rmiga's opensuse image, as is default in kubecf).

## Types of changes

## Checklist:

### DIFF
--- a/dev/minikube/start.sh
+++ b/dev/minikube/start.sh
@@ -3,6 +3,7 @@
 set -o errexit -o nounset
 
 if ! "${MINIKUBE}" status > /dev/null; then
+  # shellcheck disable=SC2086
   "${MINIKUBE}" start \
     --kubernetes-version "${K8S_VERSION}" \
     --cpus "${VM_CPUS}" \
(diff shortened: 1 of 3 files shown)


dev/minikube/start.sh
```diff
@@ -11,7 +12,8 @@ if ! "${MINIKUBE}" status > /dev/null; then
     --iso-url "${ISO_URL}" \
     ${VM_DRIVER:+--vm-driver "${VM_DRIVER}"} \
     --extra-config=apiserver.runtime-config=settings.k8s.io/v1alpha1=true \
-    --extra-config=apiserver.enable-admission-plugins=MutatingAdmissionWebhook,PodPreset
+    --extra-config=apiserver.enable-admission-plugins=MutatingAdmissionWebhook,PodPreset \
+    ${MINIKUBE_EXTRA_OPTIONS:-}
```
andreas-kupries: Please update the associated documentation files (`doc/dev/minikube.md` - Example of registry mirror would work, would it not ? )

mook-as: Thanks for the pointer to the location! Fixed in a5dde5de7985b72fcadfb22621c2ba0e087ce7f1

andreas-kupries: LGTM.

PR #218, opened 2019-11-29
Create kubecf+operator bundle files as an artifact

mook-as:
## Description
This makes it easier for QA to install the matching versions.

## Motivation and Context
For QA, they'd like to be sure they get the correct cf-operator for a given kubecf build without having to dig through the sources to find the file.  This is related to (but not the same as) #124.

Fixes jsc#CAP-1029.

## How Has This Been Tested?
Locally ran CI's [build.sh] and checked that the expected files are generated.  Will also trigger CI.

[build.sh]: https://github.com/SUSE/kubecf/blob/03408ebb8e373449f4ed896c6fc129147e46a8fd/.gitlab/pipelines/stages/build/build.sh

## Types of changes

## Checklist:


.gitlab/pipelines/stages/publish.sh
```diff
@@ -3,5 +3,7 @@
 set -o errexit -o nounset
 
 mc config host add "s3" "https://s3.amazonaws.com" "${AWS_ACCESS_KEY}" "${AWS_SECRET_KEY}"
-mc cp output/kubecf-*.tgz "s3/scf-v3"
-echo "URL: https://scf-v3.s3.amazonaws.com/$(basename output/kubecf-*.tgz)"
+for file in output/kubecf-*.tgz ; do
```
f0rmiga: This is an unnecessary change. `kubecf-*` should match only one file under `output`. If we get more than one, something is wrong.

mook-as: Ah, yeah, I had made the bundle be called `kubecf-bundle-blah`, but then had to change it to `bundle-blah` because actually deploying it got annoying. Will fix this.

mook-as: Fixed in cb180aa56f8c159344d7888cb62c3976dae57d0d

PR #219, opened 2019-11-30
Bump to cf-deployment 12.4.0

### LINKED ISSUE #49: Bump to cf-deployment v12.4.0
Current version 8 is quite old.

Bump to cf-deployment v12.4.0

**Tasks**
- [ ] update the docker image release pipeline to contain this version
- [ ] update kubecf

viovanov:
## Description
- bumps cf-deployment to 12.4 and implements fixes so everything still works.
- bumps cf-operator

## Motivation and Context
Fixes #49 

## How Has This Been Tested?
Ran an app and it worked. Used cf-operator `cf-operator-v0.4.2-187.gee3ea37b.tgz`.

-

## Types of changes

## Checklist:


loewenstein:
@viovanov Do we really want to bump to v12.4.0 while v12.14.0 is already out there? We should at least have a follow-up, shouldn't we?

bosh/releases/pre_render_scripts/uaa/uaa/jobs/patch_pre-start.sh
```diff
@@ -6,39 +6,20 @@ target="/var/vcap/all-releases/jobs-src/uaa/uaa/templates/bin/pre-start.erb"
...
+@@ -32,9 +32,8 @@
+     <% end %>
 
- update_ca_certificate
+     log "Trying to run update-ca-certificates..."
+-    # --certbundle is an undocumented flag in the update-ca-certificates script
+-    # https://salsa.debian.org/debian/ca-certificates/blob/master/sbin/update-ca-certificates#L53
+-    timeout --signal=KILL 180s /usr/sbin/update-ca-certificates -f -v --certbundle "$(basename "${OS_CERTS_FILE}")"
++    timeout --signal=KILL 180s /usr/sbin/update-ca-certificates -f -v
++    mv /var/lib/ca-certificates/ca-bundle.pem /etc/ssl/certs/"$(basename "${OS_CERTS_FILE}")"
```
mook-as: Wouldn't this fail on Ubuntu (where it's still `/etc/ssl/…`)?

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -121,8 +121,7 @@
               - /bin/sh
               - -c
               - >
-                printf "GET /healthz HTTP/1.0\r\n\r\n" |
-                nc -q 2 -U /var/vcap/data/cloud_controller_ng/cloud_controller.sock |
+                curl -sI --unix-socket /var/vcap/data/cloud_controller_ng/cloud_controller.sock http:/healthz |
```
mook-as: :tada: we finally have `curl --unix-socket`

Doesn't that mean we can just use `curl --fail` and not need the pipe to `grep`? And then we can drop the `/bin/sh` too… at which point long args seem better.

f0rmiga:
Great! I like that this PR cleans up some stuff that I wanted to have cleaned up, like a different version for CAPI, the UAA patch and moving to SLE15.

We still want to have the other releases to use SLE15 instead of openSUSE.

deploy/helm/kubecf/assets/operations/instance_groups/router.yaml
```diff
@@ -39,3 +44,14 @@
...
+# Trust the diego_instance_identity_ca certificate - trusting its CA
+# is insufficient with the cf-operator 
+- type: replace
+  path: /instance_groups/name=router/jobs/name=gorouter/properties/router/ca_certs
+  value: |
+    ((diego_instance_identity_ca.ca))
+    ((cc_tls.ca))
+    ((uaa_ssl.ca))
+    ((network_policy_server_external.ca))
+    ((diego_instance_identity_ca.certificate))
```
mook-as: Do we need to trust _both_ the CA and the cert?

deploy/helm/kubecf/values.yaml
```diff
@@ -21,24 +21,25 @@ releases:
...
     stemcell:
+      os: opensuse-42.3
       version: 36.g03b4653-30.80-7.0.0_348.gc8fb3864
-  capi:
-    version: 1.83.0
-    stemcell:
-      version: 36.g03b4653-30.80-7.0.0_360.g0ec8d681
   cf-acceptance-tests:
     stemcell:
+      os: opensuse-42.3
```
mook-as: IIRC, CATS only had a custom stemcell because the default one didn't have a CATS image.  If we're switching to a newer stemcell we can probably drop the custom stemcell (just delete the block, the helm templating should be smart enough to pick up the default one).

Please also check the other releases to see if they can be switched to the default stemcell.

gaktive:
scf will be targeting cf `12.12` shortly so ideally kubecf would eventually be there or further. Let me know if we need another story to do that bump in the backlog.

f0rmiga:
> scf will be targeting cf `12.12` shortly so ideally kubecf would eventually be there or further. Let me know if we need another story to do that bump in the backlog.

@gaktive Please, add another story to the backlog to bump it to the latest version. When we pick the ticket it maybe even further than 12.14. We probably want to keep on 12.x for now though.

gaktive:
https://github.com/SUSE/kubecf/issues/229 filed

deploy/helm/kubecf/assets/operations/instance_groups/router.yaml
aduffeck: No, this is only required temporarily to work around a [bug in the operator](https://www.pivotaltracker.com/story/show/170044974).

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -231,11 +225,12 @@
           readiness:
             httpGet:
               port: 4002
+              scheme: HTTPS
     post_start:
       condition:
         exec:
           # policy-server doesn't support HTTP HEAD requests
-          command: [curl, --fail, --silent, http://localhost:4002/]
+          command: [curl, -k, --fail, --silent, https://localhost:4002/]
```
f0rmiga: [nit] Use the long version instead: `--insecure`. Whenever possible, use the long version for the command line flags.

deploy/helm/kubecf/assets/operations/instance_groups/router.yaml
f0rmiga: Then we need a TODO here.

fargozhu:
> scf will be targeting cf `12.12` shortly so ideally kubecf would eventually be there or further. Let me know if we need another story to do that bump in the backlog.
There's already one that covers the bump to 12.4 #49

deploy/helm/kubecf/values.yaml
viovanov: @mook-as this sounds like it should be another ticket
@aduffeck - did you try with the default and it failed?

aduffeck: @mook-as @viovanov I switched the CATS to the SLE15 based stemcell in 63c4f76. The remaining opensuse stemcells are currently required and will be tackled separately later on.

PR #235, opened 2019-12-04
Prepended all k8s resources with the kubecf installation name

f0rmiga:
## Description

Fixes https://github.com/SUSE/kubecf/issues/192.

## Motivation and Context

Some of the k8s resources were not being prepended with the kubecf installation name. To avoid naming collisions, we prepend the resources.

## How Has This Been Tested?

Deployed kubecf with HA, Eirini and Ingress enabled.

## Types of changes

## Checklist:


mook-as:
So we're trying to support installing multiple kubecf deployments into the same namespace? (A lot of these resources are namespaced, so not doing this is fine as long as we don't support multiple in the same ns.)

f0rmiga:
@mook-as We are not in the stage of actually _doing_ multiple kubecf in the same namespace, but this is definitely a pre-requisite.

PR #237, opened 2019-12-05
Enabled DB encryption validation

f0rmiga:
## Description

In the early stages of the development of kubecf, we had to disable the DB encryption validation as it was failing.

Fixes https://github.com/SUSE/kubecf/issues/24.

## Motivation and Context

We should do the validation, which is the default behaviour with cf-deployment.

## How Has This Been Tested?

Deployed locally and the `api` instance group doesn't fail anymore.

## Types of changes

## Checklist:


f0rmiga:
@mook-as I missed a bit of the code. We actually need to perform the migration in the pre-start. It's fine now that the api instance group waits for the database to be ready.

PR #239, opened 2019-12-06
Removed dev/kubecf/values.yaml

f0rmiga:
## Description

The file `dev/kubecf/values.yaml` is currently empty and not needed anymore.

## Motivation and Context

We only had one value in this file before, which was `deployment_name`. Since we removed this property, this file has no use anymore.

## How Has This Been Tested?

N/A.

## Types of changes

## Checklist:

### DIFF
--- a/dev/kubecf/.gitignore
+++ b/dev/kubecf/.gitignore
@@ -1,2 +1,1 @@
 *values.yaml
-!values.yaml


f0rmiga:
@mook-as I don't think so. The docs state that any `*values.yaml` under that folder will be used by Helm.

PR #243, opened 2019-12-10
Add Drone CI

f0rmiga:
## Description

Added initial Drone CI pipelines and scripts.
It includes scripts for deploying Drone CI server and runners.
Instead of using k3s with Drone, we are moving to kind. It will allow choosing the k8s version instead of depending on the version that k3s ships.

## Motivation and Context

Drone CI has a lightweight server that we can deploy and connect dedicated runners using dedicated hardware. It also allows running privileged containers on those dedicated runners.

## How Has This Been Tested?

Deployed the server and connected a dedicated runner to it.

I'm consistently getting no flakes on CATS, which is an improvement over Gitlab right now:

```
Ran 86 of 213 Specs in 1074.933 seconds
SUCCESS! -- 86 Passed | 0 Failed | 0 Flaked | 1 Pending | 126 Skipped
```

## Types of changes

## Checklist:


.drone/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+  path: kubecf
+
+clone:
+  disable: true
+
+# Dockerfile source - https://github.com/f0rmiga/bazel-docker-image.
+.bazel_image: &bazel_image thulioassis/bazel-docker-image:1.2.1
+
+steps:
+- name: clone
```
mook-as: Why do we need custom cloning? It doesn't appear that we are doing anything here that drone's default cloning doesn't handle?

If we need it, please comment on why.

.drone/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+workspace:
+  base: /home/runner/builds
+  path: kubecf
+
+clone:
+  disable: true
+
+# Dockerfile source - https://github.com/f0rmiga/bazel-docker-image.
+.bazel_image: &bazel_image thulioassis/bazel-docker-image:1.2.1
```
mook-as: Would it be better if we have a section of junk like this?

```suggestion
.anchors:
  bazel_image: &bazel_image thulioassis/bazel-docker-image:1.2.1
```

.drone/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+steps:
+- name: clone
+  image: *bazel_image
+  commands:
+  - git clone --depth=2 --branch="${DRONE_BRANCH}" "${DRONE_REMOTE_URL}" .
+  - git checkout "${DRONE_COMMIT}"
+- name: lint
+  image: *bazel_image
+  commands:
+  - ./dev/linters/shellcheck.sh
```
mook-as: Style question:

Would it be better if we had separate steps for these?
```yaml
- name: lint:shellcheck
  image: *bazel_image
  commands: [ dev/linters/shellcheck.sh ]
- name: lint:yaml
  image: *bazel_image
  commands: [ dev/linters/yamllint.sh ]
- name: lint:helm
  image: *bazel_image
  commands: [ dev/linters/helmlint.sh ]
```

Mostly wanting this so that if we need to point somebody at a failure, we can do so showing shorter, more relevant logs.

.drone/deploy/runner/start.sh
```diff
@@ -0,0 +1,32 @@
...
+printf "RPC secret: "
+read -r rpc_secret
+stty echo
+printf "\\n"
+if [ -z "${rpc_secret}" ]; then
+  >&2 echo "The RPC secret cannot be empty"
+  exit 1
+fi
+
+docker run \
```
mook-as: Should probably give a `--name` here since running multiple drone runners on the same machine isn't the best idea (especially given `--restart always`).

.drone/deploy/server/image/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+
+set -o errexit -o nounset
+
+export DRONE_SERVER_PORT=":${PORT}"
+
+/bin/drone-server "${@}"
```
mook-as:
```suggestion
exec /bin/drone-server "${@}"
```
There's no need for the `sh` process to stick around.

.drone/deploy/server/image/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
```
mook-as: Add a comment that this script exists only to translate `$PORT` to `$DRONE_SERVER_PORT` so that we can run it in Heroku?

.drone/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+  commands:
+  - .drone/pipelines/default/steps/build/build.sh
+- name: test
+  image: *bazel_image
+  volumes:
+  - name: docker
+    path: /var/lib/docker
+  privileged: true
+  commands:
+  - ./.drone/pipelines/default/steps/test/test.sh
```
mook-as: I'd really like to split this into multiple steps, so that it's easier to know where the failures are.  We can make things [detached](https://docker-runner.docs.drone.io/configuration/steps/#detach) if necessary (but it doesn't look like anything in that script actually needs to run in the background?)

rules/external_binary/BUILD.bazel
```diff
@@ -1 +1,24 @@
...
+    "shellcheck",
+    "yaml2json",
+    "containerd",
+    "containerd-shim",
+    "ctr",
+    "docker",
+    "dockerd",
+    "docker-init",
+    "docker-proxy",
+    "runc",
```
mook-as: Sort?

f0rmiga: I actually wanted to keep the docker binaries groups. I'll add a comment.

.drone/config.yaml
f0rmiga: Since I moved to use the root user in the image, all the file permissions related to the repo + bazel went away. I also don't think Drone expects us to use any non-root user.

.drone/config.yaml
f0rmiga: Docker itself needed to run in the background. My trials with making docker process detached didn't work well. Using the `host` network mode proved very unstable for a reason that I couldn't easily figure out. My current approach is to get rid of the inner docker daemon and use the host docker daemon, given that the CI runner itself is of type docker.

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+bazel run //testing/acceptance_tests
+
+cf_acceptance_tests_pod_name() {
+  kubectl get pods --namespace "${KUBECF_NAMESPACE}" --output name 2> /dev/null | grep "acceptance-tests"
+}
+
+# Wait for cf-acceptance-tests to start.
+wait_for_cf_acceptance_tests_pod() {
+  local timeout="300"
+  until kubectl get pods --namespace "${KUBECF_NAMESPACE}" --output name 2> /dev/null | grep --quiet "acceptance-tests" || [[ "$timeout" == "0" ]]; do sleep 1; timeout=$((timeout - 1)); done
```
mook-as: If `cf_acceptance_tests_pod_name()` uses `--ignore-not-found` we can just call it and check for empty.

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+# shellcheck disable=SC1091
+source ".drone/pipelines/default/runtime/binaries.sh"
+# shellcheck disable=SC1091
+source ".drone/pipelines/default/runtime/config.sh"
+
+# Trigger cf-acceptance-tests.
+bazel run //testing/acceptance_tests
+
+cf_acceptance_tests_pod_name() {
+  kubectl get pods --namespace "${KUBECF_NAMESPACE}" --output name 2> /dev/null | grep "acceptance-tests"
```
mook-as: Don't we have a selector we can use here?

dev/linters/shellcheck.sh
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
-set -o errexit
+set -o errexit -o nounset
+
+workspace=$(bazel info workspace)
```
mook-as: Nit: Quote?

rules/external_binary/BUILD.bazel
```diff
@@ -1 +1,17 @@
 package(default_visibility = ["//visibility:public"])
+
+load(":def.bzl", "binary_location")
+
+[binary_location(
+    name = "{}_binary_location".format(name).replace("-", "_"),
```
mook-as: Would it make sense to just use the name as-is (without the `_binary_location` suffix)?

.drone/config.star
```diff
@@ -0,0 +1,171 @@
...
+
+    pipeline["volumes"] = volumes()
+
+    return pipeline
+
+
+BAZEL_CACHE_VOLUME_NAME = "bazel-cache"
+DOCKER_SOCK_VOLUME_NAME = "docker-sock"
+DOCKER_SOCK_PATH = "/var/run/docker.sock"
+KUBE_VOLUME_NAME = "kube"
```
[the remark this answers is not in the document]
f0rmiga: No, the whole kube volume.

.drone/deploy/server/deploy.sh
```diff
@@ -0,0 +1,16 @@
...
+
+dockerfile_path_real=$(readlink "{dockerfile_path}")
+dockerfile_dirname=$(dirname "${dockerfile_path_real}")
+
+cd "${dockerfile_dirname}"
+
+heroku container:push \
+  --recursive \
+  --app "{app_name}" \
+  --arg "DRONE_IMAGE=drone/drone:{drone_image_version}@sha256:{drone_image_sha256},DRONE_CONVERT_STARLARK_IMAGE=drone/drone-convert-starlark@sha256:{drone_convert_starlark_image_sha256}"
```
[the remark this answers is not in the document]
f0rmiga: I got hit by it...

.drone/deploy/server/image/run.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# This script wraps the original drone-server executable but maps the provided $PORT environment
+# variable to the $DRONE_SERVER_PORT variable.
+
+set -o errexit -o nounset
+
+# DRONE_CONVERT_PLUGIN_SECRET is required, but in this case it is running inside the container with
+# no public access, so we don't need a real secret.
+convert_starlark_secret="dont_need_to_be_secret"
```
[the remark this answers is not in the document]
f0rmiga: I really don't care about generating a secret here. Port 3000 is not exposed and is used for internal communication only.

.drone/pipelines/default/steps/deploy/kubecf.sh
```diff
@@ -0,0 +1,55 @@
...
+source ".drone/pipelines/default/runtime/config.sh"
+# shellcheck disable=SC1091
+source ".drone/pipelines/default/steps/build/output_chart.sh"
+
+node_ip=$(kubectl get node kubecf-control-plane \
+  --output jsonpath='{ .status.addresses[?(@.type == "InternalIP")].address }')
+system_domain="${node_ip}.nip.io"
+
+# Create a values.yaml for the test.
+values="$(mktemp -t values_XXXXXXXX.yaml)"
```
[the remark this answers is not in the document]
f0rmiga: I'll remove the `mktemp`. It doesn't make sense to have it in this context.

dev/linters/shellcheck.sh
f0rmiga: Don't need quote in variable declaration.

rules/external_binary/BUILD.bazel
f0rmiga: Maybe. I'll try it.

.drone/config.star
```diff
@@ -0,0 +1,171 @@
...
+            "refs/heads/test/**",
+            "refs/heads/master",
+        ],
+    }
+
+    steps = []
+
+    steps.append(step(
+        name = "external-binaries",
+        commands = ["bash {}/runtime/binaries.sh".format(DRONE_BASE_PATH)],
```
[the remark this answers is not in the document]
f0rmiga: This bash is unnecessary... I will make the script executable and remove the explicit `bash` call. This script is also sourced later by other scripts.

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
f0rmiga: I'd prefer not changing the logic of these scripts too much in this PR given that they already work and it's mostly a copy over from the Gitlab one. We can improve this script when we move it to be part of the bazel target itself.

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
f0rmiga: https://github.com/SUSE/kubecf/pull/243#discussion_r358526502

.drone/pipelines/default/steps/deploy/wait_kubecf.sh
```diff
@@ -0,0 +1,48 @@
...
+set -o errexit -o nounset -o pipefail
+
+# shellcheck disable=SC1091
+source ".drone/pipelines/default/runtime/binaries.sh"
+# shellcheck disable=SC1091
+source ".drone/pipelines/default/runtime/config.sh"
+
+echo "Waiting for the kubecf pods to be ready..."
+
+secret_name="${KUBECF_INSTALL_NAME}.with-ops"
```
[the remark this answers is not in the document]
f0rmiga: I don't think there is an action item for this one right now. We can switch when it's required (when we bump to a cf-operator that requires it).

.drone/config.star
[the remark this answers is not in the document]
mook-as: Thanks for getting rid of the explicit bash.

I'm thinking of making it be something like `commands = ["@@/runtime/binaries.sh"]`, and have `step()` take care of subbing `@@` (or even just `{}`) with `DRONE_BASE_PATH` (because we essentially do it everywhere and makes things harder to read).  Does that seem like a reasonable idea?

.drone/config.star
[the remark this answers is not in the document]
mook-as: But it holds just `~/.kube/config` and other config-stuff-for-kube?  What do you mean by "the whole kube volume", that's… separate from client configuration?

.drone/config.star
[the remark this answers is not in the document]
f0rmiga: I don't think it's the correct approach given that a step doesn't necessarily call a script under the Drone directory. An example is each linting step. We could eventually just call bazel for some of them.

.drone/config.star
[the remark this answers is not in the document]
f0rmiga: I see what you mean. I thought that calling it `KUBE_CONFIG_VOLUME_NAME` would confuse with only the `config` file.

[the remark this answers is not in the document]
f0rmiga: I amended the change to the previous commit.

.drone/config.star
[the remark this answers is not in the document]
mook-as: I guess I'll live with this for now…
I'm not saying we should prefix _all_ commands; just run a replace on it. But that might end up more confusing than helpful, so ¯\\\_(ツ)\_/¯.

.drone/deploy/runner/def.bzl
```diff
@@ -0,0 +1,52 @@
...
+            "{runner_image_sha256}": ctx.attr._runner_image_sha256,
+        },
+        is_executable = True,
+    )
+
+    runfiles = []
+
+    return [DefaultInfo(
+        executable = script,
+        runfiles = ctx.runfiles(files = runfiles),
```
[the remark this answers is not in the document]
f0rmiga: Yes... it is probably a left-over from something I was experimenting with. I'll remove it.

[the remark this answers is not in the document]
f0rmiga: Fixed in https://github.com/SUSE/kubecf/pull/243/commits/b6ab6b1777fe54d38b47a458baac0db5dd8c110f.

PR #244, opened 2019-12-10
Aduffeck/feature flags for tests

### LINKED ISSUE #234: Make the errand tests default to disabled (not smoke)
**Describe the bug**

Currently, CATS and brain are enabled by default. We should deploy them only when set on `values.yaml`.

In environments that don't need to run those, the overhead of downloading the images is big.
On the security side, brain-tests need a cluster-admin role that we should not create by default.

A good approach would be to have `testing` key on `values.yaml`:

```yaml
testing:
  smoke-tests:
    enabled: true
  acceptance-tests:
    enabled: false
  brain-tests:
    enabled: false
  sync-integration-tests:
    enabled: false
```

**To Reproduce**

Deploy kubecf and check that all the errands are present in the final manifest by default.

**Expected behavior**

By default, they should not be present in the final manifest.

**Additional context**

In production environments, only smoke-tests make sense.

aduffeck:
## Description
This pr adds feature flags for the different test suites (smoke, acceptance and brain) to the helm chart.

## Motivation and Context
Fixes #234 

## How Has This Been Tested?

I deployed kubecf with each test enabled/disabled and verified that the according instance groups and service accounts existed/did not exist.

## Types of changes

## Checklist:

- [x] My code has security implications (removes a superfluous service account in the default settings)


deploy/helm/kubecf/values.yaml
```diff
@@ -181,6 +181,16 @@ features:
         password: ~
         username: ~
 
+# Enable or disable instance groups for the different test suites.
+# Only smoke tests should be run in production environments.
+testing:
+  acceptance_tests:
```
f0rmiga: Since there are many `acceptance-tests` for CF components ([check here](https://github.com/cloudfoundry?utf8=%E2%9C%93&q=acceptance-tests&type=&language=)), I think it makes sense for us to start being explicit here as `cf-acceptance-tests` as we will definitely have other acceptance-tests added. Could you also use kebab-case instead of snake-case?

f0rmiga: An example of accessing the Helm value as kebab-case: https://github.com/SUSE/kubecf/blob/04bca19d055fdbc522564a6bcfd27a794c0527c4/deploy/helm/kubecf/assets/operations/instance_groups/tcp-router.yaml#L1

aduffeck: @f0rmiga Yes, I'm aware of the kebab-case issues and how to work around them. I wonder if it's worth it, though.

In the end I opted for snake-case because it makes things easier and the current `values.yaml` is inconsistent anyway. For example, `system_domain` and `kube.storage_class` and others are snake_case and it's even used for instance groups already, e.g. in `sizing.tcp_router`.

I think it would be good to streamline it one way or the other, ideally before the release. wdyt?

f0rmiga: The project uses snake-case for pretty much everything but respects names. I'm aware that we merged code that skipped that informal rule even after I requested changes related to that in sizing. I'd like to have that fixed before release - to any direction we decide to follow.

I can see that from a developer perspective making everything snake-case is easier - we just use dots everywhere.

An open question: for the final user - the cluster operator, is it easier to use the real name of the release, instance-group, job, etc, or just make anything snake-case everywhere?

f0rmiga: Note that from a developer perspective, we can make the code way simpler in many cases by using the real names, being that kebab or snake-case. It can be achieved by iterating over the keys of a map to generate template snippets.

aduffeck: My personal opinion is that using snake-case everywhere is the best compromise because it's consistent on the user-facing side **and** simplifies the code consuming the values. But then again that's just me, it should probably be decided in a larger group.

viovanov: Let's go with snake case - seems to be easier to deal with in code, and it's consistent with other BOSH things like properties.

aduffeck: I'll leave it in snake case for now and created https://github.com/SUSE/kubecf/issues/258 as a follow-up issue.

PR #261, opened 2019-12-12
Refactored patches to use sentinel files

f0rmiga:
## Description

A refactor to use a more reliable approach for checking if a file was already patched - sentinel files.

## Motivation and Context

During the work on https://github.com/SUSE/kubecf/pull/256, I noticed that the current approach doesn't play well with diffs with multiple hunks. This PR is a follow-up from that.

## How Has This Been Tested?

Deployed locally.

## Types of changes

## Checklist:


deploy/helm/kubecf/assets/operations/instance_groups/app-autoscaler.yaml
```diff
@@ -995,7 +995,20 @@
...
+        target="/var/vcap/all-releases/jobs-src/postgres/postgres/templates/pre-start.sh.erb"
+        sentinel="${target}.patch_sentinel"
+        if [[ -f "${sentinel}" ]]; then
+          echo "Patch already applied. Skipping"
+          exit 0
+        fi
+
+        sed -i "s/sysctl/#sysctl/g" "${target}"
+
+        touch "${sentinel}"
```
mook-as: If the script needs to be this complicated, it should live in a separate file.

(Or we could just do `sed -i 's/^sysctl/#sysctl/g'` because that has no issues with re-patching)

f0rmiga: Done in https://github.com/SUSE/kubecf/pull/261/commits/059445e69f2611d0af7e51a29e5aa38ae9a45874.

PR #272, opened 2019-12-17
Move the auctioneer to its own instance group

f0rmiga:
## Description

Following up on #271, we are moving the auctioneer to its own instance group in order to have it in active-passive HA mode and the scheduler instance group in active-active HA mode.

## Motivation and Context

The auctioneer doesn't support active-active HA mode while the rest of the jobs under the scheduler instance group do.

## How Has This Been Tested?

Deployed kubecf with 2 auctioneer instances and 3 scheduler instances and ran cf-acceptance-tests.

```yaml
sizing:
  auctioneer:
    instances: 2
  scheduler:
    instances: 3
```

## Types of changes

## Checklist:


f0rmiga:
@mook-as It does `15:39:32 main [ERRO] asked to process document index 0 but there are only 0 document(s)`. Not the clearest message coming from `jq` though...

PR #273, opened 2019-12-18
Fix the bbs readiness probe

f0rmiga:
## Description

The readiness probe for `bbs` was broken given that it always responded as healthy while not being.

## Motivation and Context

We want the `bbs` readiness probe to respond correctly so that the service endpoints can route to the instance that acquired the lock.

## How Has This Been Tested?

Deployed kubecf and ran CATS.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
@@ -70,7 +70,7 @@
             # The /ping endpoint on the healthcheck endpoint is unusable for this readiness probe as
             # it always returns success if the server it up and running. The port 8889, on the other
             # hand, is useful because bbs will start listening on it only if it acquired the lock
-            # on the locket.
+            # via locket.
             # https://github.com/cloudfoundry/bbs/blob/513a9420c3244966c4391d3f3a830f1732f1398a/handlers/ping_handler.go#L17-L21
             tcpSocket:
               port: 8889
--- a/deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
@@ -67,8 +67,10 @@
       healthcheck:
         bbs:
           readiness:
-            httpGet:
-              port: 8890
+            # The /ping endpoint on the healthcheck endpoint is unusable for this readiness probe.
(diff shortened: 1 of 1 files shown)


f0rmiga:
@mook-as The commit message indeed doesn't state the reason, but anyone looking at the small code that it contains will see that the comment does state the reason. The ping always returns success HTTP code despite the bbs instance acquiring the lock or not. On the other hand, bbs only listens on port 8889 if it had acquired the lock, therefore the tcpSocket probe works as expected.

mook-as:
@f0rmiga In that case, please add that second part (about acquiring the lock) to the comment, since that's what we _actually_ care about?

f0rmiga:
@mook-as Done.

deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
```diff
@@ -67,8 +67,13 @@
       healthcheck:
         bbs:
           readiness:
-            httpGet:
-              port: 8890
+            # The /ping endpoint on the healthcheck endpoint is unusable for this readiness probe as
+            # it always returns success if the server it up and running. The port 889, on the other
```
mook-as: Nit: typo:
```suggestion
            # it always returns success if the server it up and running. The port 8889, on the other
```

deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
```diff
@@ -67,8 +67,13 @@
...
+            # hand, is useful because bbs will start litening on it only if it acquired the lock
```
mook-as:
```suggestion
            # hand, is useful because bbs will start listening on it only if it acquired the lock
```

deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
```diff
@@ -67,8 +67,13 @@
...
+            # on the locket.
```
mook-as:
```suggestion
            # via locket.
```

deploy/helm/kubecf/assets/operations/instance_groups/diego-api.yaml
```diff
@@ -67,8 +67,13 @@
...
+            # https://github.com/cloudfoundry/bbs/blob/513a9420c3244966c4391d3f3a830f1732f1398a/handlers/ping_handler.go#L17-L21
```
mook-as: This link doesn't _actually_ point at the bit where it acquires locket…
```suggestion
```

PR #274, opened 2019-12-18
Replaced yaml2json with yq

f0rmiga:
## Description

Since `yq` has more options than `yaml2json`, and we introduced it in the project already, this PR removes `yaml2json` completely from the project.

## Motivation and Context

`yaml2json` was introduced in the past to simply convert YAML to JSON, but it's not capable of doing the inverse conversion - JSON to YAML. We then introduced `yq`, but left some usages of `yaml2json` behind.

## How Has This Been Tested?

The scripts using `yaml2json` and `yq` are part of the CI.

## Types of changes

## Checklist:

### DIFF
--- a/.drone/pipelines/default/steps/deploy/wait_kubecf.sh
+++ b/.drone/pipelines/default/steps/deploy/wait_kubecf.sh
@@ -27,7 +27,7 @@ while IFS='' read -r line; do instance_groups+=("${line}"); done < <(
     --namespace "${KUBECF_NAMESPACE}" \
     --output jsonpath='{ .data.manifest\.yaml }' \
     | base64 --decode \
-    | yq \
+    | yq read --tojson - \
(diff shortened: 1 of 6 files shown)


.drone/pipelines/default/steps/deploy/wait_kubecf.sh
```diff
@@ -27,7 +27,7 @@ while IFS='' read -r line; do instance_groups+=("${line}"); done < <(
     --namespace "${KUBECF_NAMESPACE}" \
     --output jsonpath='{ .data.manifest\.yaml }' \
     | base64 --decode \
-    | yaml2json \
+    | yq \
```
mook-as: Doesn't this need to be
```suggestion
    | yq read --tojson - \
```

f0rmiga: It does. Thanks for catching this.

PR #301, opened 2020-01-09
Exposed the healthcheck port for tcp-router

f0rmiga:
## Description

Since `tcp-router` will only listen on the TCP port range when it's required, it provides healthcheck on port 80 (default) for upstream load balancers to probe.

## Motivation and Context

Upstream load balancers like the classic AWS ELB require the healthcheck to pass in order to mark the instances as `In Service`.

## How Has This Been Tested?

Deployed kubecf on EKS and verified that the classic ELB marks the instances as `In Service`.

## Types of changes

## Checklist:


f0rmiga:
It's actually haproxy instead of nginx.

Issue #324, opened 2020-01-15
Test containers restart on failure


andreas-kupries:
**Describe the bug**

When tests like smoke, brain, or cf-acceptance fail their container is restarted.
This is not desirable.

This seems to be a policy inherited indirectly from the quarks-job which creates the job, which creates the pod, the container belongs to.

**To Reproduce**
Run one of the test suites in a way that has them fail.
(For brains and cf-acceptance it seems that just running them at all is enough at the moment).

**Expected behavior**
The failed container should not restart.
I suspect that it is desirable to have the controlling pod to be terminated and deleted.
I suspect that it is desirable to have the controlling job to be terminated and deleted.
The controlling quarks-job however should not be touched.

**Environment**
 - cf-operator-v1.0.3-0.g5dd886a5
 - kubecf 46b5561fff753772645fedcc954134fe7d04e341 (Implies cf-operator above)

**Additional context**

f0rmiga:
@fargozhu Could we schedule this to be fixed in the next sprint, please?

viovanov:
Might need help in the operator.

Issue #144, opened 2019-11-13
CATS: Failure in internetless suite


fargozhu:
**Describe the bug**

Seeing this when running CATS, with only the internetless test group enabled.

```
Summarizing 1 Failure:

[Fail] [internetless] App external traffic denied [It] denies external traffic from app containers 
/var/vcap/packages/acceptance-tests/src/github.com/cloudfoundry/cf-acceptance-tests/internetless/app_external_traffic_denied.go:66
```

```
denies external traffic from app containers

[2019-11-06 07:40:25.37 (UTC)]> curl -k -H Expect: -s https://CATS-1-APP-9a5e25d61cfd5d90.10.104.255.255.nip.io/curl/www.google.com/80 
{"stdout":"HTTP/1.1 200 OK\r\nDate: Wed, 06 Nov 2019 07:40:26 GMT\r...
```

**To Reproduce**
N/A

**Expected behavior**
To pass

**Environment**
N/A

**Additional context**
N/A

bikramnehra:
Full running logs on latest `master`: [CATs_internetless_logs.txt](https://github.com/SUSE/kubecf/files/3881738/CATs_internetless_logs.txt)

bikramnehra:
By looking at the [docs](https://github.com/cloudfoundry/cf-acceptance-tests#explanation-of-test-groups) and [PR](https://github.com/cloudfoundry/cf-acceptance-tests/pull/346) for `internetless` CATs test suite, it seems like its expected to have a network/environment with no access to external connectivity for the application containers.

The [internetless test](https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/internetless/app_external_traffic_denied.go) seems to be passing by simulating such conditions via deleting the [Default Public ASG](https://docs.cloudfoundry.org/concepts/asg.html#default-asg) which provides access to outside network:

`cf delete-security-group public_networks`

Here are the attached logs from the CATs; 
[CATs_internetless_logs_passing.txt](https://github.com/SUSE/kubecf/files/3889122/CATs_internetless_logs_passing.txt)

I have posted a [question](https://cloudfoundry.slack.com/archives/CAK2KNWC9/p1574709479004800) on CF slack#cats-users channel to confirm this. If the hypothesis is correct then either we need to drop the `internetless` test suite or have a mechanism in place to simulate such environment before running it.

bikramnehra:
@ericpromislow is helping us following up with upstream cloudfoundry to check on this issue.

[...]

PR #346, opened 2020-01-15
Fix CAT problems with internetless suite. Runs separate from main CATs

andreas-kupries:
- Factored the helpers into separate file.
- Added log colorization
- Hacking around #324 by explicitly looking up and killing the job when the container is done, failure or not.
- Added code to re-configure the cluster (and wait to settle).
- Apply/Remove network policy creating the isolated environment for the suite.

## Motivation and Context

See #144

## How Has This Been Tested?

Tested with the Drone CI.

## Types of changes

## Checklist:


.drone/pipelines/default/runtime/cats_common.sh
```diff
@@ -0,0 +1,191 @@
...
+	if [[ -n "${exit_code}" ]]; then
+	    break
+	fi
+	sleep 1
+    done
+    echo "${exit_code}"
+}
+
+# Core of CAT execution
+cf_acceptance_tests_go() {
```
f0rmiga: Maybe call this `run_cf_acceptance_tests`?

.drone/pipelines/default/steps/test/cats_internetless.sh
```diff
@@ -0,0 +1,148 @@
...
+    { ".drone/pipelines/default/runtime/cats_kubecf.sh" >& /dev/null; } || true
+    wait_for_ig_job || exit 1
+}
+
+deconfigure_cats_suites() {
+    # shellcheck disable=SC2005
+    echo "$(blue "Configure CATS: defaults")"
+
+    # Undo the limited configuration of the cats.
+    { ".drone/pipelines/default/runtime/decats_kubecf.sh" >& /dev/null; } || true
```
f0rmiga: I think `decats_kubecf` is a misnomer. I only understood it after I read the whole logic of this file. `configure_cats_suites` is also a misnomer. Even though it's "configuring" CATS suites, it's very specific to the `internetless` suite.

I don't think it's the responsibility of the internetless suite to "deconfigure" CATS. Just leave the way it exited. If another configuration is needed in the future, it should be the responsibility of the next CI job or pipeline.

dev/linters/shellcheck.sh
```diff
@@ -6,4 +6,5 @@ workspace=$(bazel info workspace)
 
 # shellcheck disable=SC2046
 # We want word splitting with find.
-bazel run @shellcheck//:binary -- $(find "${workspace}" -name '*.sh')
+bazel run @shellcheck//:binary -- \
+      $(find "${workspace}" -name '*.sh' | grep -v '/_work/')
```
f0rmiga: This command should ignore everything under `.gitignore` instead of hardcoding `_work` here.

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
```diff
@@ -6,41 +6,19 @@ set -o errexit -o nounset
 source ".drone/pipelines/default/runtime/binaries.sh"
 # shellcheck disable=SC1091
 source ".drone/pipelines/default/runtime/config.sh"
-
```
f0rmiga: I dislike the changes in this file. Ideally, we want to have this logic managed by Bazel under the `//testing/...`. This script and the smoke-tests one should be very similar and do only what they are currently doing. So the end result would be:

```
$ bazel run //testing/cf_acceptance_tests
... waiting for pod to exists
... a bunch of logs from CATS...
$ echo $?
0
```

The CI pipelines would then:

1. Deploy kubecf with CATS configured without the internetless suite.
2. `bazel run //testing/cf_acceptance_tests`
3. Delete the hanging k8s job that ran CATS.
4. Change the CATS properties to only run the internetless suite.
5. `bazel run //testing/cf_acceptance_tests`

f0rmiga: Note that I'm not asking to move this script and the smoke-tests to be managed by Bazel. Just try not to change it.

.drone/pipelines/default/runtime/cats_common.sh
andreas-kupries: Accepted.

.drone/pipelines/default/steps/test/cats_internetless.sh
andreas-kupries: So the actions you are asking for are:
- Rename `configure_cats_suites` to `configure_cats_for_internetless`
- Drop the entire `decats_...` function and script.
right ?
If yes, accepted.

dev/linters/shellcheck.sh
andreas-kupries: Now idea how to do that. Essentially a tool reading .gitignore, and a list of paths, and returning the filtered set of paths. ... Has git a subcommand for that ?

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
andreas-kupries: So you want to inline `cats-common.sh` again ?

.drone/pipelines/default/steps/test/cats_internetless.sh
f0rmiga: Correct

dev/linters/shellcheck.sh
andreas-kupries: Hm. https://git-scm.com/docs/gitignore mentions `git ls-files` and `git read-tree` ... Might be possible to use that, merging find+grep into one command.

andreas-kupries: Generally accepted, will have to trial locally a bit.

f0rmiga: How about inverting the rationale here?
```
bazel run @shellcheck//:binary -- $(git ls-files | awk '/.sh$/ { print $0 }')
```

f0rmiga: Oh, when I commented, your comments were not visible yet. I'm glad that we though in the same way.

f0rmiga: Actually, you don't even need the `awk`:
```
bazel run @shellcheck//:binary -- $(git ls-files '*.sh')
```

f0rmiga: Shellcheck doesn't like relative paths. You will need:
```
bazel run @shellcheck//:binary -- \
  $(git ls-files --full-name '*.sh' | sed "s|^|${workspace}/|")
```

.drone/pipelines/default/runtime/cats_common.sh
andreas-kupries: Resolved with commit 97720239811d858b79602230c00ca923b197896d

.drone/pipelines/default/steps/test/cats_internetless.sh
andreas-kupries: Hopefully resolved with commit af340e6f707d501c3928efc88ed42389180f7845

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
andreas-kupries: So, points 1. and 2. are currently distinct steps in the Drone CI. 
My understanding right now is that you wish to have 3. and 4. also as distinct steps.
Ditto for 5., calling the same script as in step 2.
Note that your steps miss the `isolate_network` operation (and the complementary removal of same).

Personally I would move step 3 into the script for 2, and then simply remove it later, when #324 is solved upstream and in the deployment in general.

So, the full set of Drone steps you are looking for is

  1. Deploy kubecf, standard
  2. Run CATS
       - Drop hanging job
  3. Redeploy kubecf for internetless
  4. Isolate network
  5. Run CATS
  6. Remove network isolation
  7. (Redeploy kubecf for standard)

Note that in this setup failure of the 2nd CATS prevents undoing the network isolation.
The current scripts take care to undo even in presence of failure.

dev/linters/shellcheck.sh
andreas-kupries: Resolved with commit b087191de9161718ed1878a2aadc97d15f28d4e1

andreas-kupries:
The referenced commits are not pushed yet, wish to resolve the last thing first, and then trial the changes locally.

.drone/pipelines/default/steps/test/cf_acceptance_tests.sh
andreas-kupries: Resolved with commit 3e7c18ef8275f6170fa0508edeb62fcf52201e05.
Step 4 above moved into step 3.
Left out steps 6 and 7.

.drone/config.star
```diff
@@ -67,6 +67,8 @@ def main(ctx):
 
     steps.append(test_step("smoke_tests"))
     steps.append(test_step("cf_acceptance_tests"))
+    steps.append(test_step("kubecf_redeploy_cats_internetless"))
+    steps.append(test_step("cf_acceptance_tests"))
```
f0rmiga: Got an error for this line: https://kubecf-drone-ci-server.herokuapp.com/SUSE/kubecf/62.
`default: Linter: duplicate step name`.
The `test_step` does some magic on the name and path. You can try something like:
```python
def test_step(name, subset_name = ""):
    return step(
        name = "test:{name}{subset_name}".format(
            name = name,
            subset_name = ":{}".format(subset_name) if subset_name != "" else ""
        ),
        commands = [
            "{base_path}/steps/test/{name}.sh".format(
                base_path = DRONE_BASE_PATH,
                name = name,
            ),
        ],
        extra_volumes = [
            kube_volume(),
        ],
        network_mode = "host",
    )
```
Then use the `subset_name` arg with the `internetless` value.

andreas-kupries: Resolved with commit 08d087cd6c60e6a46c1475aad1731ea05e188c61

mook-as:
This PR is now obsolete: `.drone/` no longer exists.

PR #347, opened 2020-01-15
Add PSP and default bindings

f0rmiga:
## Description

Fixes https://github.com/SUSE/kubecf/issues/306.

By default, KubeCF will create a PSP for its enjoyment, while allowing an existing PSP to be passed via the `kube.psp.default` Helm value. The PSP is then bound to the default service account in the namespace kubecf is installed.

## Motivation and Context

Please, visit https://github.com/SUSE/kubecf/issues/306 for more context.

## How Has This Been Tested?

Deployed on CaaSP 4, which is very restrictive on PSPs.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/templates/rbac.yaml
+++ b/deploy/helm/kubecf/templates/rbac.yaml
@@ -49,7 +49,7 @@ spec:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: {{ .Release.Name }}-default
+  name: {{ .Release.Name }}-default-psp
   namespace: {{ .Release.Namespace | quote }}
 rules:
 - apiGroups: [policy]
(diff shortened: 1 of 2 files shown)


deploy/helm/kubecf/templates/rbac.yaml
```diff
@@ -0,0 +1,78 @@
...
+  - awsElasticBlockStore
+  - azureDisk
+  - azureFile
+  - vsphereVolume
+{{- end }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Release.Name }}-default
```
jandubois: I think the Role name (and RoleBinding name below) should be `{{ .Release.Name }}-default-psp` to show that they are about default PSPs. Calling a role just `default` is kind of meaningless.

f0rmiga: Indeed.

f0rmiga: Done in https://github.com/SUSE/kubecf/pull/347/commits/7de118a7c5929a7ed23f0e24f79a4b55059a850b.

PR #356, opened 2020-01-23
Automatic Helm chart versioning

### LINKED ISSUE #350: Use an obvious non-release chart version on master

Right now, we keep the latest released version on master. We need to only set the version on a release branch, so users don't get confused with the charts in the S3 bucket.

Set the version of the helm chart to a very obvious non-release version on master. Then set the proper version whenever the release branch is created.

Start using the release branches.

**Additional context**

The S3 bucket is a dump from CI.

f0rmiga:
## Description

The Helm chart version will be set using the git branch name if it complies with the SemVer, otherwise, it will be set to v0.0.0-(short_commit_hash).

## Motivation and Context

Instead of having to set the chart version on every release, the chart builder will use the branch name. On master, it will be tagged in a very non-obvious release version.

Resolves #350.

## How Has This Been Tested?

Built the chart and inspected it:

```
helm inspect chart bazel-bin/deploy/helm/kubecf/kubecf.tgz | less
```

## Types of changes

## Checklist:


fargozhu:
@f0rmiga I think I am missing something here but...

**Steps:**

1. git branch -m 0.2.0
2. helm package 
3. helm inspect

and the output is:
```
apiVersion: v1
appVersion: v0.0.0
description: A Helm chart for KubeCF
name: kubecf
version: v0.0.0
```

rules/helm/package.tmpl.sh
```diff
@@ -0,0 +1,33 @@
...
+for g in ${GENERATED}; do
+  cp "${g}" "${build_dir}"/ > /dev/null
+done
+
+"${HELM}" init --client-only > /dev/null
+
+version='v0.0.0-{{ (ds "workspace_status").STABLE_GIT_COMMIT_SHORT }}'
+# If the branch name is a valid SemVer, use it as the chart version.
+if [[ -n "$(echo '{{ (ds "workspace_status").STABLE_GIT_BRANCH }}' | awk '/(?in)^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-([a-z-][\da-z-]+|[\da-z-]+[a-z-][\da-z-]*|0|[1-9]\d*)(\.([a-z-][\da-z-]+|[\da-z-]+[a-z-][\da-z-]*|0|[1-9]\d*))*)?(\+[\da-z-]+(\.[\da-z-]+)*)?$/ { print $0 }')" ]]; then
+  version='{{ (ds "workspace_status").STABLE_GIT_BRANCH }}'
```
mook-as: That looks… horrible? Scary? Nightmare-inducing? 😛 

Since the workspace status script is Ruby, can we just make _that_ give us a semver string instead, falling back to `v0.0.0-{{…}}`? Call it `STABLE_RELEASE_VERSION` or something?

rules/workspace_status/status_merger.tmpl.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# This script merges the two workspace status files into one.
+
+set -o errexit
+
+echo "$(cat "{info_file}") $(cat "{version_file}")" | "{jq}" --slurp add > "{workspace_status}"
```
mook-as: FWIW:

- `cat` takes multiple file names (it stands for con*cat*enate after all)
- `jq` takes files, so this is just
```suggestion
"{jq}" --slurp add "{info_file}" "{version_file}" > "{workspace_status}"
```

rules/helm/package.tmpl.sh
f0rmiga: The script was rewritten using Ruby.

f0rmiga:
@fargozhu You need to build using Bazel since it's the one managing everything, like generated files, etc.

1. bazel build //deploy/helm/kubecf
2. helm inspect chart bazel-bin/deploy/helm/kubecf/kubecf.tgz

.gitlab/pipelines/stages/publish.sh
```diff
@@ -2,10 +2,19 @@
...
+if [ -z "${version}" ]; then
+  >&2 echo "Version not set!"
+  exit 1
+fi
+
 mc config host add "s3" "https://s3.amazonaws.com" "${AWS_ACCESS_KEY}" "${AWS_SECRET_KEY}"
 
-mc cp output/kubecf-*.tgz "s3/scf-v3"
-echo "URL: https://scf-v3.s3.amazonaws.com/$(basename output/kubecf-*.tgz)"
+mc cp "output/kubecf.tgz" "s3/scf-v3/kubecf-${version}.tgz"
```
mook-as: We should factor out `scf-v3` and put it somewhere easier to change (since I expect we _will_ want to change it).

But that's totally optional for _this_ PR.

.gitlab/pipelines/stages/publish.sh
```diff
@@ -2,10 +2,19 @@
...
+echo "URL: https://scf-v3.s3.amazonaws.com/kubecf-${version}.tgz"
 
-mc cp output/bundle-*.tgz "s3/kubecf"
-echo "URL: https://kubecf.s3.amazonaws.com/$(basename output/bundle-*.tgz)"
+mc cp "output/kubecf.tgz" "s3/kubecf/kubecf-${version}.tgz"
+echo "URL: https://kubecf.s3.amazonaws.com/kubecf-${version}.tgz"
```
mook-as: Huh, we're copying it to `scf-v3` _and_ `kubecf`? Never mind that previous comment then, I'm just plain confused now :D

rules/helm/def.bzl
```diff
@@ -131,3 +114,44 @@ template = rule(
...
     },
 )
+
+def _version_impl(ctx):
+    script = ctx.actions.declare_file("{}.rb".format(ctx.attr.name))
+    output = ctx.actions.declare_file("version.txt")
+    outputs = [output]
+    contents = """
+        open('{output}', 'w') do |f|
+          f << `{helm} inspect chart {chart}`[/version: (.*)/, 1]
```
mook-as: Would this bubble up unexpected errors correctly?

f0rmiga: Bazel itself should fail.

.gitlab/pipelines/stages/publish.sh
f0rmiga: The idea is that we deprecate scf-v3, and start telling people to use kubecf.

.gitlab/pipelines/stages/publish.sh
f0rmiga: I'll do that once we move to Concourse.

rules/helm/def.bzl
f0rmiga: Actually, we need to add `raise if not $?.success?` to actually handle the exit code.

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
+require "fileutils"
+require "open3"
+
+# Variables expanded by Bazel.
+package_dir = "[[package_dir]]"
+multipath_sep = "[[multipath_sep]]"
+tarsStr = "[[tars]]"
+generatedStr = "[[generated]]"
+helm = "[[helm]]"
+output_tgz = "[[output_tgz]]"
```
mook-as: Don't _really_ like the substitution (I'd prefer passing them in as environment variables to be evaluated by Ruby), but :man_shrugging:

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
...
+FileUtils.mkdir_p(build_dir)
+
+# Copy the source static files to the temporary build directory.
+Dir.glob(File.join(package_dir, "**", "*")) do |file|
+  dest = File.join(tmp_build_dir, file)
+  ftype = File.ftype(file)
+  if ftype == "directory"
+    FileUtils.mkdir_p(dest)
+  else
+    file = File.readlink(file) if ftype == "link"
```
mook-as:
```suggestion
    file = File.readlink(file) while File.ftype(file) == "link"
```

Actually shouldn't we completely resolve the link before checking if it's a directory?

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
...
+    FileUtils.cp(file, dest)
+  end
+end
+
+# Extract the tar files into the temporary build directory.
+tars = tarsStr.split(multipath_sep)
+tars.each do |file|
+  file = File.readlink(file) if File.ftype(file) == "link"
```
mook-as:
```suggestion
  file = File.readlink(file) while File.ftype(file) == "link"
```

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
...
+  `tar xf "#{file}" -C "#{build_dir}"`
```
mook-as: Check exit status? Don't eat `STDOUT`, but show it to the user instead? (I prefer `Process.wait Process.spawn(…)` and never capture `STDOUT` to begin with.)

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
...
+end
+
+# Copy the generated files into the temporary build directory.
+generated = generatedStr.split(multipath_sep)
+generated.each do |file|
+  file = File.readlink(file) if File.ftype(file) == "link"
+  FileUtils.cp(file, build_dir)
+end
+
+`"#{helm}" init --client-only`
```
mook-as: Same as above (and any other unnecessary uses of backticks).

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
...
+
+# Handle chart versioning based on git state.
+version = "v0.0.0-#{git_commit_short}"
+
+# A semver that matches what Helm uses.
+# https://github.com/Masterminds/semver/blob/910aa146bd66780c2815d652b92a7fc5331e533c/version.go#L41-L43
+semver_regex = /^v?([0-9]+)(\.[0-9]+)?(\.[0-9]+)?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?$/
```
mook-as: That's a lot of unnecessary captures…
Also, please use `/x` and multiple lines to make this more readable.

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
...
+if git_branch.match(semver_regex)
+  version = git_branch
+end
+
+# Package and return the output path.
+package_cmd = "'#{helm}' package '#{build_dir}' --version='#{version}' --app-version='#{version}'"
+package_output = Open3.popen3(package_cmd) do |stdin, stdout, stderr, wait_thread|
+  stderr.each {|l| STDERR.puts l }
+  status = wait_thread.value
+  raise if not status.success?
```
mook-as:
```suggestion
  raise unless status.success?
```
What, exactly, are we `raise`ing?

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
...
+# A semver that matches what Helm uses.
+# https://github.com/Masterminds/semver/blob/910aa146bd66780c2815d652b92a7fc5331e533c/version.go#L41-L43
+semver_regex = /^v?([0-9]+)(\.[0-9]+)?(\.[0-9]+)?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?$/
+if git_branch.match(semver_regex)
+  version = git_branch
+end
+
+# Package and return the output path.
+package_cmd = "'#{helm}' package '#{build_dir}' --version='#{version}' --app-version='#{version}'"
+package_output = Open3.popen3(package_cmd) do |stdin, stdout, stderr, wait_thread|
```
mook-as: It looks like this block can be `Open3.capture2` instead (so you don't need to worry about dumping `STDERR` or handling the thread, just check the exit status after the fact).

rules/helm/package.tmpl.rb
f0rmiga: I prefer not changing this at all in order to match what the reference link points to, which is what Helm uses for the check.

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,70 @@
+require "fileutils"
+require "open3"
+
+# Variables expanded by Bazel.
```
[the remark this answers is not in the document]
f0rmiga: ¯\_(ツ)_/¯

rules/helm/package.tmpl.rb
f0rmiga: And I love the fact that I can avoid environment variables. :sweat_smile:

rules/helm/package.tmpl.rb
```diff
@@ -0,0 +1,78 @@
...
+    file = File.readlink(file) while File.ftype(file) == "link"
+    FileUtils.cp(file, build_dir)
+  end
+
+  pid = Process.spawn("'#{helm}' init --client-only")
+  _, status = Process.wait2 pid
+  exit 1 unless status.success?
+
+  # Handle chart versioning based on git state.
+  version = "v0.0.0-#{git_commit_short}"
```
mook-as: Thinking about it some more, fallback should probably be closer to `git describe` output (so that the _current_ builds would look something like 0.2.0-derived).

But I'm totally happy to punt that to a future PR.

PR #366, opened 2020-01-29
Fixing out of memory issue when scaling doppler instances

bikramnehra:
## Description
fixes: https://github.com/SUSE/kubecf/issues/241

## Motivation and Context

`log-cache` by default is configured to 50% of total memory available to the instance:

https://github.com/cloudfoundry/log-cache-release/blob/v2.6.4/jobs/log-cache/spec#L35-L37

which exceeds the amount of memory actually available to be used:

```
/:/var/vcap/jobs/log-cache# cat /proc/meminfo   
MemTotal:       15652300 kB
MemFree:          569484 kB
MemAvailable:    9008472 kB
```

Therefore, this fix will limit the memory to 25% which falls under the actual available memory.
 
Discussion on CF Slack: https://cloudfoundry.slack.com/archives/CBFB7NP9B/p1580161281014400

## How Has This Been Tested?

Locally on minikube by scaling the doppler instances to two and running smoke-tests.
```
❯ k logs kubecf-smoke-tests-ca77bb96d4d42c8c-znzvg -n kubecf -c smoke-tests-smoke-tests -f
kubectl logs kubecf-smoke-tests-ca77bb96d4d42c8c-znzvg -n kubecf -c smoke-tests-smoke-tests -f
Running smoke tests...
Running binaries smoke/isolation_segments/isolation_segments.test
smoke/logging/logging.test
smoke/runtime/runtime.test
[1580245198] CF-Isolation-Segment-Smoke-Tests - 4 specs - 4 nodes SSSS SUCCESS! 13.545755112s 
[1580245198] CF-Logging-Smoke-Tests - 2 specs - 4 nodes S• SUCCESS! 1m3.746576247s 
[1580245198] CF-Runtime-Smoke-Tests - 2 specs - 4 nodes S• SUCCESS! 1m4.298290082s 

Ginkgo ran 3 suites in 2m21.645606928s
Test Suite Passed

```

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/doppler.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/doppler.yaml
@@ -3,6 +3,10 @@
   path: /variables/name=log_cache_ca/options/common_name
   value: log-cache-ca
 
+- type: replace
+  path: /instance_groups/name=doppler/jobs/name=log-cache/properties/memory_limit_percent?
+  value: 25
+
 # Add quarks properties for doppler.
 - type: replace
   path: /instance_groups/name=doppler/jobs/name=doppler/properties/quarks?
--- a/deploy/helm/kubecf/templates/sizing.yaml
+++ b/deploy/helm/kubecf/templates/sizing.yaml
@@ -151,11 +151,15 @@ data:
 {{- end }}
 {{- end }}
 
-    # Temporarily fixes https://github.com/SUSE/kubecf/issues/241.
-    # TODO: properly investigate and fix log-cache.
+{{- if .Values.sizing.doppler.instances }}
+    - type: replace
+      path: /instance_groups/name=doppler/instances
+      value: {{.Values.sizing.doppler.instances}}
+{{- else if not .Values.high_availability }}
     - type: replace
       path: /instance_groups/name=doppler/instances
       value: 1
+{{- end }}
 
 {{- if .Values.sizing.log_api.instances }}
     - type: replace
--- a/deploy/helm/kubecf/values.yaml
+++ b/deploy/helm/kubecf/values.yaml
@@ -191,6 +191,8 @@ sizing:
     instances: ~
   diego_cell:
     instances: ~
+  doppler:
+    instances: ~
   eirini:
     instances: ~
   log_api:


mook-as:
So chasing down all the layers of stuff, the core problem is that when we get the memory usage we [parse /proc/meminfo](https://github.com/cloudfoundry/gosigar/blob/a3af48c7e16a418be8366ffed86df345c5c4703d/sigar_linux.go#L65) (via [here](https://github.com/cloudfoundry/log-cache-release/blob/ef3677bab7a9044b5cc39e95980c8d80ba321146/src/internal/cache/memory_analyzer.go#L53-L54)) which doesn't reflect the container limits?

Just bumping it down doesn't _really_ help, since now you can't scale past 4 instances (on the same node).  The necessary fix is really to change how we get the available memory limits.

loewenstein:
@mook-as I think we had it crash also with rather low requests==limits set.

f0rmiga:
> @mook-as I think we had it crash also with rather low requests==limits set.

The pod is terminated if the memory limit is exceeded. `/proc/meminfo` always reports the total memory in the system, not what the container is allowed to use. Being only able to set a percentage is not really helpful.

bikramnehra:
This mechanism was designed while having VMs in mind, ideal solution would be to fix it on upstream by relying on a different mechanism, which the `log-cache` team tried but haven't been successful yet: https://cloudfoundry.slack.com/archives/CBFB7NP9B/p1580254462021300?thread_ts=1580161281.014400&cid=CBFB7NP9B

Also, correct me if I am wrong - `log-cache` is just an in-memory caching layer for logs therefore if there isn't enough memory available for caching things must still work though slightly slower. How would this impact scaling?

f0rmiga:
> How would this impact scaling?

As reported, this bug only shows up when using more than one instance of log-cache.

mook-as:
Taking some notes for how the memory limit is used:

- The percentage (50 by default) is eventually passed to [`store.NewPruneConsultant()`](https://github.com/cloudfoundry/log-cache-release/blob/v2.6.12/src/internal/cache/store/prune_consultant.go#L22).
- Which is used in [`store.PruneConsultant.GetQuantityToPrune()`](https://github.com/cloudfoundry/log-cache-release/blob/v2.6.12/src/internal/cache/store/prune_consultant.go#L44).
- That depends on values from [`cache.MemoryAnalyzer.Memory()`](https://github.com/cloudfoundry/log-cache-release/blob/v2.6.12/src/internal/cache/memory_analyzer.go#L49).
- The heap in use part comes from [`runtime.ReadMemStats()`](https://godoc.org/runtime#ReadMemStats)
- The total part comes from [`/proc/meminfo`](https://github.com/cloudfoundry/gosigar/blob/v1.1.0/sigar_linux.go#L69).
- While [`cache.MemoryAnalyzer.Memory()`](https://github.com/cloudfoundry/log-cache-release/blob/v2.6.12/src/internal/cache/memory_analyzer.go#L49) returns the `MemAvailable` field of `/proc/meminfo`, it appears to be unused.

So, I _think_ we might be able to get away with mounting over `/proc/meminfo` for the short term; we'll most likely need to experiment.

f0rmiga:
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: proc-meminfo
data:
  meminfo: |-
    MemTotal:         256000 kB
    MemFree:          256000 kB
    MemAvailable:     256000 kB
    Buffers:               0
    Cached:                0
    SwapCached:            0
    Active:                0
    Inactive:              0
    Active(anon):          0
    Inactive(anon):        0
    Active(file):          0
    Inactive(file):        0
    Unevictable:           0
    Mlocked:               0
    SwapTotal:             0
    SwapFree:              0
    Dirty:                 0
    Writeback:             0
    AnonPages:             0
    Mapped:                0
    Shmem:                 0
    Slab:                  0
    SReclaimable:          0
    SUnreclaim:            0
    KernelStack:           0
    PageTables:            0
    NFS_Unstable:          0
    Bounce:                0
    WritebackTmp:          0
    CommitLimit:           0
    Committed_AS:          0
    VmallocTotal:          0
    VmallocUsed:           0
    VmallocChunk:          0
    HardwareCorrupted:     0
    AnonHugePages:         0
    ShmemHugePages:        0
    ShmemPmdMapped:        0
    HugePages_Total:       0
    HugePages_Free:        0
    HugePages_Rsvd:        0
    HugePages_Surp:        0
    Hugepagesize:          0
    DirectMap4k:           0
    DirectMap2M:           0
    DirectMap1G:           0
---
apiVersion: v1
kind: Pod
metadata:
  name: example
spec:
  containers:
    - name: opensuse
      image: opensuse/leap:15.1
      command: ["/bin/sh", "-c", "cat /proc/meminfo"]
      volumeMounts:
      - name: proc-meminfo
        mountPath: /proc/meminfo
        subPath: meminfo
  volumes:
  - name: proc-meminfo
    configMap:
      name: proc-meminfo
  restartPolicy: Never
```

f0rmiga:
I'm closing this PR since it doesn't introduce a fix. The bug is still tracked on https://github.com/SUSE/kubecf/issues/241.

PR #369, opened 2020-01-29
Replace internal IP range settings with automatic setup

### LINKED ISSUE #304: Remove security group that allows apps to communicate with internal network
**Describe the bug**

Replaces https://github.com/SUSE/kubecf/issues/297.

The pod and service CIDRs are currently required to created an app security group that allows connections to the internal k8s network. It allows connection to service brokers like `credhub`.

**To Reproduce**

Always.

**Expected behavior**

This approach is erroneous and we should allow only the `credhub` IP.
An auto-errand can take care of it by resolving `credhub.service.cf.internal` and creating a security group allowing apps to communicate with this IP. We also should add `features.credhub.enabled=true` as default, but remove `credhub` and the auto-errand when this setting is `false`, allowing the usage of an external `credhub` instance.

**Environment**

Kubecf 0.1, all platforms.

mook-as:
## Description
This replaces the `kube.service_cluster_ip_range` and `kube.pod_cluster_ip_range` helm values with automatic detection for setting the application security groups.

## Motivation and Context
We currently have some internal app security group stuff for CredHub (from #281); this requires the administrator to set `kube.service_cluster_ip_range` and `kube.pod_cluster_ip_range` in their helm values.  This is unwieldy and error-prone; we should instead auto-detect the appropriate values and set the application security groups from code.

This fixes #304.

## How Has This Been Tested?
Locally ran CATS with focus on `credhub` (on kind).  The expectation is that creating this PR will _also_ trigger CI.

## Additional information
This is the first draft of getting things to work; it needs to be split into a few sub-PRs, I think.  But I'd like to get eyes on it early so it doesn't surprise too many people.

## Types of changes

Note that I don't expect this to bump versions for us, because:
1. We are at < 1.0
2. Any stale helm configuration would still be accepted, it just wouldn't _do_ anything.

## Checklist:


deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	"net/url"
+	"path/filepath"
+	"time"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+func makeHTTPClientWithCA(serverName string, caCert []byte) (*http.Client, error) {
+	certPool, err := x509.SystemCertPool()
```
f0rmiga: SystemCertPool is a hidden dependency. Ideally, `makeHTTPClientWithCA` will be a public method of the `HttpClient` struct. You can have something like:

```go
type HttpClient struct {
  systemCertPool func() (*x509.CertPool, error)
}

func NewHttpClient(systemCertPool func() (*x509.CertPool, error)) *HttpClient {
  return &HttpClient{
    systemCertPool: systemCertPool,
  }
}

func (httpClient *HttpClient) MakeWithCACert(serverName string, caCert []byte) (*http.Client, error) {
  certPool, err := httpClient.systemCertPool()
  ...
}
```

When using it, just construct with:

```go
httpClient := uaa.NewHttpClient(x509.SystemCertPool)
```

And there is another option would be to have another constructor for `HttpClient`:

```go
func NewDefaultHttpClient() *HttpClient {
  return &HttpClient{
    systemCertPool: x509.SystemCertPool,
  }
}
```

This is very important in order to be able to unit test this and ensure long-term maintainability.

f0rmiga: Please, create a new issue to add unit tests to all this so we can cover 100%.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get system cert pool: %w", err)
```
f0rmiga: The error coming from `err` is expected to describe that the error happened on getting the system cert pool. The message propagating `err` here should be something like `failed to make HTTP client with CA cert: %v`.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	ok := certPool.AppendCertsFromPEM(caCert)
+	if !ok {
+		fmt.Printf("Warning: failed to add CA certs for %s\n", serverName)
```
f0rmiga: I would expect it to be an error. It's ok to ignore this particular error if the caller judges to be the correct behaviour. After all, this is the main behaviour of this function.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+				ServerName: serverName,
+			},
+		},
+		Timeout: 60 * time.Second,
+	}, nil
+}
+
+// authenticate with UAA, returning the access token and refresh token.
+func authenticate(ctx context.Context, clientID, clientSecret string) (*http.Client, error) {
+	link, err := getCCLinkData(ctx)
```
f0rmiga: `getCCLinkData` is a hidden dependency.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+				RootCAs:    certPool,
+				ServerName: serverName,
+			},
+		},
+		Timeout: 60 * time.Second,
+	}, nil
+}
+
+// authenticate with UAA, returning the access token and refresh token.
+func authenticate(ctx context.Context, clientID, clientSecret string) (*http.Client, error) {
```
f0rmiga: This should also be a method of a struct. See below.

deploy/containers/credhub_setup/auth.go
f0rmiga: I don't think the resolution of the link should happen magically inside of this function. The link information (as separate InternalServiceHostname and CACert properties) should be passed to either the constructor of the struct or via the parameter list of the method.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	link, err := getCCLinkData(ctx)
+	if err != nil {
+		return nil, err
+	}
+	ccClient, err := makeHTTPClientWithCA(link.CC.InternalServiceHostname, []byte(link.CC.PublicTLS.CACert))
```
f0rmiga: `makeHTTPClientWithCA` is a hidden dependency. An instance of `HttpClient` should be passed as a dependency to the constructor of the struct. What we accept should be an interface that can be mocked in order to unit test this behaviour.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not make CC HTTP client: %w", err)
```
f0rmiga: The errors being propagated here and below should also reflect the current method, not the function called that caused the error.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	tokenURL, err := getTokenURL(ctx, link, ccClient)
```
f0rmiga: `getTokenURL` is also a hidden dependency. Same idea as the others.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	credentialsConfig := clientcredentials.Config{
+		ClientID:     clientID,
+		ClientSecret: clientSecret,
+		TokenURL:     tokenURL.String(),
+		Scopes:       []string{"cloud_controller.admin"},
+	}
+
+	fmt.Printf("Got UAA token URL: %s\n", tokenURL.String())
```
f0rmiga: `fmt.Printf` is a hidden dependency here. We should be able to inject a no-op logger for testing, and more robust loggers for production environments.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	certPath := filepath.Join(getMountRootFromContext(ctx), "run/uaa-ca-cert/ca.crt")
```
f0rmiga:
```suggestion
	certPath := filepath.Join(getMountRootFromContext(ctx), "run", "uaa-ca-cert", "ca.crt")
```

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	uaaCABytes, err := ioutil.ReadFile(certPath)
```
f0rmiga: `ioutil.ReadFile` is a hidden dependency.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+type ccInfoData struct {
+	AuthorizationEndpoint string `json:"authorization_endpoint"`
+	TokenEndpoint         string `json:"token_endpoint"`
+}
+
+func getCCLinkData(ctx context.Context) (*ccEndpointLinkData, error) {
+	var link ccEndpointLinkData
+	err := resolveLink(ctx, "cloud_controller_https_endpoint", &link)
```
f0rmiga: `cloud_controller_https_endpoint` is a magic string here. I don't think this function is needed at all. `resolveLink` can be called with a constant.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	return &link, nil
+}
+
+// getTokenURL fetches the OAuth token URL from the cloud controller
+func getTokenURL(ctx context.Context, link *ccEndpointLinkData, ccClient *http.Client) (*url.URL, error) {
+	ccURL := url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("%s:%d", link.CC.InternalServiceHostname, link.CC.PublicTLS.Port),
+		Path:   "/v2/info",
```
f0rmiga: `/v2/info` is a magic string. Ideally, it should come from the parameter list, but could also be a constant.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	var link ccEndpointLinkData
+	err := resolveLink(ctx, "cloud_controller_https_endpoint", &link)
+	if err != nil {
+		return nil, fmt.Errorf("could not get link: %w", err)
+	}
+	return &link, nil
+}
+
+// getTokenURL fetches the OAuth token URL from the cloud controller
+func getTokenURL(ctx context.Context, link *ccEndpointLinkData, ccClient *http.Client) (*url.URL, error) {
```
f0rmiga: `link` should be unwrapped to respect the principle of least knowledge (Law of Demeter).

f0rmiga: `ccClient` should be an interface to allow mocking it and asserting the behaviour of `getTokenURL` when the methods of `ccClient` fail.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	ccURL := url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("%s:%d", link.CC.InternalServiceHostname, link.CC.PublicTLS.Port),
+		Path:   "/v2/info",
+	}
+	infoResp, err := ccClient.Get(ccURL.String())
+	if err != nil {
+		return nil, fmt.Errorf("could not get CC info: %w", err)
```
f0rmiga: Same thing here about error propagation.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	var ccInfo ccInfoData
+	err = json.NewDecoder(infoResp.Body).Decode(&ccInfo)
+	if err != nil {
+		return nil, fmt.Errorf("could not read CC info response: %w", err)
+	}
+
+	fmt.Printf("CC info response: %+v\n", ccInfo)
```
f0rmiga: Hidden dependency.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+		Host:   fmt.Sprintf("%s:%d", link.CC.InternalServiceHostname, link.CC.PublicTLS.Port),
+		Path:   "/v2/info",
+	}
+	infoResp, err := ccClient.Get(ccURL.String())
+	if err != nil {
+		return nil, fmt.Errorf("could not get CC info: %w", err)
+	}
+
+	var ccInfo ccInfoData
+	err = json.NewDecoder(infoResp.Body).Decode(&ccInfo)
```
f0rmiga: Since I'm catching all hidden dependencies, why not `json.NewDecoder`?

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not read CC info response: %w", err)
+	}
+
+	fmt.Printf("CC info response: %+v\n", ccInfo)
+	tokenURL, err := url.Parse(ccInfo.TokenEndpoint)
```
f0rmiga: `url.Parse` is a hidden dependency.

deploy/containers/credhub_setup/auth.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid token url: %w", err)
+	}
+	tokenURL.Path += "/oauth/token"
```
f0rmiga: `/oauth/token` is a magic string.

deploy/containers/credhub_setup/auth_test.go
```diff
@@ -0,0 +1,209 @@
+package main
```
f0rmiga: This is not a unit testing file, therefore, it should be on a separate package and labelled accordingly using Bazel, so we can ignore it while doing only unit testing.

deploy/containers/credhub_setup/credhub_info.go
```diff
@@ -0,0 +1,103 @@
+package main
+
+// credhub_info.go contains functions to determine the credhub information
```
f0rmiga: Same thing here about structure. If it's doing `credhub` stuff, let's put it on a `credhub` package!

.drone/pipelines/default/runtime/helpers.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
mook-as: How does `runtime/bazel.sh` sound?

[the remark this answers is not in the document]
f0rmiga: Sounds ok. I'd still stick a comment on the top to restrict the scope of the file.

.gitlab/pipelines/stages/deploy/wait_kubecf.sh
```diff
@@ -54,5 +54,12 @@ for instance_group in "${instance_groups[@]}"; do
     --namespace "${KUBECF_NAMESPACE}"
 done
 
+# Wait for non-quarks jobs
```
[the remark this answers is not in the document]
mook-as: Fixed in 278c50010c0fb1bc376c006326cac83bcebb4b44

deploy/containers/credhub_setup/auth.go
mook-as: Moved things around so that the token URL, and other things are all passed in.  Re-reading the comments, though, I suspect I should be passing in the UAA CA cert as well; I should do that…

deploy/containers/credhub_setup/auth.go
mook-as: Fixed in 108a986313519c3b1834f1095bf0743b3cea83b8.

deploy/containers/credhub_setup/auth.go
mook-as: This is getting to the point where it seems like _every_ function call is being considered a hidden dependency? Why isn't `append()` (it allocates memory!) one? I think I may be not interpreting it correctly: is "hidden dependency" some specific jargon with non-obvious meaning?

deploy/containers/credhub_setup/auth.go
mook-as: This got refactored into oblivion… But eventually removed in 72c7993c03b86129773751ddebd5e15cbc0ddb5a

WORKSPACE
```diff
@@ -51,3 +51,32 @@ pip_install()
...
+container_repositories()
+
+load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_toolchains")
+
+go_rules_dependencies()
+
+go_register_toolchains()
+
+# gazelle:repo bazel_gazelle
+# gazelle:repository_macro deploy/containers/credhub_setup/repositories.bzl%go_repositories
```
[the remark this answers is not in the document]
f0rmiga: Gotcha.

deploy/containers/credhub_setup/auth.go
f0rmiga: `append` is not a concerning hidden dependency because we don't want to recover if it fails. Failing `append` most likely means we are about to crash the process.

If a function could genuinely fail and we need to take an action on that failure, and that function does not come from either the constructor or the parameter list, it's a hidden dependency that we should care about.

deploy/containers/credhub_setup/cmd/main.go
```diff
@@ -0,0 +1,77 @@
...
+		},
+		cc.PortInfo{
+			Addresses:   uaaAddrs,
+			Port:        uaaPort,
+			Description: "UAA service access",
+		},
+	}
+
+	err = cc.SetupCredHubApplicationSecurityGroups(ctx, client, ports)
+	if err != nil {
```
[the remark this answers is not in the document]
mook-as: I believe I didn't do it originally because the line was already kind of long, but I don't have any real objections against it either.

deploy/containers/credhub_setup/pkg/cc/secgroup.go
```diff
@@ -0,0 +1,234 @@
...
+// CF API
+type SecurityGroupDefinition struct {
+	Metadata struct {
+		GUID string `json:"guid"`
+	} `json:"metadata"`
+	Entity securityGroupEntity `json:"entity"`
+}
+
+// PortInfo describes a port to be opened
+type PortInfo struct {
```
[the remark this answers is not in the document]
mook-as: I'll do this independently of the GitHub suggestions system, then. But I'll still take it :D

deploy/containers/credhub_setup/pkg/cc/secgroup.go
```diff
@@ -0,0 +1,234 @@
...
+	return "", nil
+}
+
+// CreateOrUpdateSecurityGroup creates a new security group, or updates an
+// existing security group if one already exists.  The security group definition
+// is read from the io.Reader.
+func CreateOrUpdateSecurityGroup(ctx context.Context, client *http.Client, baseURL *url.URL, contentReader io.Reader) (string, error) {
+	groupID, err := GetExistingSecurityGroup(ctx, client, baseURL)
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
mook-as: But that was "failed to look up existing security group". (But I'll add something here.)

deploy/containers/credhub_setup/pkg/cc/secgroup.go
```diff
@@ -0,0 +1,234 @@
...
+		method = http.MethodPut
+	}
+	updateURL = baseURL.ResolveReference(updateURL)
+	req, err := http.NewRequestWithContext(ctx, method, updateURL.String(), contentReader)
+	if err != nil {
+		return "", fmt.Errorf("could not create update request: %w", err)
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("could not submit update request: %w", err)
```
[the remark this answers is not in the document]
mook-as: Hmm, using the same error string for both makes it harder to troubleshoot _which_ path went wrong (even though, yes, both are doing the same thing).  I'll at least rephrase both to `failed to…` though.

deploy/containers/credhub_setup/pkg/cc/secgroup.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
+		return "", fmt.Errorf("got response %s", resp.Status)
```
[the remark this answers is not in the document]
mook-as: Oh wow I had a terrible message here. I'll rephrase (but into something unique).

deploy/containers/credhub_setup/manifest.tmpl.yaml
```diff
@@ -0,0 +1,50 @@
+# Extra job to configure credhub access
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-credhub-setup
+  namespace: {{ .Release.Namespace | quote }}
+spec:
+  template:
+    metadata:
+      name: {{ .Release.Name }}-credhub-setup-pod
```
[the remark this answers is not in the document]
mook-as: I know I don't _need_ the uniqueness (pods are in a different namespace… not the kube namespace, the scope thing).  But it was still nice to be able to tell which was which.

deploy/containers/credhub_setup/pkg/cc/link.go
```diff
@@ -0,0 +1,37 @@
...
+	// The name of the BOSH link
+	ccEntanglementName = "cloud_controller_https_endpoint"
+)
+
+// NewHTTPClient returns a HTTP client that is set up to communicate with
+// (unauthenticated) endpoints on the cloud controller.
+func NewHTTPClient(ctx context.Context) (*http.Client, error) {
+	link, err := quarks.ResolveLink(ctx, ccEntanglementName, ccEntanglementName)
+	if err != nil {
+		return nil, fmt.Errorf("could not get link: %w", err)
```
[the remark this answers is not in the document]
mook-as: I prefer something more detailed. Probably `failed to create new HTTP client: could not get link: %w`.

deploy/containers/credhub_setup/pkg/credhub/link.go
```diff
@@ -0,0 +1,59 @@
+package credhub
+
+// credhub_info.go contains functions to determine the credhub information
+// (addresses and port) required for our security groups.
```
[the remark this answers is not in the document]
mook-as: Will remove (the whole package was just the one file, so `doc.go` covers it well enough).

deploy/containers/credhub_setup/pkg/quarks/deployment.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+)
+
+// GetDeploymentName returns the deployment name, which is embedded in some
+// names and paths.
+func GetDeploymentName(ctx context.Context) (string, error) {
+	mountRoot := GetMountRootFromContext(ctx)
+	deploymentFileName := filepath.Join(mountRoot, "run", "pod-info", "deployment-name")
```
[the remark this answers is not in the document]
mook-as: But that would be weird, since nobody else should have to know that the deployment name comes from a file (as opposed to, say, an environment variable).  I could see it being a `const`, but that's… kind of unnecessary?

deploy/containers/credhub_setup/pkg/quarks/resolver.go
```diff
@@ -0,0 +1,143 @@
...
+		baseResolver = net.DefaultResolver
+	}
+
+	var addrs []string
+	for {
+		var dnsError *net.DNSError
+		addrs, err = baseResolver.LookupHost(ctx, nameServer)
+		if err != nil && errors.As(err, &dnsError) {
+			if dnsError.Temporary() || dnsError.IsNotFound {
+				time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
mook-as: Yep, will change in a follow-up PR. (I'd just make the sleep duration a thing that we could override instead; `time.Sleep(0)` keeps tests plenty fast.)

deploy/containers/credhub_setup/pkg/quarks/resolver.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, fmt.Errorf("error looking up host %s: %w", hostname, err)
+		}
+		// If CredHub has not finished starting up, DNS resolution will fail
+		// (even though the credhub service address is fixed); wait a bit and
+		// try again until we succeed (to not blow through our retry quota and
+		// end up in CrashLoopBackoff).
+		fmt.Fprintf(os.Stderr, ".")
+		time.Sleep(10 * time.Second)
+	}
+	panic("Should not reach unreachable code")
```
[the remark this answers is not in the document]
mook-as: Yep, go happily informs me that this is unreachable code too. Just putting this here in case we modify the code later and screw up.  It's basically a static assert.

deploy/containers/credhub_setup/pkg/uaa/link.go
```diff
@@ -0,0 +1,25 @@
...
+	"context"
+	"fmt"
+
+	"credhub_setup/pkg/quarks"
+)
+
+// GetUAAAddrs retrieves the UAA IP addresses and port required for credhub
+// setup.
+func GetUAAAddrs(ctx context.Context) ([]string, int, error) {
+	// TODO: fetch this correctly
```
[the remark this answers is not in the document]
mook-as: No, because I'll need to change this in my upcoming PR anyway. I'm not _done_ this task even after this PR is merged.

deploy/containers/credhub_setup/pkg/quarks/link.go
```diff
@@ -0,0 +1,47 @@
...
+	"path/filepath"
+)
+
+// Link describes a quarks link; data can be read from it.
+type Link struct {
+	path string
+}
+
+// ResolveLink reads the quarks entanglements (BOSH links) data.
+func ResolveLink(ctx context.Context, linkType, linkName string) (*Link, error) {
```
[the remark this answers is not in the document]
mook-as: I'm confused: the callers don't have `linkPath`?

deploy/containers/credhub_setup/pkg/cc/link.go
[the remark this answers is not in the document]
f0rmiga: Sounds good!

deploy/containers/credhub_setup/pkg/cc/secgroup.go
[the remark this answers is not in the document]
f0rmiga: The way you handled it post-review is very nice!

deploy/containers/credhub_setup/pkg/quarks/link.go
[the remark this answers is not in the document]
f0rmiga: My bad, to construct `linkPath` it needs more than `linkType` and `linkName`.

f0rmiga:
Closing this for now in favour of https://github.com/cloudfoundry-incubator/kubecf/pull/434. This approach will be re-assessed post-1.0 release.

PR #387, opened 2020-02-07
Bump Helm to version 3.0.3

f0rmiga:
## Description

This PR bumps Helm to version 3 and switches from `helm template ... | kubectl apply -f -` to use `helm upgrade` and `helm delete`.

## Motivation and Context

Helm 3 is out, it's the new shiny thing. We also want to exercise `helm upgrade` and `helm delete` more from development. It also allows for easier changes to the installation on CI.

## How Has This Been Tested?

Manually installing and deleting cf-operator and kubecf + CI will take over a few other pieces.

## Types of changes

## Checklist:

### DIFF
--- a/.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
+++ b/.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
@@ -48,5 +48,8 @@ EOF
 
 chart="${workspace}/output/kubecf.tgz"
 
-# Render and apply the kubecf chart.
-bazel run @helm//:binary -- template "${chart}" --name "${KUBECF_INSTALL_NAME}" --namespace "${KUBECF_NAMESPACE}" --values "${values}" | "${KUBECTL}" apply -f -
+# Install the KubeCF chart.
+bazel run @helm//:binary -- upgrade "${KUBECF_INSTALL_NAME}" "${chart}" \
+  --install \
+  --namespace "${KUBECF_NAMESPACE}" \
+  --values "${values}"
--- a/.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
+++ b/.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
@@ -49,7 +49,8 @@ EOF
 chart="${workspace}/output/kubecf.tgz"
 
 # Install the KubeCF chart.
-bazel run @helm//:binary -- upgrade "${KUBECF_INSTALL_NAME}" "${chart}" \
+bazel run @helm//:binary -- \
+  upgrade "${KUBECF_INSTALL_NAME}" "${chart}" \
   --install \
   --namespace "${KUBECF_NAMESPACE}" \
   --values "${values}"
--- a/def.bzl
+++ b/def.bzl
@@ -18,16 +18,16 @@ project = struct(
         ),
         "helm": struct(
             sha256 = {
-                "darwin":  "34fc397ec4a992a451d130a962944315ca782242bbd05a8d732a2e74ca2b9153",
-                "linux":   "7eebaaa2da4734242bbcdced62cc32ba8c7164a18792c8acdf16c77abffce202",
-                "windows": "414d09b2559316c3dcb81cc448ba44cbbbf54a08a475998211d8dbe7217dd138",
+                "darwin":  "5e27bc6ecf838ed28a6a480ee14e6bec137b467a56f427dbc3cf995f9bdcf85c",
+                "linux":   "fc75d62bafec2c3addc87b715ce2512820375ab812e6647dc724123b616586d6",
+                "windows": "c52065cb70ad9d88b195638e1591db64852f4ad150448e06fca907d47a07fe4c",
             },
             url = {
                 "darwin":  "https://get.helm.sh/helm-v{version}-darwin-amd64.tar.gz",
                 "linux":   "https://get.helm.sh/helm-v{version}-linux-amd64.tar.gz",
                 "windows": "https://get.helm.sh/helm-v{version}-windows-amd64.zip",
             },
-            version = "2.16.1",
+            version = "3.0.3",
         ),
         "jq": struct(
             sha256 = {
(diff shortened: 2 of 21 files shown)


f0rmiga:
Ping @andreas-kupries.

.gitlab/pipelines/stages/deploy/deploy_kubecf.sh
```diff
@@ -48,5 +48,8 @@ EOF
 
 chart="${workspace}/output/kubecf.tgz"
 
-# Render and apply the kubecf chart.
-bazel run @helm//:binary -- template "${chart}" --name "${KUBECF_INSTALL_NAME}" --namespace "${KUBECF_NAMESPACE}" --values "${values}" | "${KUBECTL}" apply -f -
+# Install the KubeCF chart.
+bazel run @helm//:binary -- upgrade "${KUBECF_INSTALL_NAME}" "${chart}" \
```
mook-as: Since we're touching it already…
```suggestion
bazel run @helm//:binary -- \
  upgrade "${KUBECF_INSTALL_NAME}" "${chart}" \
```

def.bzl
```diff
@@ -4,6 +4,7 @@ Project-wide constant definitions
 project = struct(
     deployment_name = "kubecf",
     namespace = "kubecf",
+    cf_operator_namespace = "cf-operator",
```
mook-as: Yay! Separate namespace is back!

deploy/helm/kubecf/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: ~
+appVersion: v0.0.0
 description: A Helm chart for KubeCF
 name: kubecf
-version: ~
+version: v0.0.0
```
mook-as: I thought we had placeholder values here to make sure we _did_ replace them before helm sees it?

dev/cf_operator/BUILD.bazel
```diff
@@ -1,25 +1,28 @@
...
-load("//:def.bzl", "project")
+INSTALL_NAME = "cf-operator"
 
-helm_template(
-    name = "rendered_template",
-    values = glob(["**/*values.yaml"]),
-    install_name = "cf-operator",
-    namespace = project.namespace,
+helm_upgrade(
+    name = "install_or_upgrade",
```
mook-as: Doesn't `apply` still work? The name here is kind of long…

dev/kubecf/BUILD.bazel
```diff
@@ -1,23 +1,24 @@
...
+)
 load("//:def.bzl", "project")
 
-helm_template(
-    name = "rendered_template",
-    values = glob(["**/*values.yaml"]),
+package(default_visibility = ["//visibility:public"])
+
+helm_upgrade(
+    name = "install_or_upgrade",
```
mook-as: Same as above; `apply` no good?

rules/helm/delete.tmpl.rb
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env ruby
```
mook-as: Is this really better as a ruby script? Seems like shell (or even, `sh_binary`) might be good enough here.

rules/helm/delete.tmpl.rb
```diff
@@ -0,0 +1,21 @@
...
+# frozen_string_literal: true
+
+require 'open3'
+
+# Variable interpolation via Bazel template expansion.
+helm = '[[helm]]'
+install_name = '[[install_name]]'
+namespace = '[[namespace]]'
+
+stdout, stderr, status = Open3.capture3(
```
mook-as: This eats `stderr` on success…

```suggestion
exit Proess.wait2(Process.spawn(helm, 'delete', install_name, '--namespace', namespace)).last.exitstatus
```

rules/helm/upgrade.tmpl.rb
```diff
@@ -0,0 +1,43 @@
...
+  ['--values', path]
+end
+args.concat(*values)
+
+set_values = set_values.map do |key, value|
+  ['--set', "#{key}=#{value}"]
+end
+args.concat(*set_values)
+
+stdout, stderr, status = Open3.capture3(*args)
```
mook-as: See above.

rules/helm/delete.tmpl.rb
f0rmiga: I'll accept this, but I'll do it in a separate commit as I need to remove the `open3` requirement.

deploy/helm/kubecf/Chart.yaml
f0rmiga: Helm 3 doesn't like the nil values.

dev/cf_operator/BUILD.bazel
f0rmiga: `apply` was a reference to `kubectl apply`. Do you think it's fine to keep it here?

mook-as: I think that's still fine, since it's logically doing the same thing. It applies the correct configuration.

rules/helm/delete.tmpl.rb
mook-as: That's fine… I actually rarely look at the per-commit level, and just look at the PR as-is.

rules/helm/delete.tmpl.rb
f0rmiga: I wrote using Ruby for consistency with the other rule templates (even though there is still one shell script under these rules).

deploy/helm/kubecf/Chart.yaml
mook-as: Yeah, but I thought the point was to ensure we processed the files before helm got to it?

f0rmiga: No, this file is not a template. `helm package` takes the versions and apply here.

andreas-kupries:
@f0rmiga @mook-as See #396.

PR #411, opened 2020-02-26
Introduce feature_flag for `credhub`

qu1queee:
For [issue 399](https://github.com/SUSE/kubecf/issues/399)

Add conditionals in all YAMLs files where we set spec properties for `credhub`, based
on the feature flag conditional(`.Values.features.credhub.enabled`).

Added  overrides for the base `cf-deployment.yml` file, based on the current v12.18.0 version.

## Description
Adding the feature flag conditionals in all related places where we define spec properties for `credhub`. Also, because `credhub` is by default defined in the `cf-deployment.yml`, we need to remove some spec properties also there.

## Motivation and Context
The reasoning is that in IBM, we do not use the `credhub` component. Also, I will expect in the future more use cases, where `credhub` requires to be `feature_flag`.

## How Has This Been Tested?
I deploy this current PR branch in an IKS cluster two times, the first with the feature_flag disable, the second one with the feature_flag enable. Both deployments were successful, and I was able to push applications.

## Types of changes

## Checklist:


f0rmiga:
@qu1queee Thanks for the PR! Could you resolve the conflict, please?

deploy/helm/kubecf/templates/sizing.yaml
```diff
@@ -82,6 +82,8 @@ data:
       value: 1
 {{- end }}
 
+
+{{- if .Values.features.credhub.enabled -}}
```
f0rmiga:
```suggestion
{{- if .Values.features.credhub.enabled }}
```

qu1queee:
Working on this, I was on holidays during the end of last week,

f0rmiga:
I had an issue with credhub disabled and CATS enabled. It's fairly simple to fix. I'll push a commit to fix it here. Other than that, it worked fine! Thanks again.

PR #424, opened 2020-02-29
feat: allow custom certs to be used with router

### LINKED ISSUE #50: Be able to pass certificates and keys for uaa and gorouter
## Description

Currently, gorouter and UAA's certificates are auto-generated. We should be able to pass those certificates via Helm values.

f0rmiga:
## Description

Fixes #50.

## Motivation and Context

The default certificates used by the router are generated by cf-operator, therefore they are not generated by a well-known CA, making it unsuitable for production usage. This PR fixes that by allowing custom certs to be passed to the router.

## How Has This Been Tested?

Followed my own instructions passing certificates generated by `mkcert`, which is trusted on my system. Also, I asserted that the certificate used by the router was the one provided by me.

## Types of changes

## Checklist:


deploy/helm/kubecf/templates/ingress.yaml
```diff
@@ -98,7 +98,11 @@ spec:
             backend:
               serviceName: "{{ .Release.Name }}-router"
               servicePort: 443
+
+# If the ingress feature (ingress-controller) is not enabled, create public services to allow
+# public ingress to the cluster.
```
mook-as: Shouldn't this be _below_ the `{{else}}`? Otherwise it's going to get rendered into the YAML files in the wrong condition. Yes, it's a comment, but it's helpful when debugging (I often just render the ops files locally to see what is going on).

deploy/helm/kubecf/values.yaml
```diff
@@ -225,6 +225,20 @@ services:
       start: 20000
       end: 20008
 
+router:
+  # tls is an array of objects containing the `crt` and `key` keys. E.g.
```
mook-as: Hmm, would the stuttering be slightly better if we flip them ("…containing the `key` and `crt` keys.")?

doc/router_certificates.md
```diff
@@ -0,0 +1,31 @@
+# Router certificates
+
+The default KubeCF installation generates a `router_ca` and a `router_ssl` variables. The former
```
mook-as:
```suggestion
The default KubeCF installation generates the `router_ca` and `router_ssl` variables. The former
```
It sounds really odd to have "a… and a… (thing)s", even though the numbers add up to be correct, IMHO.

doc/router_certificates.md
```diff
@@ -0,0 +1,31 @@
...
+is the CA certificate that signs the later, which in turn is used in the gorouter for TLS
```
mook-as:
```suggestion
is the CA certificate that signs the latter, which in turn is used in the gorouter for TLS
```

Yes English is terrible. 😛

doc/router_certificates.md
```diff
@@ -0,0 +1,31 @@
...
+termination.
+
+For production systems, where certificates signed by a well-known Certificate Authority is required,
+KubeCF allows those certificates to be passed via Helm properties. To do so, set the values:
```
mook-as: Hmm, wouldn't _production_ systems want either ingress or external SSL termination? I guess some shops want as much traffic as possible encrypted, so they punt the decryption to the router…

doc/router_certificates.md
```diff
@@ -0,0 +1,31 @@
+# Router certificates
+
+The default KubeCF installation generates a `router_ca` and a `router_ssl` variables. The former
+is the CA certificate that signs the later, which in turn is used in the gorouter for TLS
+termination.
+
+For production systems, where certificates signed by a well-known Certificate Authority is required,
```
mook-as:
```suggestion
For production systems, where certificates signed by a well-known certificate authority is required,
```

While it's jargon, it's not a proper noun, nor a term we're defining in this legal document.

deploy/helm/kubecf/templates/ingress.yaml
f0rmiga: Oh, right. It indeed should be below.

doc/router_certificates.md
f0rmiga: I don't think so. In some setups, it would make sense to have the ingress-controller to take care of a broader set of endpoints. For that, we already have some docs and is out of scope for this PR. For other SSL termination, especially the ones terminated at the public cloud load balancers, I consider them extraneous and believe them to be out of scope for this document and especially this PR.

deploy/helm/kubecf/values.yaml
f0rmiga:
```suggestion
  # tls is an array of objects containing the `key` and `crt` keys. E.g.
```

deploy/helm/kubecf/templates/ingress.yaml
```diff
@@ -98,7 +98,11 @@ spec:
...
 {{- else }}
+
```
f0rmiga:
```suggestion

{{- else }}
# If the ingress feature (ingress-controller) is not enabled, create public services to allow
# public ingress to the cluster.

```

deploy/helm/kubecf/values.yaml
```diff
@@ -212,6 +212,27 @@ services:
       start: 20000
       end: 20008
 
+settings:
+  router:
+    # tls sets up the public TLS for the router. The tls keys:
+    #   ca: the CA in the PEM format. Only required when the certificate is self-signed.
```
mook-as:
```suggestion
    #   ca: the CA in the PEM format. Not required if the certificate is signed by a well-known authority.
```
… Because technically a cert signed by the SUSE internal CA would _not_ be self-signed, but we would still require to give the CA cert here.

deploy/helm/kubecf/values.yaml
```diff
@@ -212,6 +212,27 @@ services:
...
       end: 20008
 
+settings:
+  router:
+    # tls sets up the public TLS for the router. The tls keys:
+    #   ca: the CA in the PEM format. Not required if the certificate is signed by a well-known authority.
+    #   crt: the certificate in the PEM format. Required.
+    #   key: the private key in the PEM format. Required.
+    tls: {}
+    # ca: |
```
mook-as: Please document that, (for OpenSSL reasons) the CA cert _must_ be self-signed.  An intermediate does _not_ work here. (I will mention tomorrow morning if not supplying it works.)

mook-as: Hmm, so, for an LE cert, _not_ supplying a CA works fine.

PR #434, opened 2020-03-03
fix: allow only credhub and uaa for apps

### LINKED ISSUE #304: Remove security group that allows apps to communicate with internal network

f0rmiga:
## Description

Removes the internal security group that allows apps to communicate to the k8s service and pod IP ranges. Instead, create internal security groups allowing apps to communicate with credhub and UAA only.

## Motivation and Context

We currently have some internal app security group stuff for CredHub (from #281); this requires the administrator to set `kube.service_cluster_ip_range` and `kube.pod_cluster_ip_range` in their helm values. This is unwieldy and error-prone; we should instead auto-detect the appropriate values and set the application security groups from code.

This fixes #304.

## How Has This Been Tested?

Manually, running CATS and asserting that the credhub suite pass.

## Types of changes

## Checklist:


deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
```diff
@@ -24,16 +25,56 @@
...
+            args:
+            - -c
+            - |-
+              {{- .Files.Get "assets/scripts/jobs/cf-cli-6-linux/setup_internal_security_group.sh" | nindent 14 }}
+        envs:
+        - name: CF_CLI_PATH
+          value: /var/vcap/packages/cf-cli-6-linux/bin
+        - name: CF_API
+          value: https://api.((system_domain))
+        - name: CF_USERNAME
```
mook-as: This should be done with a client credential instead (so if the admin ever changes the password restarting the job won't be a problem).

deploy/helm/kubecf/assets/operations/instance_groups/uaa.yaml
```diff
@@ -31,6 +31,49 @@
...
+            args:
+            - -c
+            - |-
+              {{- .Files.Get "assets/scripts/jobs/cf-cli-6-linux/setup_internal_security_group.sh" | nindent 14 }}
+        envs:
+        - name: CF_CLI_PATH
+          value: /var/vcap/packages/cf-cli-6-linux/bin
+        - name: CF_API
+          value: https://api.((system_domain))
+        - name: CF_USERNAME
```
mook-as: Same as above.

deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
```diff
@@ -24,16 +25,56 @@
...
+            executable: /bin/bash
+            args:
+            - -c
+            - |-
+              {{- .Files.Get "assets/scripts/jobs/cf-cli-6-linux/setup_internal_security_group.sh" | nindent 14 }}
+        envs:
+        - name: CF_CLI_PATH
+          value: /var/vcap/packages/cf-cli-6-linux/bin
+        - name: CF_API
+          value: https://api.((system_domain))
```
mook-as: Should we be using the internal host name for CC for this? (Mostly thinking about things like "what if the external endpoint is behind a load balancer we can't reach from the inside", and "which SSL cert is that going to use")

deploy/helm/kubecf/assets/scripts/jobs/cf-cli-6-linux/setup_internal_security_group.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -o errexit -o nounset -o xtrace
+
+export PATH="${CF_CLI_PATH}:${PATH}"
+
+until curl --insecure --fail --silent --head "${CF_API}"; do
```
mook-as:
```suggestion
until curl --insecure --fail --silent --head "${CF_API}/v2/info"; do
```
We should also add a comment on _why_ poking at the API before UAA is ready is fine.

deploy/helm/kubecf/assets/scripts/jobs/cf-cli-6-linux/setup_internal_security_group.sh
```diff
@@ -0,0 +1,34 @@
...
+    sleep 1
+done
+
+cf api --skip-ssl-validation "${CF_API}"
```
mook-as: Can we do anything to avoid the `--skip-ssl-validation`?

viovanov: +1, we can't leave insecure in 
also `--silent` is a bad idea

deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
```diff
@@ -24,16 +25,56 @@
...
           registration_interval: 10s
           tags:
             component: credhub
           uris:
           - credhub.((system_domain))
-{{- end }}
 
-{{- if not .Values.features.credhub.enabled }}
+- type: replace
+  path: /instance_groups/name=credhub/jobs/-
```
viovanov: You're adding a container to a pod with an entrypoint that's not long running afaict.

f0rmiga: The idea was to also have a `trap` to clean up the security group on `EXIT`.
As we discussed on a chat, I'll change this to a job that is aware of the replicas.

deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
f0rmiga: Hmm. I'll try to connect to the internal hostname and load the CA cert to be able to remove `--skip-ssl-validation` from the `cf api` command.

deploy/helm/kubecf/assets/scripts/jobs/cf-cli-6-linux/setup_internal_security_group.sh
f0rmiga: @viovanov I guess the `--silent` is regarding the `curl` command. That command will be triggered multiple times, and we don't want it to spam logs.

For the `--skip-ssl-validation`, I have a plan for it that ties with the commend above about using the internal hostname.

deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
f0rmiga: After re-assessing with Vlad, doing it as s job brings other complexities, like adding a service that performs load balancing to the replicas and then replacing the BOSH DNS aliases. That is a direction we don't want to take at this point. I'll follow my original idea here.

mook-as: :+1: — but I'm still exploring it as a job (`post-start` + `drain` scripts), using the same `POD_IP` mechanism you're using, to see how it goes.  That's _without_ the load balancing part — it'll just be co-located with each uaa / credhub instance.

deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
f0rmiga: Created https://github.com/cloudfoundry-incubator/kubecf/issues/449.

deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
f0rmiga: Created https://github.com/cloudfoundry-incubator/kubecf/issues/450.

deploy/helm/kubecf/assets/scripts/jobs/cf-cli-6-linux/setup_internal_security_group.sh
f0rmiga: Created https://github.com/cloudfoundry-incubator/kubecf/issues/451.

PR #435, opened 2020-03-04
Remove template chunks that trim on both ends

### LINKED ISSUE #414: kubecf-scheduler service is not created.
**Describe the bug**
`cf ssh` does not work as kubecf/kubecf-scheduler service is not created.

**To Reproduce**
Deploy cfo and kubecf with Ingress Controller. Do the TCP forwarding as indicated in the deployment instructions for ssh-proxy, e.g. 

```helm install stable/nginx-ingress \
  --name ingress \
  --namespace ingress \
  --set "tcp.2222=kubecf/kubecf-scheduler:2222" \
```

**Expected behavior**
`kubecf-scheduler` service should be created in the kubecf namespace.

**Environment**
`cf-operator-2.3.0+0.g27a91cdf`
`kubecf-v0.0.0-e7534b6.tgz`

**Additional context**

### LINKED ISSUE #437: Linter for {{- -}} in Go templates

As seen in:

https://github.com/cloudfoundry-incubator/kubecf/blob/63d9b24462f7e6a74ffe73ef625d46849e5761f4/deploy/helm/kubecf/assets/operations/instance_groups/scheduler.yaml#L57-L60

we cannot have opening `{{-` with closing `-}}`. The `-` will chop the blanks.

We need a linter to check that we don't end up having this. The suggestion should be always opening `{{-` and closing `}}`.

No alternatives.

**Additional context**

No additional context.

viovanov:
## Description
Using both `{{-` and `-}}` causes line breaks on both ends to be removed, sometimes resulting in unintended consequences.

e.g. 
[This code](https://github.com/cloudfoundry-incubator/kubecf/blob/63d9b24462f7e6a74ffe73ef625d46849e5761f4/deploy/helm/kubecf/assets/operations/instance_groups/scheduler.yaml#L57-L60) results in the following render:

```
- type: replace
  path: /instance_groups/name=scheduler/jobs/name=ssh_proxy/properties/quarks?
  value:
    ports:
    - name: ssh-proxy
      protocol: TCP
      internal: 2222
    run:
      healthcheck:
        ssh_proxy:
          readiness:
            httpGet:
              port: 2223

# Add necessary labels to the scheduler instance group so that the service can select it to create
# the endpoint.- type: replace
  path: /instance_groups/name=scheduler/env?/bosh/agent/settings/labels/app.kubernetes.io~1component
  value: "ssh-proxy"
```

The array dash is no more, and the path and value keys override the ones from the previous element.

## Motivation and Context
Fixes #414.
Fixes #437.

## How Has This Been Tested?
The change contains a script that lints for this, so it doesn't happen again.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/buildpacks/set_suse_buildpacks.yaml
+++ b/deploy/helm/kubecf/assets/operations/buildpacks/set_suse_buildpacks.yaml
@@ -17,7 +17,7 @@
 {{- $buildpacks = append $buildpacks "suse-nginx-buildpack" }}
 {{- $buildpacks = append $buildpacks "suse-binary-buildpack" }}
 
-{{- $root := . -}}
+{{- $root := . }}
 {{- range $buildpack := $buildpacks }}
 - type: replace
   path: /releases/-
--- a/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.testing.cf_acceptance_tests.enabled -}}
+{{- if .Values.testing.cf_acceptance_tests.enabled }}
 
 - path: /instance_groups/-
   type: replace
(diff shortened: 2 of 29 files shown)


f0rmiga:
I deployed 2 variations of this PR:
- with credhub and diego.
- without credhub, with eirini.
Both deployed fine and passed smoke-tests. With diego, it also passed CATS. With eirini, it fails CATS, but nothing related to this PR.
I'll go ahead and merge while CI is not ready for Helm 3.

PR #440, opened 2020-03-04
fix: helm: reduce duplication in sizing ops

mook-as:
## Description
This makes [`sizing.yaml`] repeat itself less, like [`releases.yaml`].

[`sizing.yaml`]: https://github.com/cloudfoundry-incubator/kubecf/blob/master/deploy/helm/kubecf/templates/sizing.yaml
[`releases.yaml`]: https://github.com/cloudfoundry-incubator/kubecf/blob/master/deploy/helm/kubecf/assets/operations/releases.yaml

## Motivation and Context
This makes sure we don't end up with a copy/paste error somewhere in _one_ sizing block: we move all the complicated logic on setting the sizes to one shared block, where we can focus our attention, and adding a new instance group is much simpler now.

As a side effect, also makes it possible to set size to 0 (since I'm now doing an explicit nil check).

## How Has This Been Tested?
Rendered the ops file manually (via `helm template`) to check that the output looks reasonable.  Checked HA mode (a much reduced output), setting explicit sizes on `api` and `cc-worker` (to test the snake case thing).

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/templates/sizing.yaml
+++ b/deploy/helm/kubecf/templates/sizing.yaml
@@ -13,139 +13,42 @@ metadata:
     helm.sh/chart: {{ include "kubecf.chart" . }}
 data:
   ops: |
-{{- if .Values.sizing.adapter.instances }}
-    - type: replace
-      path: /instance_groups/name=adapter/instances
-      value: {{.Values.sizing.adapter.instances}}
-{{- else if not .Values.high_availability }}
-    - type: replace
-      path: /instance_groups/name=adapter/instances
-      value: 1
-{{- end }}
+    {{- $instance_groups := list }}
 
-{{- if .Values.sizing.api.instances }}
-    - type: replace
-      path: /instance_groups/name=api/instances
-      value: {{.Values.sizing.api.instances}}
-{{- else if not .Values.high_availability }}
-    - type: replace
-      path: /instance_groups/name=api/instances
-      value: 1
-{{- end }}
+    {{- /* Core instance groups that always exist */}}
+    {{- $instance_groups = append $instance_groups "adapter" }}
+    {{- $instance_groups = append $instance_groups "api" }}
+    {{- $instance_groups = append $instance_groups "auctioneer" }}
+    {{- $instance_groups = append $instance_groups "cc-worker" }}
+    {{- $instance_groups = append $instance_groups "diego-api" }}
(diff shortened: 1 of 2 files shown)


deploy/helm/kubecf/templates/sizing.yaml
```diff
@@ -13,212 +13,58 @@ metadata:
...
+    {{- $instance_groups = append $instance_groups "uaa" }}
+    {{- $instance_groups = append $instance_groups "tcp-router" }}
+
+    {{- /* Autoscaler-only instance groups */}}
+    {{- if .Values.features.autoscaler.enabled }}
+    {{- $instance_groups = append $instance_groups "asapi" }}
+    {{- $instance_groups = append $instance_groups "asactors" }}
+    {{- $instance_groups = append $instance_groups "asmetrics" }}
+    {{- $instance_groups = append $instance_groups "asnozzle" }}
+    {{- end }}{{/* if autoscaler */}}
```
f0rmiga: I wonder if this works:

```suggestion
    {{- end /* if autoscaler */}}
```

deploy/helm/kubecf/templates/sizing.yaml
```diff
@@ -13,212 +13,58 @@ metadata:
...
-      path: /instance_groups/name=uaa/instances
-      value: {{.Values.sizing.uaa.instances}}
-{{- else if not .Values.high_availability }}
-    - type: replace
-      path: /instance_groups/name=uaa/instances
-      value: 1
-{{- end }}
-
-{{- if .Values.sizing.tcp_router.instances }}
+{{- $root := . }}
```
f0rmiga: Instead of `$root := .`, we can access the root with `$.Release.Name` for example.

deploy/helm/kubecf/templates/sizing.yaml
mook-as: It does not. Comments _must_ start right after the brace, unless (and only if) it's preceded by the whitespace-eating modifier. That is, `{{/*` and `{{- /*` only.

deploy/helm/kubecf/templates/sizing.yaml
mook-as: Changed in 90c8a63399c4c1c0082a58ba14ddaa4f72d71b3a.

PR #445, opened 2020-03-05
Bump catapult to support helm3

jimmykarily:
Also pin helm3 version in the pipeline

Signed-off-by: Dimitris Karakasilis <DKarakasilis@suse.com>

master branch no longer works with helm 2. This PR is adding CI support for helm 3.

### DIFF
--- a/.concourse/pipeline.yaml
+++ b/.concourse/pipeline.yaml
@@ -52,7 +52,7 @@ resources:
   source:
     uri: https://github.com/SUSE/catapult
   version:
-    ref: eb4b8fe1453c7a5f1c6a2082e5f490dd6e953664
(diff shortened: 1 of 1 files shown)


.concourse/pipeline.yaml
```diff
@@ -97,6 +97,7 @@ deploy_args: &deploy_args
   export SCF_CHART="$(readlink -f s3.kubecf-ci/*.tgz)"
   export SCF_OPERATOR=true
   export FORCE_DELETE=true
+  export HELM_VERSION="v3.1.1"
```
jimmykarily: @f0rmiga we tried to find some bazel command to extract this value from [here](https://github.com/cloudfoundry-incubator/kubecf/blob/master/def.bzl#L31) but we couldn't. Help is welcome.

Issue #450, opened 2020-03-05
Use a client credential instead of admin password while setting up security groups


f0rmiga:
Following up on https://github.com/cloudfoundry-incubator/kubecf/pull/434/files#r387252584.

fargozhu:
PR merged

f0rmiga:
@fargozhu No, this ticket is _following up_ on the merged PR.

PR #459, opened 2020-03-06
fix: use client credentials over admin password

f0rmiga:
## Description

Fixes https://github.com/cloudfoundry-incubator/kubecf/issues/450.

## Motivation and Context

If the admin password is changed, the credhub and uaa setups will fail.

## How Has This Been Tested?

Locally, checked that the security groups were created as expected not using the admin password.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
@@ -60,11 +60,11 @@
               name: {{ printf "%s.var-router-ssl" .Release.Name | quote }}
               key: ca
         - name: CF_USERNAME
-          value: admin
+          value: credhub_setup
         - name: CF_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: {{ printf "%s.var-cf-admin-password" .Release.Name | quote }}
+              name: {{ printf "%s.var-credhub-setup-client-secret" .Release.Name | quote }}
               key: password
         - name: POD_IP
           valueFrom:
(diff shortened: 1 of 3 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
```diff
@@ -60,11 +60,11 @@
               name: {{ printf "%s.var-router-ssl" .Release.Name | quote }}
               key: ca
         - name: CF_USERNAME
-          value: admin
+          value: credhub_setup
         - name: CF_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: {{ printf "%s.var-cf-admin-password" .Release.Name | quote }}
+              name: {{ printf "%s.var-credhub-setup-client-secret" .Release.Name | quote }}
```
mook-as: Looks ugly, but I guess because we don't have a good way to reference the secret using operator things… we can't just `((credhub_setup_client_secret))` because then it isn't treated as a secret anymore :(

deploy/helm/kubecf/assets/operations/instance_groups/uaa.yaml
```diff
@@ -60,11 +60,11 @@
               name: {{ printf "%s.var-router-ssl" .Release.Name | quote }}
               key: ca
         - name: CF_USERNAME
-          value: admin
+          value: uaa_setup
```
mook-as: Do we need a separate account for this? Is `credhub_setup` no good for this?  I realize that this is in the UAA group, but it's only used if credhub is enabled (and is only used to help credhub).

deploy/helm/kubecf/assets/operations/instance_groups/credhub.yaml
f0rmiga: Yep.

deploy/helm/kubecf/assets/operations/instance_groups/uaa.yaml
f0rmiga: Even though the outcome of this setup is to help credhub, there's no credhub specifics here other than it being on a conditional block only if credhub is enabled. I'd prefer to keeping it this way.

mook-as: Yeah, I was thinking about not needing more credentials around that could get leaked, but I guess not _really_ since they should be similar in difficulty to exfiltrate.

PR #465, opened 2020-03-09
Add Helm templates to enable KubeCF with native Eirini

herrjulz:
"Native" Eirini here means the standalone Helm Chart that the official Eirini team provides to deploy Eirini along a CF on Kubernetes.

## Description

We added additional Helm charts and ops files to the helm release. These files disable Diego and set all the required Eirini properties as well as required certificates.

What this PR does not include is the copy of required certificates from the KubeCF to the Eirini apps namespace.

In Eirini we used to have a job (called "secret-smuggler") for SCF which does this, but as this is not a generic approach for different CF deployments on top of Kubernetes (eg CF-for-K8s) we decided to remove it completely and let whoever is consuming Eirini handle this. We don't have enough context on KubeCF on how to do it in a native way, so if you have any suggestions let us know. If it turns out to be a manual step, it should be documented somewhere.

These changes will work with the upcoming Eirini-Release (~v1.4.0) or with any eirini-release branch containing the following commits: 
https://github.com/cloudfoundry-incubator/eirini-release/commit/711df88387e401bd08d8432d0e3b25b3eff992b3 and https://github.com/cloudfoundry-incubator/eirini-release/commit/609640ad43009dec187d59cd1e4f367771e7d2ff

## Motivation and Context
The eirini-bosh-release is not supported by the official Eirini team (or not supported at all). So in order to consume the latest-and-greatest Eirini, using our Helm-Chart is the way. 

## How Has This Been Tested?
We deployed it with KubeCF version `0.2.0-1ed5484`, CF-Operator version `v2.0.0-0` on IKS. We ran the set of CF-CATs that Eirini usually runs successfully.

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [ x ] My code follows the code style of this project.
- [ x ] My change requires a change to the documentation.

Thx to @HeavyWombat for helping us on this!

/cc @gdankov

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml
@@ -1,56 +1,35 @@
-{{- if .Values.features.eirini-native.enabled }}
-- type: remove
-  path: /instance_groups/name=diego-cell
-
-# Remove bbs from diego-api.
-# TODO: remove bbs in the future - when the clock and worker no longer need it.
-# - type: remove
-#   path: /instance_groups/name=diego-api/jobs/name=bbs
-
-# Remove auctioneer, tps and ssh_proxy from scheduler.
-- type: remove
-  path: /instance_groups/name=auctioneer
-- type: remove
-  path: /instance_groups/name=scheduler/jobs/name=tps
-- type: remove
-  path: /instance_groups/name=scheduler/jobs/name=ssh_proxy
-
-# Enable OPI in CC
-# cloud_controller_ng
+{{- if and (.Values.features.eirini.enabled) (.Values.features.eirini.use_helm_release) }}
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/opi?/enabled?
   value: true
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/opi?/url?
-  value: https://eirini-opi.{{ .Release.Namespace }}:8085 
+  value: "https://eirini-opi.{{ .Release.Namespace }}.svc.cluster.local:8085"
 
-  # --- Set endpoints and alternative names for eirini staging.
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/packages?/webdav_config?/private_endpoint?
-  value: https://{{ .Release.Name }}-singleton-blobstore.{{ .Release.Namespace }}:4443
+  value: "https://{{ .Release.Name }}-singleton-blobstore.{{ .Release.Namespace }}.svc.cluster.local:4443"
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/buildpacks?/webdav_config?/private_endpoint?
-  value: https://{{ .Release.Name }}-singleton-blobstore.{{ .Release.Namespace }}:4443
+  value: "https://{{ .Release.Name }}-singleton-blobstore.{{ .Release.Namespace }}.svc.cluster.local:4443"
 - type: replace
   path: /variables/name=blobstore_tls/options/alternative_names?/-
-  value: "{{ .Release.Name }}-singleton-blobstore.{{ .Release.Namespace }}"
+  value: "{{ .Release.Name }}-singleton-blobstore.{{ .Release.Namespace }}.svc.cluster.local"
 - type: replace
   path: /variables/name=cc_bridge_cc_uploader_server/options/alternative_names?/-
-  value: "{{ .Release.Name }}-api.{{ .Release.Namespace }}"
+  value: "{{ .Release.Name }}-api.{{ .Release.Namespace }}.svc.cluster.local"
 - type: replace
   path: /instance_groups/name=api/jobs/name=cc_uploader/properties/internal_hostname?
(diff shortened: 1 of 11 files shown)


mudler:
Thanks :100: ! 

CI seems to fail on Linting:

```
+ ./dev/linters/helmlint.sh

./deploy/helm/kubecf/assets/operations/instance_groups/eirini-native.yaml:{{- if .Values.features.eirini-native.enabled -}}

Found double minus templates {{- ... -}}
```

(see https://github.com/cloudfoundry-incubator/kubecf/pull/435 for the rationale)

herrjulz:
@mudler thx. I just removed the dash and pushed. It should be ok now.

herrjulz:
Hi again. We created a gist with all the values files we used when deployed Eirini with KubeCF for this PR. We thought it might be helpful for you to reproduce what we did.

# Deploy Steps

1. Deploy KubeCF
2. Deploy Eirini Helm Release
3. Deploy Bits-Service  Helm Release (v2.34)
4. Copy Certificates 

See all required values files in [this gist](https://gist.github.com/herrjulz/09a1d2e224bf27cf0e51115c148d4353)

viovanov:
@herrjulz we'd like to also integrate the eirini helm chart as a sub chart
Can we expect that as a separate PR?

herrjulz:
@viovanov we can add it for the initial version as soon as we got the eirini helm chart that includes the required changes released. However, this might be a problem for Bits-Service as the Bits-Service requires the `blobstore user password` as an input on helm deploy and this password is generated by KubeCF after it was deployed. This could only work if there is some kind of ability to set the blobstore password upfront?

f0rmiga:
It's somewhat concerning that the blobstore password ends up being in a configmap: https://github.com/cloudfoundry-incubator/bits-service-release/blob/ceef5b1fc7c394ed9dc7190cfdd446ec8b1ab429/helm/bits/templates/bits.yaml#L65.

Ideally, bits-service would allow the `secrets` values to be passed as a secret name that it mounts as a volume and generate the final config file using an init container here: https://github.com/cloudfoundry-incubator/bits-service-release/blob/ceef5b1fc7c394ed9dc7190cfdd446ec8b1ab429/helm/bits/templates/bits.yaml#L92.

Then it joins the config file that **has no secrets** here https://github.com/cloudfoundry-incubator/bits-service-release/blob/ceef5b1fc7c394ed9dc7190cfdd446ec8b1ab429/helm/bits/templates/bits.yaml#L109-L110 with the file mounted from the secret into an ephemeral volume.

Assuming we are able to PR this for the bits-service, then we can pass the secret name to the sub-chart and it will be forced to wait for the cf-operator to generate the secret. We do something similar with the current database and database seeder. This wouldn't require any manual step (prone to fat fingering).

herrjulz:
@viovanov @f0rmiga That is definitely a better approach, but we would like to avoid init containers as much as possible. So our idea is to modify the bits-service a bit such that it is able to consume the blobstore password from a file or env variable (as we do [here](https://github.com/cloudfoundry-incubator/eirini-release/blob/4676ded2e437bcf42c633125f26748b2d35ec5fc/helm/eirini/templates/routing.yaml#L36-L40)). That way it will still be forced to wait for the secret to become available to come up but we avoid init containers.

f0rmiga:
@herrjulz If we can modify the bits-service to do it, perfect. Fewer steps that could go wrong. :)

gdankov:
@f0rmiga We already updated Bits to consume the password from env variable. 🙂  This change is in the `2.36.0` version of the bits helm chart. 

We are currently preparing a follow-up PR to add bits and eirini as conditional sub-chart for the kubecf chart.

Regards,
Georgi && @herrjulz

f0rmiga:
@gdankov Awesome! Looking forward to it!

f0rmiga:
@herrjulz Do we need to pick anything from this PR, or can we close it completely in favour of https://github.com/cloudfoundry-incubator/kubecf/pull/477?

gdankov:
@f0rmiga You need both. 

#477 is only for consuming Eirini & Bits, while this one is for setting all the necessary plumbing between kubecf & eirini.

f0rmiga:
@gdankov I'll mark this as blocked by https://github.com/cloudfoundry-incubator/kubecf/pull/477, then.

deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml
```diff
@@ -0,0 +1,116 @@
+{{- if and (.Values.features.eirini.enabled) (.Values.features.eirini.use_helm_release) }}
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/opi?/enabled?
+  value: true
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/opi?/url?
+  value: "https://eirini-opi.{{ .Release.Namespace }}.svc.cluster.local:8085"
```
f0rmiga: The cluster domain, in this case `cluster.local`, can be set to something else in different clusters.
To increase compatibility, you have to leave `cluster.local` out.

```suggestion
  value: "https://eirini-opi.{{ .Release.Namespace }}.svc:8085"
```

Just let the DNS server resolve it for you.

jimmykarily:
Superseded by: https://github.com/cloudfoundry-incubator/kubecf/pull/651

PR #477, opened 2020-03-12
Consume Eirini & Bits service as subcharts

gdankov:
## Description
This PR adds Eirini and Bits service as conditional sub-charts to kubecf. These sub-charts will only be installed if the property `features.eirini_native.enabled` is set to `true`.

This PR should be merged after #465

## Motivation and Context
See #465 

## How Has This Been Tested?
We deployed a patched kubecf that includes these changes (based on version `0.2.0`) and cf-operator version `2.0.0` on IKS.

## Types of changes

## Checklist:

/cc @herrjulz

### DIFF
--- a/deploy/helm/kubecf/values.yaml
+++ b/deploy/helm/kubecf/values.yaml
@@ -237,6 +237,7 @@ settings:
 features:
   eirini:
     enabled: false
+    use_helm_release: false
     registry:
       service:
         nodePort: 32123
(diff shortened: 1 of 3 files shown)


deploy/helm/kubecf/values.yaml
```diff
@@ -328,3 +330,124 @@ k8s-host-url: ""
...
+    ingress_endpoint: ~
+    kube:
+      external_ips: []
+    services:
+      loadbalanced: false
+
+    deny_app_ingress: false
+
+    cc_api:
+      serviceName: "kubecf-api"
```
f0rmiga: The `kubecf` prefix here and below only works incidentally due to the Helm installation name being `kubecf`. I.e. if someone deploys KubeCF with `helm install kubecf-experiment ...`, it blows up!

f0rmiga: I don't have a solution for this... we need to think.

deploy/helm/kubecf/values.yaml
```diff
@@ -328,3 +330,124 @@ k8s-host-url: ""
 k8s-service-token: ""
 k8s-service-username: ""
 k8s-node-ca: ""
+
+eirini:
```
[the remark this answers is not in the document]
f0rmiga: Hmmm, not sure if we can actually do it.

deploy/helm/kubecf/values.yaml
```diff
@@ -240,6 +240,8 @@ features:
     registry:
       service:
         nodePort: 32123
+  eirini-native:
```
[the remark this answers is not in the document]
herrjulz: @f0rmiga we will probably call the property something like `eirini.helm` as you suggested.

deploy/helm/kubecf/values.yaml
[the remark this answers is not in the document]
herrjulz: I'm not sure if it is really possible. As the top-level property `eirini` relates to the subchart name (which is eirini). But maybe there is a way around it?
Another possibility would be to keep this config in the eirini chart only. But that way you would have to specify the properties all the time.

deploy/helm/kubecf/values.yaml
herrjulz: This is indeed true. The default is set to `kubecf`, because in general the it is deployed with this release name. Users could overwrite the configuration in case the release name deviates. Which is not nice.
The problem might be solvable by not setting the secret names based on the release-name, but have a common prefix for the charts (which could be kubecf). We were wondering why the kubecf helm charts are designed like this in first place? Can this be changed?

deploy/helm/kubecf/values.yaml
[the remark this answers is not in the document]
f0rmiga: Yeah, it doesn't seem to be possible. Having to specify the properties all the time definitely wouldn't work. :sweat_smile:

deploy/helm/kubecf/values.yaml
f0rmiga: @mook-as pointed out to me that `.Release.Name` is still available on the sub-chart. If we change the sub-charts to call the `tpl` function on the values, we can do magic like:

```yaml
# -----------------------------------
# Sub-chart bar
apiVersion: v1
kind: Secret
metadata:
  name: {{ tpl .Values.secretName . }}

# -----------------------------------
# values.yaml of foo
bar:
  secretName: '{{ printf "%s.var-tls" .Release.Name }}'
```

`helm template foo-experiment .` renders to:

```yaml
---
# Source: foo/charts/bar/templates/secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: foo-experiment.var-tls
```

This would require the Eirini and bits-service Helm charts to allow this trick.

gdankov: Hey @f0rmiga! I think from our point of view we would like to change as little as possible in the Eirini helm chart and make it as generic as possible. Our current goal is to detach from all cf-on-top-of-k8s deployment methods so we (1) can concentrate on providing a concise Eirini helm chart and (2) can be easily consumed by whoever. Having said that, I am a bit reluctant to make such implementation-specific changes to our helm chart because I don’t know how this will play out with cf-for-k8s (which we also support).

Have you considered removing the `.Release.name` prefix from all kubecf components? This seems like the proper approach to me. I’m not sure what value having this prefix has for you. It might also cause similar issues in the future when you consume other sub-charts.

f0rmiga: Hey! Adding `tpl` wouldn't cause any side-effects for other users. But I understand that such a change is not normal and that we most likely will have problems with other charts.

@viovanov Could we revisit the prefixes added by cf-operator to k8s resources?

viovanov: @f0rmiga we could - those are "bosh deployment names" though, not helm release names.

viovanov: https://www.pivotaltracker.com/story/show/171854659

I wouldn't block this on this issue @f0rmiga

herrjulz:
@f0rmiga @viovanov We updated both PRs to consume the eirini-helm release with the property `features.eirini.use_helm_release` enabled, which requires to have `features.eirini.enabled` to be set to true as well.

deploy/helm/kubecf/values.yaml
f0rmiga: @viovanov, the BOSH deployment name and the Helm release name happen to match for us. Thanks for creating the story. I'll add the blocked label.

f0rmiga:
Blocked on https://www.pivotaltracker.com/story/show/171854659.

f0rmiga:
@herrjulz Thanks for updating them!

viovanov:
@f0rmiga would it be possible to alter kubecf to ensure that the deployment name is always the namespace name? 

then, @gdankov @herrjulz - you should have access to the namespace name, correct? Would it be possible to use that?

Just as a workaround until https://www.pivotaltracker.com/story/show/171854659 is complete (since it may take us a bit of time to implement)

gdankov:
> then, @gdankov @herrjulz - you should have access to the namespace name, correct? Would it be possible to use that?

Do you mean using the namespace name when we reference secrets/service names directly in the Eirini helm chart? If so, sadly this wouldn't work for us, because it will break our compatibility with cf-for-k8s. We can't think of a good workaround which still depends on dynamic names for these components. If you have any other ideas, please let us know. 🙂 

Georgi && @schaefm-ibm

f0rmiga:
@gdankov @herrjulz This PR should be unblocked by https://github.com/cloudfoundry-incubator/kubecf/pull/581 now. We don't have the prefixes to care about anymore!

jimmykarily:
Superseded by: https://github.com/cloudfoundry-incubator/kubecf/pull/651

PR #479, opened 2020-03-12
Use pxc image from the cfcontainerization org

aduffeck:
The pxc image lives in the cfcontainerization org like all the other images now.


bikramnehra:
@aduffeck actually you can get rid of this param altogether and let it use the defaults: https://github.com/cloudfoundry-incubator/kubecf/blob/master/deploy/helm/kubecf/values.yaml#L41

Also, we should use `version` and not `tag` see other examples.

PR #489, opened 2020-03-16
Use a PVC for diego cell ephemeral disks

### LINKED ISSUE #441: /var/vcap/data on diego-cell needs to be a separate volume with quota support
# The problem

garden and grootfs both assume that they together have exclusive control over the disk space mounted at `/var/vcap/data`. The property `grootfs.reserved_space_for_other_jobs_in_mb` (default: 15360) can be used to carve out space for other jobs.

Currently an `emptyDir` volume is mounted to various parts of the filesystem. I don't understand what purpose this serves, as the disk space will still be allocated on the local node storage. Yes, it survives a pod restart, but I'm not even sure this is desirable.

The real issue though is that now grootfs is going to allocate sparse files as backing store for the privileged and unprivileged stores that claim all the size except for 15360MB of the local node storage, and formats those sparse files with XFS.

There is more to this, including the way garden calculates available disk space for apps, and how it does garbage collection for the layer cache. That code assumes that `/var/vcap/data/garden/depot` is on the same disk as `/var/vcap/data` and the grootfs stores.

# Potential fix

Use a PVC and mount it to `/var/vcap/data`. This still requires that the storageclass for this PVC implements quotas correctly. The NFS test setup used during development typically does not, so would still allocate huge stores.

This is the mechanism SCF uses to move the grootfs away from the local node disk. For some reason we found it necessary to wipe this PVC every time the container restarts. I don't know why this was necessary, but the current `emptyDir` volume used by kubecf would keep data between restarts and only be wiped when the pod is deleted.

# Alternative

Assuming that we can require worker nodes to have sufficient disk space for the grootfs, we might be able to set a `limit.disk` for the pod to get a quota on the pod filesystem. That way we could get rid of the volume mount completely.

This would also provide faster storage (just like the current `emptyDir` setup) in case the node uses e.g. SSD drives.

viovanov:
## Motivation and Context
Fixes #441 

## Types of changes

## Checklist:

Needs https://github.com/cloudfoundry-incubator/cf-operator/pull/852


deploy/helm/kubecf/values.yaml
```diff
@@ -172,6 +172,10 @@ sizing:
   diego_api:
     instances: ~
   diego_cell:
+    # Size of the ephemeral disk used to store applications
+    ephemeral_disk_size: 40960
```
f0rmiga: How about:
```yaml
diego_cell:
  ephemeral_disk:
    size: ...
    storage_class: ...
```

jandubois:
Running with an NFS storage class I ran into the following error when pushing an app:

```
Cell diego-cell-0 failed to create container for instance bce46e29-9dbc-4497-95f5-138cb0f54fbf: runc run: exit status 1: container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"/var/vcap/data/rep/shared/garden/download_cache/d526d817085c861ebec065cc8e10b4ec-1584494029961314460-1.d\\\" to rootfs \\\"/var/vcap/data/grootfs/store/unprivileged/images/bce46e29-9dbc-4497-95f5-138cb0f54fbf/rootfs\\\" at \\\"/tmp/buildpacks/d526d817085c861ebec065cc8e10b4ec\\\" caused \\\"stat /var/vcap/data/rep/shared/garden/download_cache/d526d817085c861ebec065cc8e10b4ec-1584494029961314460-1.d: permission denied\\\"\""
```

In SCF we changed permissions on the PVC mountpoint in a pre-start script (and wiped the directory in case of a pod restart):

```
$ cat container-host-files/etc/scf/config/scripts/cleanup-garden-graph.sh
#!/bin/bash
set -o errexit -o nounset -o xtrace

rm -rf /var/vcap/data/grootfs/*

# Ensure that runc and container processes can stat everything
chmod ugo+rx /var/vcap/data/grootfs
```

I suspect we need to do the same for kubecf as well, but for `/var/vcap/data`.

jandubois:
@viovanov I've tested this PR on AKS and everything works as expected.

I've looked also some more at the permission issue when using an NFS based storage class, and I don't understand what is causing the error during staging. The directories are already world readable/executable, and when I exec into the diego-cell container, I can run the `stat` command without error, both as the `root` and as the `vcap` user.

I compared directory permissions with the AKS setup (which didn't throw the permissions error), and the only difference I detected was that everything on the PVC had the SETGUID bit set. I changed this on the NFS PVC to match, but don't understand why it would be relevant. It didn't make any difference, so it probably doesn't. 😄 

Any ideas?

viovanov:
Needs #560

deploy/helm/kubecf/assets/operations/instance_groups/diego-cell.yaml
```diff
@@ -94,6 +143,11 @@
 - type: replace
   path: /instance_groups/name=diego-cell/jobs/name=rep/properties/quarks?
   value:
+    # Disable the ephemeral rep disk
+    bpm:
+      processes:
+      - name: rep
+        ephemeral_disk: false
```
jandubois: I'm not sure why this is needed. I assume is is necessary to avoid the operator from provisioning an *additional* volume because you already defined one earlier via the `env/agent/settings/disks` modifications.

If there is another reason, please let me know!

deploy/helm/kubecf/values.yaml
```diff
@@ -175,6 +175,11 @@ sizing:
   diego_api:
     instances: ~
   diego_cell:
+    ephemeral_disk:
+      # Size of the ephemeral disk used to store applications
```
jandubois: Comment could say that the value is supposed to be in MB, but maybe this is obvious from the default value.

fargozhu:
> Needs #560

@viovanov PR merged...

PR #490, opened 2020-03-17
CI: Bump catapult version to latest

mudler:
- Bump Catapult version to latest to include brain tests fixes https://github.com/SUSE/catapult/commit/543bc71abc8bd22e5ef81fdf94a7afeb2f99066d
- Fixes a leftover from the rotation test that hardcodes the deployment name, while there is a variable for it to consume

### DIFF
--- a/.concourse/pipeline.yaml
+++ b/.concourse/pipeline.yaml
@@ -52,7 +52,7 @@ resources:
   source:
     uri: https://github.com/SUSE/catapult
   version:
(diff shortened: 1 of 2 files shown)


viovanov:
@mudler @jimmykarily 
no labels and no description - a bit hard to tell what this is for

jimmykarily:
It will improve when catapult [gets proper versioning](https://github.com/SUSE/catapult/issues/137). I think Ettore added a description for now.

PR #498, opened 2020-03-18
fix: disable brain tests on CI

f0rmiga:
The updates to `cleanup-eirini-cluster` were left behind. This PR fixes it.
This is a follow-up on https://github.com/cloudfoundry-incubator/kubecf/pull/497.


f0rmiga:
> Do we not need a matching diego one?

No, diego is doing the clean-up after the correct job already.

PR #538, opened 2020-03-23
Templatize concourse pipeline

jimmykarily:
## Description
Make it possible to easily deploy a copy of the concourse pipeline pointing to different git repositories and S3 buckets.

## Motivation and Context
People contributing to this project may want to:

- Deploy their own version of the pipeline, consuming their own hardware resources to speed up development.
- Deploy a copy of the pipeline in order to make changes to the pipeline itself before opening a PR.

## How Has This Been Tested?
Run the "fly" wrapper script with no config.yaml in place. It deploys the same pipeline that is currently deployed. When run with a config.yaml, it makes the relevant adjustments.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/README.md
+++ b/.concourse/README.md
@@ -48,8 +48,14 @@ point to your new EKCP API endpoint.
 ## Deploy on Concourse
 
 If you wish to deploy the pipeline on Concourse, run the following
-command:
+command and use the `fly` script in this directory:
 
 ```
-fly -t target set-pipeline -c pipeline.yaml -p kubecf
+./fly -t target set-pipeline -c pipeline.yaml -p kubecf
 ```
+
+## Pipeline development
+
+If you wish to deploy a copy of this pipeline without publishing artifacts to
+the official buckets, create a `config.yaml` file ( you can use `config.yaml.sample`
+as a guide ) and deploy the same command above.
--- a/.concourse/README.md
+++ b/.concourse/README.md
@@ -48,10 +48,10 @@ point to your new EKCP API endpoint.
 ## Deploy on Concourse
 
 If you wish to deploy the pipeline on Concourse, run the following
-command and use the `fly` script in this directory:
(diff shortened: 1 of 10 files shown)


.concourse/README.md
```diff
@@ -48,8 +48,14 @@ point to your new EKCP API endpoint.
 ## Deploy on Concourse
 
 If you wish to deploy the pipeline on Concourse, run the following
-command:
+command and use the `fly` script in this directory:
 
 ```
-fly -t target set-pipeline -c pipeline.yaml -p kubecf
+./fly -t target set-pipeline -c pipeline.yaml -p kubecf
```
bikramnehra: Can't we have `fly` binary in `$PATH` to maintain status quo.?

.concourse/config.yaml.sample
```diff
@@ -0,0 +1,11 @@
+# All null (~) values have defaults in the pipeline
```
bikramnehra: How can we absolutely make sure that there is no/minimal possibility for secret leaks? As the pipeline might be having public access.

.concourse/README.md
```diff
@@ -48,8 +48,14 @@ point to your new EKCP API endpoint.
...
 ```
+
+## Pipeline development
```
bikramnehra: Cool. Good to have an example :+1:

.concourse/README.md
jimmykarily: This `fly` here is the script we include in this PR. It's wrapping fly to render the template. See the fly script below.

.concourse/config.yaml.sample
jimmykarily: We removed the secrets from the values here. You are right, we don't need to templetize them. If someone is deploying on a different concourse instance, they will simpy add their own secrets in their secret storage solution.

jimmykarily:
@bikramnehra we addressed your comments and got @mudler approval. Merging.

Issue #277, opened 2019-12-18
Add mechanism to generate file with all release images info


bikramnehra:
**Is your feature request related to a problem? Please describe.**

Parent Issue: https://github.com/SUSE/kubecf/issues/220

At the time of release, we would require a YAML file to be generated with information about all the release images being used.

**Describe the solution you'd like**

To get a file containing the list of images published that are being used by kubecf.

**Describe alternatives you've considered**

Create a Bazel rule that takes the required files to calculate the images being used by kubecf in the same way that the cf-operator does.

**Additional context**

A use case for this is to copy the images to a production registry.

bikramnehra:
closing as https://github.com/cloudfoundry-incubator/kubecf/pull/557 is DONE.

PR #557, opened 2020-03-25
Print image list script for kubecf manifest

rohitsakala:
This bash script prints the list of images used in the kubecf manifest after interpolation of ops files and implicit variables.

## How Has This Been Tested?
locally.

## Types of changes

## Checklist:

### DIFF
--- a/dev/kubecf/fetchImageList.sh
+++ b/dev/kubecf/fetchImageList.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+
+set -eu
+
+if [ "$#" -ne 2 ]; then
+  echo "Usage: $0 <HELM_TEMPLATE_FILE> <VALUES_FILE>"
+  echo ""
+  echo "Example: $0 output.yaml values.yaml"
+  exit 1
+fi
+
+manifestIdentifyString="# Source: kubecf/templates/cf_deployment.yaml"
+variableIdentifyString="# Source: kubecf/templates/implicit_vars.yaml"
+qjobName="QuarksJob"
+
+# Count the number of yaml blocks
+yamlFileCount=$(grep -c "# Source: kubecf/templates/" "$1")
+yamlFileCount=$((yamlFileCount - 1))
+
+# Fetch manifest file
+for fileIndex in $(seq 0 "$yamlFileCount")
+do
+    yamlOutput=$(yq r -d"$fileIndex" "$1")
+    if [[ $yamlOutput =~ $manifestIdentifyString ]]
+    then
+        yq r -d"$fileIndex" "$1" data.manifest > manifest.yaml
+    fi
+done
+
+## Interpolate ops files
+for fileIndex in $(seq 0 "$yamlFileCount")
+do
+	yamlOutput=$(yq r -d"$fileIndex" "$1" data.ops)
+    if [ -n "$yamlOutput" ]
+    then
+        yq r -d"$fileIndex" "$1" data.ops > ops"$fileIndex".yaml
+        bosh interpolate manifest.yaml --ops-file ops"$fileIndex".yaml > manifest"$fileIndex".yaml
+        cp manifest"$fileIndex".yaml manifest.yaml
+        
+        rm ops"$fileIndex".yaml
+        rm manifest"$fileIndex".yaml
+    fi
+done
+
+## Interpolate implicit variables
+for fileIndex in $(seq 0 "$yamlFileCount")
+do
+	yamlOutput=$(yq r -d"$fileIndex" "$1")
+    if [[ $yamlOutput =~ $variableIdentifyString ]]
+    then
+        key=$(yq r -d"$fileIndex" "$1" metadata.name)
+        key=$(cut -d '-' -f2- <<< "$key")
+        key=${key//-/_}
+        value=$(yq r -d"$fileIndex" "$1" stringData.value)
+        if [ -n "$value" ]
+        then
+            bosh interpolate manifest.yaml -v "$key"="$value" > manifest"$fileIndex".yaml
+            cp manifest"$fileIndex".yaml manifest.yaml
+            
+            rm manifest"$fileIndex".yaml
+        fi
+    fi
+done
+
+# Print image info from kube manifest
+imageCount=$(yq r manifest.yaml --length releases)
+for imageIndex in $(seq 0 "$imageCount")
+do
+    value=$(yq r manifest.yaml releases["$imageIndex"].stemcell)
+    if [ -n "$value" ]
+    then 
+        imageName=$(yq r manifest.yaml releases["$imageIndex"].name)
+        imageUrl=$(yq r manifest.yaml releases["$imageIndex"].url)
+        imageOS=$(yq r manifest.yaml releases["$imageIndex"].stemcell.os)
+        imageStemcellVersion=$(yq r manifest.yaml releases["$imageIndex"].stemcell.version)
+        imageVersion=$(yq r manifest.yaml releases["$imageIndex"].version)
+    
+        imageElement="${imageUrl}/${imageName}:${imageOS}-${imageStemcellVersion}-${imageVersion}"
+        echo "$imageElement"
+    fi
+done
+
+# Print image info from QuarksJob
+for fileIndex in $(seq 0 "$yamlFileCount")
+do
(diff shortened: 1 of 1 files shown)


bikramnehra:
We had a similar backlog item: https://github.com/cloudfoundry-incubator/kubecf/issues/277

f0rmiga:
There are a few binary dependencies here that we cannot assume people will have installed or that will have the right version. We have Bazel to deal with that. Also, we need to handle images that don't come from the BOSH manifest.

mudler:
I agree with @f0rmiga , there are dependencies in this script that are assumed to be installed (`yq` and `bosh` as far as I can tell). But for me a bazel target is not necessary. It's fine also by using a script that uses a docker image containing those deps, you might get away with `splatform/bosh-cli`.

fargozhu:
Christian Richter is looking into the script and we may take a different approach by generating the imagelist based on the values.yaml in the helm chart atm.

I am holding this issue until I get feedback.

viovanov:
> I agree with @f0rmiga , there are dependencies in this script that are assumed to be installed (`yq` and `bosh` as far as I can tell). But for me a bazel target is not necessary. It's fine also by using a script that uses a docker image containing those deps, you might get away with `splatform/bosh-cli`.

@mudler that would severely break reproducible builds. There's another ticket to deal with bazel's shortcomings, but we'll not address them in this issue. This issue should be implemented using the current build system of this project.

viovanov:
> Christian Richter is looking into the script and we may take a different approach by generating the imagelist based on the values.yaml in the helm chart atm.
> 
> I am holding this issue until I get feedback.

@dragonchaser you can't base that on helm output alone, you need bosh to interpolate ops files.

mudler:
> > I agree with @f0rmiga , there are dependencies in this script that are assumed to be installed (`yq` and `bosh` as far as I can tell). But for me a bazel target is not necessary. It's fine also by using a script that uses a docker image containing those deps, you might get away with `splatform/bosh-cli`.
> 
> @mudler that would severely break reproducible builds. There's another ticket to deal with bazel's shortcomings, but we'll not address them in this issue. This issue should be implemented using the current build system of this project.

I don't see how it can break reproducible builds, can you elaborate? `splatform/bosh-cli` is tagged and running it inside a container makes more reproducible, not less. 

I don't think we should block the current work waiting just for a bazel rule, we have already other scripts in the kubecf repo that are not bazel targets.

viovanov:
> > > I agree with @f0rmiga , there are dependencies in this script that are assumed to be installed (`yq` and `bosh` as far as I can tell). But for me a bazel target is not necessary. It's fine also by using a script that uses a docker image containing those deps, you might get away with `splatform/bosh-cli`.
> > 
> > 
> > @mudler that would severely break reproducible builds. There's another ticket to deal with bazel's shortcomings, but we'll not address them in this issue. This issue should be implemented using the current build system of this project.
> 
> I don't see how it can break reproducible builds, can you elaborate? `splatform/bosh-cli` is tagged and running it inside a container makes more reproducible, not less.

Then you'd need a role to freeze the version of that docker image.
And you need a pipeline to build the image. You're left with supporting a new artifact - `splatform/bosh-cli`, for no good reason.

> I don't think we should block the current work waiting just for a bazel rule, we have already other scripts in the kubecf repo that are not bazel targets.

Yes, we should - we should follow the rules of the project, that's what will keep the project maintainable.
We have other scripts that pull in dependencies?

mudler:
> > > > I agree with @f0rmiga , there are dependencies in this script that are assumed to be installed (`yq` and `bosh` as far as I can tell). But for me a bazel target is not necessary. It's fine also by using a script that uses a docker image containing those deps, you might get away with `splatform/bosh-cli`.
> > > 
> > > 
> > > @mudler that would severely break reproducible builds. There's another ticket to deal with bazel's shortcomings, but we'll not address them in this issue. This issue should be implemented using the current build system of this project.
> > 
> > 
> > I don't see how it can break reproducible builds, can you elaborate? `splatform/bosh-cli` is tagged and running it inside a container makes more reproducible, not less.
> 
> Then you'd need a rule to freeze the version of that docker image.
> And you need a pipeline to build the image. You're left with supporting a new artifact - `splatform/bosh-cli`, for no good reason.

The pipeline for `splatform/bosh-cli` already exists, we already have that image, and its already being tagged. That's why I suggested to use that one. https://concourse.suse.de/teams/main/pipelines/bosh-cli . https://github.com/SUSE/cloudfoundry/blob/master/ci/pipelines/bosh-cli/pipeline.yml

> 
> > I don't think we should block the current work waiting just for a bazel rule, we have already other scripts in the kubecf repo that are not bazel targets.
> 
> Yes, we should - we should follow the rules of the project, that's what will keep the project maintainable.
> We have other scripts that pull in dependencies?

Then we should apply this logic always, and not only case-by-case basis.
No we don't have scripts that pull in dependencies, but we have scripts that calls bazel in other ways (and bazel, its still a dependency in my opinion).

viovanov:
> > > > > I agree with @f0rmiga , there are dependencies in this script that are assumed to be installed (`yq` and `bosh` as far as I can tell). But for me a bazel target is not necessary. It's fine also by using a script that uses a docker image containing those deps, you might get away with `splatform/bosh-cli`.
> > > > 
> > > > 
> > > > @mudler that would severely break reproducible builds. There's another ticket to deal with bazel's shortcomings, but we'll not address them in this issue. This issue should be implemented using the current build system of this project.
> > > 
> > > 
> > > I don't see how it can break reproducible builds, can you elaborate? `splatform/bosh-cli` is tagged and running it inside a container makes more reproducible, not less.
> > 
> > 
> > Then you'd need a rule to freeze the version of that docker image.
> > And you need a pipeline to build the image. You're left with supporting a new artifact - `splatform/bosh-cli`, for no good reason.
> 
> The pipeline for `splatform/bosh-cli` already exists, we already have that image, and its already being tagged. That's why I suggested to use that one. https://concourse.suse.de/teams/main/pipelines/bosh-cli . https://github.com/SUSE/cloudfoundry/blob/master/ci/pipelines/bosh-cli/pipeline.yml

The pipeline and the repo are private for that.
And you still need to freeze the tag for this (which seems like more work, than just adding 2 binaries to bazel). 
 
> > > I don't think we should block the current work waiting just for a bazel rule, we have already other scripts in the kubecf repo that are not bazel targets.
> > 
> > 
> > Yes, we should - we should follow the rules of the project, that's what will keep the project maintainable.
> > We have other scripts that pull in dependencies?
> 
> Then we should apply this logic always, and not only case-by-case basis.
> No we don't have scripts that pull in dependencies, but we have scripts that calls bazel in other ways (and bazel, its still a dependency in my opinion).

I agree - we should apply the logic always and sounds like we are.
+1 on scripts that call bazel - nothing wrong with that. I have a local `Makefile` that I use to compose various targets for example, we can add something like that, sure.

f0rmiga:
Of course, there's a starting point that you need to assume about dependencies. This starting point is Bazel. Some of the command line toolings are also assumed to be present given that their API is stable (cut, etc).

Scripts that wrap Bazel for some tasks should be fine, as long as you are not assuming a bunch of other dependencies that don't make sense.

mudler:
I take back what I said, there are bash scripts that are lying around calling other tools ( I just grepped for yq) : https://github.com/cloudfoundry-incubator/kubecf/search?l=Shell&q=yq&type=

Yes, the bosh-cli lives in a private repo, but we already had a jira card for it https://jira.suse.com/browse/CAP-551

f0rmiga:
@mudler This dates from the very beginning of the project and is wrong.

mudler:
> Of course, there's a starting point that you need to assume about dependencies. This starting point is Bazel. Some of the command line toolings are also assumed to be present given that their API is stable (cut, etc).


That's a bit stretched, based on the shell you use, cut/awk behaviour can change too.

But I'm fine with waiting for the proper bazel target

mudler:
> @mudler This dates from the very beginning of the project and is wrong.

what do you mean? We still use that docker image to build bosh final releases  :smirk:

viovanov:
This PR should be updated so we get a full list of all images (on stdout).

f0rmiga:
Closing as superseded by https://github.com/cloudfoundry-incubator/kubecf/pull/571.

Issue #501, opened 2020-03-19
Pool mechanism for KubeCF Pipeline


mudler:
**Describe the bug**
As soon as we stared to use the CI even more, we get a lot of errors about resource limits:
```
[./up.sh] [backend:ekcp] [cluster:kubecf-diego-1584567416-ef143df] Loading

{"Output":"","AvailableClusters":null,"Clusters":null,"ActiveEndpoints":null,"ClusterIPs":null,"LocalClusters":null,"Error":"No available resources"}/tmp/build/b0d51b9f/catapult/buildkubecf-diego-1584567416-ef143df /tmp/build/b0d51b9f/catapult/backend/ekcp
```

Just only for Diego:
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/206
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/199
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/198
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/191
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/190
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/189
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/188
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/187
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego/builds/180
.... 

The KubeCF pipeline, limits only the deployment in parallel. There is no limit between the tests that are running (that will keep the cluster for itself) and the deploy steps. This leads to the situation where new deployment starts, while tests are still runninng and didn't cleaned up the cluster yet.

A pool mechanism would be needed as we are already starting to automate even more bumps of component to KubeCF. As the pipeline runs also for master, this quickly can become a deadlock, and block all the work as we don't have no more published and tested charts in the bucket (that other pipelines are picking up).

**To Reproduce**
Is easily reproducible by merging 3 or more PR in a row, they will fire 6 jobs that exceeds our resource limits as they will require 12 kubernetes cluster.

**Expected behavior**
A pool to limit concurrent jobs

Issue #443, opened 2020-03-05
Break down CI pipeline to smaller chunks


jimmykarily:
As we add steps to [the Concourse pipeline](https://github.com/cloudfoundry-incubator/kubecf/tree/master/.concourse) it slowly becomes a very big and unmanageable yaml file. We even have [bash scripts](https://github.com/cloudfoundry-incubator/kubecf/blob/master/.concourse/pipeline.yaml#L96) defined inside the yaml.

We should extract anything that can be extracted to smaller manageable files, whether is it a script of a task definition (yaml).

viccuad:
I would love to have all bash scripts that call catapult inside of the pipeline on a folder ala:
```
scripts
├── 00-recover-ekcp.sh
├── 10-deploy-kubecf.sh
├── 20-test-smokes.sh
├── 21-test-cats.sh
├── 22-test-brains.sh
├── 23-test-sits.sh
└── 30-upgrade-kubecf.sh
```
and then provide 1 script that loops and calls those in order. Basically a simple implementation of the concourse pipeline, but locally on a shell script.

those scripts/* can just be called on the concourse yaml at the same time.

mudler:
Fair to me, but want to underline that approach has one disadvantage: it then ties the pipeline with a code checkout. If you look close at the steps of the kubecf pipelines are basically oneliners (deploy/test), I would second that approach in case scripts becomes "big chunks"

HartS:
> but want to underline that approach has one disadvantage: it then ties the pipeline with a code checkout

@mudler the pipeline as it exists on concourse will reflect the state of the template from the checkout from which it was deployed, so isn't this true either way?

viccuad:
I also like little shellscripts as they can be linted easily.

mudler:
> > but want to underline that approach has one disadvantage: it then ties the pipeline with a code checkout
> 
> @mudler the pipeline as it exists on concourse will reflect the state of the template from the checkout from which it was deployed, so isn't this true either way?

well, not necessarly `set-pipeline` is a `fly` command, so you can always deploy different pipelines *not* tight with the code

PR #561, opened 2020-03-26
CI: Use pool system for the pipeline

mudler:
## Description
Fixes: https://github.com/cloudfoundry-incubator/kubecf/issues/501 . Introduce a pool mechanism to the KubeCF pipeline, and templates Eirini and Diego job definitions.

Kind clusters are now created by a separate pipeline, kubecf-pool-reconciler. This pipeline reconciles a "max pool size" by creating clusters and adding their kubeconfigs to the pool repository until the number of clusters in circulation match the "max pool size".

The KubeCF pipeline then consumes them, making jobs that have no resources to run, and wait for a lock that eventually will be created by the reconciler pipeline. The KubeCF pipeline will destroy the cluster and remove the kubeconfig from the pool when it finishes successfully, or in any job that has failed.

## Motivation and Context
See: https://github.com/cloudfoundry-incubator/kubecf/issues/501 .

## How Has This Been Tested?
https://concourse.suse.dev/teams/main/pipelines/kubecf-test
https://concourse.suse.dev/teams/main/pipelines/pool-create-test

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/kubecf-pool-reconciler.yaml.gomplate
+++ b/.concourse/kubecf-pool-reconciler.yaml.gomplate
@@ -1,4 +1,4 @@
-{{ $kind_pool_size := 2 }}
+{{ $kind_pool_size := 5 }}
 
 resources:
 - name: kind-environments
@@ -66,4 +66,3 @@ jobs:
         args: *deploy_args
   - put: kind-environments
     params: {add: output}
-
--- a/.concourse/kubecf-pool-reconciler.yaml.gomplate
+++ b/.concourse/kubecf-pool-reconciler.yaml.gomplate
@@ -1,3 +1,5 @@
+{{ $kind_pool_size := 2 }}
+
 resources:
 - name: kind-environments
   type: pool
@@ -16,14 +18,14 @@ resources:
 
 
 deploy_args: &deploy_args
-- -xce
+- -ce
 - |
   export FORCE_DELETE=true
   export BACKEND=ekcp
   export QUIET_OUTPUT=true
   export CLUSTER_NAME="$CLUSTER_PREFIX-$(date +%s | sha256sum | base64 | head -c 32 ; echo)"
   CURRENT_CLUSTERS=$(ls kind-environments/kind/**/* | grep -v .keep | wc -l)
-  if [ $CURRENT_CLUSTERS -lt 2 ]; then
+  if [ $CURRENT_CLUSTERS -lt {{ $kind_pool_size }} ]; then
 
   pushd catapult
   make k8s
--- a/.concourse/kubecf-pool-reconciler.yaml.gomplate
+++ b/.concourse/kubecf-pool-reconciler.yaml.gomplate
@@ -25,10 +25,10 @@ deploy_args: &deploy_args
   export QUIET_OUTPUT=true
   export CLUSTER_NAME="$CLUSTER_PREFIX-$(date +%s | sha256sum | base64 | head -c 32 ; echo)"
   CURRENT_CLUSTERS=$(ls kind-environments/kind/**/* | grep -v .keep | wc -l)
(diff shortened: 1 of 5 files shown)


.concourse/kubecf-pool-reconciler.yaml.gomplate
```diff
@@ -1,3 +1,5 @@
+{{ $kind_pool_size := 2 }}
```
mudler: If changes looks good, we will bump this to the real number of available VMs when we will replace the current pipeline

.concourse/pipeline.yaml.gomplate
```diff
@@ -346,10 +361,15 @@ jobs:
       file: output/kubecf-bundle-v*.tgz
       acl: public-read
 
-- name: deploy-diego
+{{- range $_, $cfScheduler := slice "diego" "eirini" }}
+
+- name: deploy-{{ $cfScheduler }}
   max_in_flight: 2
```
mudler:
```suggestion
  max_in_flight: 1
```

With this change, we don't make the deployment wait for a lock, but they will queue up. The deployment won't happen anymore in parallel, but the test runs will.

.concourse/kubecf-pool-reconciler.yaml.gomplate
```diff
@@ -0,0 +1,69 @@
...
```
mudler:
```suggestion
{{ $kind_pool_size := 5 }}
```

.concourse/kubecf-pool-reconciler.yaml.gomplate
```diff
@@ -0,0 +1,69 @@
...
+  type: git
+  source:
+    uri: https://github.com/SUSE/catapult
+  version:
+    ref: 543bc71abc8bd22e5ef81fdf94a7afeb2f99066d
+
+
+deploy_args: &deploy_args
+- -ce
+- |
```
HartS: We should be moving away from embedding entire bash scripts into the pipeline file; tasks are capable of reading these from external files in a git repo (this one) and the commit ID used at the time the pipeline was last deployed/updated can be used to ensure the script doesn't change unless the pipeline does

.concourse/kubecf-pool-reconciler.yaml.gomplate
```diff
@@ -0,0 +1,69 @@
...
+  export FORCE_DELETE=true
+  export BACKEND=ekcp
+  export QUIET_OUTPUT=true
+  export CLUSTER_NAME="$CLUSTER_PREFIX-$(date +%s | sha256sum | base64 | head -c 32 ; echo)"
+  CURRENT_CLUSTERS=$(ls kind-environments/kind/**/* | grep -v .keep | wc -l)
+  if [ $CURRENT_CLUSTERS -lt {{ $kind_pool_size }} ]; then
```
HartS: Instead of the pipeline making this determination, can EKCP do it? Similar to the API for any other infrastructure provider (AKS/ECP/EKS/etc.), if the deploy job tries to provision a resource when the provider is at its limit, the API request to provision said resource will fail.

.concourse/kubecf-pool-reconciler.yaml.gomplate
mudler: We have a separate story for this: https://github.com/cloudfoundry-incubator/kubecf/issues/443 . It would be out of scope for this PR

.concourse/kubecf-pool-reconciler.yaml.gomplate
mudler: Yes, the EKCP fails if you don't have enough resources. This is a way to just guarantee we don't hit that error and we can always provision with a boundary limit

Issue #258, opened 2019-12-12
Streamline variable names in values.yaml


aduffeck:
**Is your feature request related to a problem? Please describe.**
This is a follow-up card to the discussion at https://github.com/SUSE/kubecf/pull/244#pullrequestreview-330095215.

We currently mix different styles for the key names in `deploy/helm/kubecf/values.yaml`. Most of them are using snake case, some (mostly the ones referring to instance group names) are using kebap case.

Kebap case keys make it impossible to use dot notation to address them in the helm chart, so 

```
{{- if .Values.testing.acceptance_tests.enabled -}}
```
becomes
```
{{- $acceptance_tests := index .Values.testing "acceptance-tests" -}}
{{- if $acceptance_tests.enabled -}}
```

Furthermore, the inconsistent naming makes it hard for users to get it right without looking things up every time.

**Describe the solution you'd like**
All keys that are under our control in values.yaml should be snake case.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #568, opened 2020-03-27
Update to snake case from kebab case in values.yaml for sync-integration-tests

championshuttler:
Fixes part of  #258

### DIFF
--- a/testing/sync_integration_tests/BUILD.bazel
+++ b/testing/sync_integration_tests/BUILD.bazel
@@ -7,7 +7,7 @@ kubectl_patch(
     name = "sync_integration_tests",
     namespace = project.namespace,
     resource_type = "qjob",
-    resource_name = "kubecf-sync-integration-tests",
+    resource_name = "kubecf-sync_integration_tests",
     patch_type = "merge",
     patch_file = ":trigger.yaml",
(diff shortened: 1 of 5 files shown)


testing/sync_integration_tests/BUILD.bazel
```diff
@@ -7,7 +7,7 @@ kubectl_patch(
     name = "sync_integration_tests",
     namespace = project.namespace,
     resource_type = "qjob",
-    resource_name = "kubecf-sync-integration-tests",
+    resource_name = "kubecf-sync_integration_tests",
```
f0rmiga: This is wrong. This is the k8s resource name, and k8s doesn't allow `_`. In fact, cf-operator converts any `_` to `-`.

f0rmiga:
```suggestion
    resource_name = "kubecf-sync-integration-tests",
```

fargozhu: @f0rmiga what about the "-" after the "kubecf" word?

viovanov:
Not something we want to pursue right now.
Any changes to the values keys should come with deprecation.

PR #574, opened 2020-03-31
Extract log-cache jobs from the doppler into a separate instance group

aduffeck:
That way doppler can be scaled and log-cache kept as a singleton.
Scaling log-cache doesn't work as of now.

*Note*: There's still one failing smoke test and pushing apps fails with some stats server error. The app does run fine and logs can be streamed, though.


f0rmiga:
The work I had locally was slightly different than this. Since I had it committed, I rebased to include @aduffeck as a co-author. https://github.com/cloudfoundry-incubator/kubecf/pull/619

Issue #545, opened 2020-03-24
Spike: improve bazel experience in the repo


viovanov:
**Is your feature request related to a problem? Please describe.**
Getting started with bazel is hard.

**Describe the solution you'd like**
- perhaps a makefile to wrap useful bazel targets
- comments in the bazel build files and rules
- FAQ section

mudler:
Very hard to achieve imho. I've already expressed my concerns, but I will add also my thoughts here on the subject.

- I see this issue a bit countersense, coming up with ways to improve its "daily" experience it just highlights that it's not good, and it is a failure (my personal point of view)  my experience with bazel

mudler:
Very hard to achieve imho, I have already expressed my thoughts on the subject already, but I'll add them here too so everyone is free to debate with me in the public :smile:  :

- I see this issue a bit a countersense. Having to come up with ways to "simplify it" shows that bazel failed in this area. No matter how you will hide the complexity, we will have to face it on "daily" basis. And given that our team is not acquainted with it, it becomes even cumbersome. That is my personal opinion with the experience I had with it.
- Creating a Makefile to wrap it, sounds pretty much that we don't need bazel. Having two ways to building things on the same project is a red-flag for me, because we could just have one that works, and is really understood by the involved parties. Bazel looks like a top-down decision that (as far as I know) I'm not the only one that disagrees with.  Usually one have to support their decision to *adopt* new things, not the opposite way around.
- There are comments, even README in the build targets, that doesn't help much. Extending the docs about them, again, shows to me that it failed very hard in usability. I expect also to have docs that are human readable from bazel itself (yes, might sound hard, but I don't find them helpful)
- To be honest: I see working on this again as a wasted engineering cycles. "Wrapping up" will make things even more complex, and even harder to understand.
- I would expect a "novel" build system to give me more than just restrictions, at least should give me something back for the effort I put on it. There is no real catch of using bazel for me (again, personal opinion, but if we don't have opinion on softwares, means we are not using it :smile:  ) except the complexity built around it, and docs that are very difficult to follow. There are no even consistent examples that you can base your work on.
[...]

PR #583, opened 2020-04-01
Build and run kubecf without bazel

jandubois:
This is just a quick proof-of-concept for the discussion in #545 

It is neither complete nor robust, and it is missing the installation/pinning of additional tools like helm and jq, but it is complete enough to build the kubecf helm chart, deploy it to minikube and run smoke tests.


scripts/tools/README.md
```diff
@@ -0,0 +1,85 @@
+# Tools definitions
+
+This directory contains definitions for additional tools used to build the project.
```
mudler, line 3: I really like it!

mudler:
First of all: **incredibly faster** , the build time was practically almost 0. I also really like the tooling management implementation, neat and can scale out.

Just a small suggestion, could we maybe add a check if the cf-deployment submodule is not present and display a message? I think it could also be easily a target for the others to depend on (if not present, do the sync, otherwise leave it untouched)

jandubois:
> could we maybe add a check if the cf-deployment submodule is not present and display a message?

Implemented in 7c416b6

jandubois:
This proof-of-concept now includes a tool dependency manager that can be used to guarantee that specific versions of tools will be used for all building and running tasks. It is activated by setting the `PINNED_TOOLS` environment variable, e.g.

```
PINNED_TOOLS=1 make tools-install
make kubecf-build
```

The `tools-install` target will make sure that *all* tools that have an associated download URL are available from the `output/bin` directory in the requested version, and that directory is at the front of the `PATH` when running any of the scripts.

All scripts will check that all tools they use are at least at the required version, even when `PINNED_TOOLS` is not set. But they will also accept later versions installed elsewhere in the filesystem in that configuration. That way developers can test the build with the latest versions of tools installed locally while the CI system can build with the verified pinned configuration.

This mechanism is slightly more generic than the one implemented in bazel, as it can also verify versions of tools that are not managed by the build system, like `ruby` or `docker`.

jandubois:
Marking this as ready-for-review. There is still lots of room for improvement, but maybe that can come in a followup PR.

jimmykarily:
Something is wrong:

```
c0fe8ee6a329:/kubecf # make tools-install
c0fe8ee6a329:/kubecf # make update-subcharts
./dev/helm/update_subcharts.sh: line 7: helm: command not found
make: *** [Makefile:14: update-subcharts] Error 127
```

`make tools-version` return `helm version is 3.0.3` and I see helm in `output/bin`.

jandubois:
> Something is wrong:

@jimmykarily You are absolutely right! The `Makefile` was calling the generic version of the `update_subcharts` script, which doesn't use the tool setup from the `scripts/` system. Fixed in 789103e (and the commits before it).

jimmykarily:
Something else is wrong now:

```
~/workspace/suse/kubecf (bash-is-beautiful)*$ docker run -it -v $PWD:/workspace -w /workspace opensuse/leap:15.2 bash
d9c8507e11c5:/workspace # zypper in -y git make curl vim python3 ruby gzip tar > /dev/null
d9c8507e11c5:/workspace # make tools-install
Installing python3
Can't find URL for python3-3.3
make: *** [Makefile:5: tools-install] Error 1
d9c8507e11c5:/workspace # python3 --version
Python 3.6.10
```

jimmykarily:
Ignore my above comment. The problem was that I run the make targets both outside docker and from within a container. The permissions of the `${VENV_DIR}/bin/activate` file didn't allow the script to read it inside the container. I removed everything from the `output` directory and now it works. To add some context, I'm running all make targets in docker to make sure I'm not calling any of the binaries installed in my system (e.g. helm, etc).

scripts/kubecf-build.sh
```diff
@@ -0,0 +1,74 @@
...
+if [[ ! "$(git submodule status -- src/cf-deployment)" =~ ^[[:space:]] ]]; then
+    die "git submodule for cf-deployment is uninitialized or not up-to-date"
+fi
+
+HELM_DIR="${TEMP_DIR}/helm"
+
+[ -d "${HELM_DIR}" ] && rm -rf "${HELM_DIR}"
+mkdir "${HELM_DIR}"
+
+cp -a deploy/helm/kubecf/ "${HELM_DIR}"
```
jimmykarily, line 15:
```suggestion
cp -a deploy/helm/kubecf/* "${HELM_DIR}"
```
otherwise `make lint` fails with

```
mkdir: cannot create directory ‘/home/dimitris/workspace/suse/kubecf/output/tmp/helm/assets’: Not a directory
```

`make kubecf-build` also fails

jimmykarily:
I compared the output of `make kubecf-build` to that of `/dev/build.sh` (which is calling `bazel build //deploy/helm/kubecf:version` inside).

They are mostly similar except for:

1. the order of keys inside the `jobs/*.yaml` files (doesn't matter)
2. imagelist.txt exists in the bash artifact but missing from the bazel artifact (which is a good thing bash target does I guess?)
3. Metadata.yaml in bash doesn't interpolate the operator version:

image

jimmykarily:
Regarding feature parity with bazel, `bazel run //dev/kind:start` was using [`LOCAL_PATH_STORAGE_YAML`](https://raw.githubusercontent.com/rancher/local-path-provisioner/58cafaccef6645e135664053545ff94cb4bc4224/deploy/local-path-storage.yaml) defined in [dependencies.yaml](https://github.com/cloudfoundry-incubator/kubecf/blob/master/dependencies.yaml#L113) in order to setup a storage class for kind. I don't see the make targets doing the same. On the other hand, with recent kind versions I never had to manually setup a storageclass so maybe it's not needed at all. Just wanted to make sure it is a delibarate decision to skip it.

jimmykarily:
Feature parity again: there is no make target equivalent to `//dev/external_database:deploy_mysql`. Personally I haven't used that bazel target but maybe it is used in pipelines (?).

rules/kubecf/image_list.rb
```diff
@@ -0,0 +1,285 @@
+# frozen_string_literal: true
```
jimmykarily, line 1: Our plan to deprecate bazel doesn't guarantee that this file won't diverge from this: https://github.com/cloudfoundry-incubator/kubecf/blob/master/rules/kubecf/image_list.tmpl.rb

I know this PR was created as a replacement to bazel, but until bazel is gone we need to make sure both build targets work the same. Still, I don't have a suggestion on how to ensure this. Maybe we should just keep that in mind when reviewing PRs.

scripts/kubecf-build.sh
jandubois: Seems to be implementation defined behaviour, and GNU and BSD implement it differently. :(

Fixed in 0b1b2e3.

jandubois:
> 3\. Metadata.yaml in bash doesn't interpolate the operator version:

Thanks for catching that. I broke it recently in 745618b. Fixed in 92b796f.

jandubois:
> `bazel run //dev/kind:start` was using [`LOCAL_PATH_STORAGE_YAML`](https://raw.githubusercontent.com/rancher/local-path-provisioner/58cafaccef6645e135664053545ff94cb4bc4224/deploy/local-path-storage.yaml)

`kind-start.sh` is doing the same thing, afaict: https://github.com/cloudfoundry-incubator/kubecf/blob/jandubois/bash-is-beautiful/scripts/kind-start.sh#L70-L75

Please let me know how this is different!

rules/kubecf/image_list.rb
jandubois: > Our plan to deprecate bazel doesn't guarantee that this file won't diverge

No, it doesn't, but it seems rather unlikely that we will make any changes to the script within the next 4 weeks or so. I don't think this is important.

The file will need to move to a new location though, and I'm not sure if it should go directly into `scripts/`, or into a subdirectory, as you are not expected to call it directly.

jandubois:
> there is no make target equivalent to `//dev/external_database:deploy_mysql`

I've now copied the shell code to `./scripts/deploy-mysql.sh` and de-bazelfied it, but it was still using helm2 syntax and a mysql chart that was incompatible with kube 1.16+, so I doubt it has been used recently.

I've fixed it enough that it runs without failure, but don't really like the way the script requires interactive input and cannot be used in automation, so I haven't created a corresponding `Makefile` target for it.

I don't want to spend any more time on it now; but it can serve as a starting point when we need this functionality again, so it is good to preserve it.

jandubois:
@jimmykarily I believe I've addressed all your feedback now; let me know if something is not satisfactory to you.

jandubois:
> 2\. imagelist.txt exists in the bash artifact but missing from the bazel artifact (which is a good thing bash target does I guess?)

Forgot to mention: I believe this was a separate step in Bazel because expanding all feature combinations to get the image list took over 5 minutes, which would slow down local development.

Now that we only evaluate all features on and all features off, the result is instantaneous, and `./scripts/kubecf-build.sh` runs in about 5 seconds, including building the image list.

So I see no reason why that should be a separate thing. I want everything as simple as possible...

jandubois:
> 1. the order of keys inside the `jobs/*.yaml` files (doesn't matter)

Forgot to mention: this is because I used `ruby` and `jq` instead of `yq` to extract parts of the deployment manifest. It avoids defining an additional tool dependency, and also `yq` is dog-slow, and I wanted `kubecf-build` to finish in a few seconds. 😄

rules/kubecf/image_list.rb
jimmykarily: Agreed, let's keep it where it is for now.

jimmykarily:
> > `bazel run //dev/kind:start` was using [`LOCAL_PATH_STORAGE_YAML`](https://raw.githubusercontent.com/rancher/local-path-provisioner/58cafaccef6645e135664053545ff94cb4bc4224/deploy/local-path-storage.yaml)
> 
> `kind-start.sh` is doing the same thing, afaict: https://github.com/cloudfoundry-incubator/kubecf/blob/jandubois/bash-is-beautiful/scripts/kind-start.sh#L70-L75
> 
> Please let me know how this is different!

Nope all good. I somehow missed that.

scripts/README.md
```diff
@@ -0,0 +1,174 @@
...
+### Development
+
+#### ./scripts/tools-install.sh
+
+This script installs the pinned version of all configured tools into `TOOLS_DIR`
+unless the installed version is already an exact match.
+
+#### ./scripts/tools-versions.sh
+
+List all configured tools and their current versions.
```
jimmykarily, line 61: Maybe worth documenting how this behaves when PINNED_TOOLS is set. E.g. If I have pinned tools and I have a version of a tool installed locally (but not the exact required version), what will this script return, my installed version or missing or other?

scripts/README.md
```diff
@@ -0,0 +1,174 @@
...
+Builds the `kubecf` helm chart. This includes the `imagelist.txt` inventory of
+all referenced container images and the `sample-values.yaml` version of the
+commented out `values.yaml`.
+
+The default output filename will be `output/kubecf-${VERSION}.tgz`, with `$VERSION`
+determined by `./scripts/version.sh`. An alternate output filename can be set (e.g.
+for the benefit of CI) via:
+
+```bash
+TARGET_FILE=out/kubecf.tgz make kubecf-build
```
jimmykarily, line 103: awesome!

scripts/include/tools.sh
```diff
@@ -0,0 +1,248 @@
...
+
+    blue "Installing ${tool}"
+
+    # Look for custom install command first (e.g. for Python module install via pip).
+    if [ -n "$(type -t "${tool}_install")" ]; then
+        eval "${tool}_install"
+        return
+    fi
+
+    # XXX (require_tools is not reentrant) require_tools file gzip sha256sum xz
```
jimmykarily, line 179: What does this mean?

scripts/tools/python3.sh
```diff
@@ -0,0 +1,26 @@
+# shellcheck shell=bash
+# shellcheck disable=SC2034
+
+TOOLS+=(python3)
+
+PYTHON3_VERSION=3.3
+
+function python3_version {
+    # Only return a version when virtualenv is setup and activated.
```
jimmykarily, line 9: But it's us who activate the vitrualenv (below) no? Which case are we trying to avoid here?

jimmykarily:
Only some minor comments (see above) but it looks great! Let's get this merged.

scripts/include/tools.sh
jandubois: Good catch. I had already fixed this, but forgot to re-enable the code here. Fixed in 8512193.

scripts/tools/python3.sh
jandubois: I can no longer remember; I suspect it is no longer serving any valid purpose. I've removed it in ce33a56. We can always add it back later if we find some breakage. But in my simple tests using `yamllint` it didn't seem to be useful.

dev/linters/shellcheck.sh
```diff
@@ -6,9 +6,29 @@ workspace=$(bazel info workspace)
...
+             -- read - binaries.shellcheck.version)
+
+# Make sure shellcheck is downloaded into the Bazel cache.
+bazel run @shellcheck//:binary -- --version &> /dev/null
+
+# Call cached bazel binary of shellcheck from current directory manually.
+# shellcheck disable=SC2046
+# We want word splitting with find.
+"./bazel-kubecf/external/shellcheck/shellcheck-v${VERSION}/shellcheck" \
+    -- $(find "${workspace}" "${find_args[@]}")
```
jimmykarily, line 34: This fails in the pipeline: https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/lint-pr/builds/17

```./dev/linters/shellcheck.sh: line 34: ./bazel-kubecf/external/shellcheck/shellcheck-v0.7.0/shellcheck: No such file or directory
```

I will try to debug locally.

jandubois: Fixed in e9adca5

Issue #241, opened 2019-12-09
Doppler out of memory when scaled


jkbschmid:
**Describe the bug**
When scaling to two Doppler instances and deploying an application, the memory consumption increases until either the k8s node crashes or Doppler is killed by the OOMKiller.

**To Reproduce**
Install kubecf master with cf-operator master.
Run `cf push` with the example Dora app.

**Expected behavior**
Doppler should consume reasonable amount of memory.

**Environment**
 - CF Version: v8 and v12

loewenstein:
@viovanov blocker for 1.0?

f0rmiga:
The problem here seems to be with log-cache. In my initial investigation, I couldn't find the exact problem. For the 0.2 release, I'm going to restrict HA for doppler. We should get back to it for 1.0 though.

f0rmiga:
I changed the priority from Critical to High given that there is a temporary fix in place already.

loewenstein:
I wouldn't call "don't scale it" a temporary fix for "this can't be scaled".
What does it mean for logs, if Doppler is a singleton? Potential delay or loss of logs? If it is the latter I would still judge it critical I guess. Unless there are other issues left that are comparable to completely failing k8s nodes, of course.

f0rmiga:
@loewenstein Do you have spare cycles to help to debug this?

bikramnehra:
Seems like the original [solution](https://github.com/SUSE/kubecf/pull/366) that was proposed is not going to work out. An [issue](https://github.com/cloudfoundry/log-cache-release/issues/29) has been filed upstream though it might take some time to be fixed as the root cause hasn't been identified yet.

Since the problematic part is the `log-cache` job therefore one potential fix as proposed by @f0rmiga is to leave the `log-cache` job outside of the `doppler` instance group and avoid scaling it altogether.

f0rmiga:
As discussed with @viovanov, we will keep it as is for now.

viovanov:
Upstream's VM deployments don't seem to be suffering from this issue based on discussions.

PR #619, opened 2020-04-03
log-cache instance group

f0rmiga:
## Description

Fixes https://github.com/cloudfoundry-incubator/kubecf/issues/241.

## Motivation and Context

Moving the log-cache jobs out of the doppler instance group allows us to scale up doppler while keeping log-cache as a single instance. log-cache scalability has been noticed by other teams, in special the cf-cli team. Their issue was not the same as KubeCF, but rather related to out-of-sync replication between log-cache instances.

This PR should fix both issues for now while log-cache finds a better solution for this problem.

## How Has This Been Tested?

Ran smoke-tests and cf-acceptance-tests on both SA and HA, checking that no tests failed and no memory spike happened.

## Types of changes

## Checklist:


deploy/helm/kubecf/job_moving.bzl
```diff
@@ -0,0 +1,39 @@
...
+# The default list of jobs related to log-cache.
+LOG_CACHE_JOBS = [
+    "log-cache",
+    "log-cache-gateway",
+    "log-cache-nozzle",
+    "log-cache-cf-auth-proxy",
+    "route_registrar",
+]
+
+def extract_log_cache(instance_group, log_cache_jobs):
```
mook-as, line 17: Can we not use the existing `yaml_extractor` for this? Split it into 5 jobs files, and the re-combine them in `move_log_cache.yaml`?

I rather strongly prefer having fewer `*.bzl` files, as every time we drop into one it feels like we're working against bazel.

deploy/helm/kubecf/templates/bosh_deployment.yaml
```diff
@@ -16,10 +16,10 @@ spec:
     name: cf-deployment
     type: configmap
   ops:
-  - name: {{ include "kubecf.ops-name" (dict "Path" "assets/operations/unordered/move_auctioneer.yaml") }}
-    type: configmap
-  - name: {{ include "kubecf.ops-name" (dict "Path" "assets/operations/unordered/move_routing_api.yaml") }}
+{{- range $path, $bytes := .Files.Glob "assets/operations/job_moving/*" }}
```
mook-as, line 19: We never use `$bytes` here.

deploy/helm/kubecf/templates/ops.yaml
```diff
@@ -20,8 +20,9 @@ data:
 
 {{- $root := . }}
 
-{{ include "kubecf.ops" (dict "Root" $root "Path" "assets/operations/unordered/move_auctioneer.yaml") }}
-{{ include "kubecf.ops" (dict "Root" $root "Path" "assets/operations/unordered/move_routing_api.yaml") }}
+{{- range $path, $bytes := .Files.Glob "assets/operations/job_moving/*" }}
+{{ include "kubecf.ops" (dict "Root" $root "Path" $path) }}
```
mook-as, line 24: We never use `$bytes` here either.

deploy/helm/kubecf/assets/operations/instance_groups/doppler.yaml
```diff
@@ -1,113 +1,17 @@
...
-    - name: doppler-tls
-      protocol: TCP
-      internal: 3459
     - name: doppler-ws
       protocol: TCP
       internal: 8081
     - name: doppler-grpc
       protocol: TCP
       internal: 8082
-    - name: log-cache-proxy # log-cache-cf-auth-proxy
```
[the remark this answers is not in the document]
f0rmiga: Hmmm no... maybe SCF selected the port by name?

deploy/helm/kubecf/assets/operations/instance_groups/log-cache.yaml
```diff
@@ -0,0 +1,132 @@
...
+  path: /instance_groups/name=log-cache/jobs/name=route_registrar/properties/quarks?
+  value:
+    run:
+      healthcheck:
+        route_registrar:
+          # The route registrar doesn't expose anything to indicate if the
+          # routes are healthy.
+          readiness: ~
+
+- type: replace
```
f0rmiga, line 119:
```suggestion
# Add log-cache.service.cf.internal DNS alias to be able to point CC to it,
- type: replace
```

deploy/helm/kubecf/job_moving.bzl
f0rmiga: Could you expand on `feels like we're working against bazel`?

deploy/helm/kubecf/assets/operations/job_moving/move_log_cache.yaml
```diff
@@ -0,0 +1,22 @@
...
+  value:
+    name: log-cache
+    instances: 1
+    stemcell: default
+    update:
+      serial: false
+    jobs:
+    {{- range $path, $bytes := $.Files.Glob "assets/jobs/log_cache_*" }}
+    - {{- $.Files.Get $path | nindent 6 }}
+    {{- end }}
```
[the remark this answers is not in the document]
f0rmiga: We never used `$bytes` because it returns `[]uint8`, not string. We can make it work with `toString`, though. I'll prepare a separate PR for fixing all the instances where we could do it, which goes way beyond this current PR.

deploy/helm/kubecf/job_moving.bzl
mook-as: I see bazel as having two languages (that are actually the same starlark): the one in `*.bazel` where we define rules, and the one in `*.bzl` where we define implementation.  The two types of files feel very different and _feel_ like completely different things with a large mental context switch between them.  Sort of like dropping into inline assembly from within a C program.

As such, every time I need to dig into an implementation file I feel like I stopped writing starlark-the-rules-language and are dipping into starlark-the-build-system-implementation-language, and it feels like I'm working _around_ the existing bazel rules to do something it doesn't want me to do.

deploy/helm/kubecf/assets/operations/instance_groups/doppler.yaml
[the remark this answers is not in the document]
jandubois: Port names were defined by rfc6335 and are limited to 15 characters. I don't know if this has changed; it is hard to find documentation about it, but I *do* know that things used to fail with longer port names.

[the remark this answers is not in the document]
f0rmiga: It doesn't seem to fail here, so I'll leave this as is. Thanks.

deploy/helm/kubecf/assets/operations/instance_groups/log-cache.yaml
```diff
@@ -0,0 +1,132 @@
...
+  path: /instance_groups/name=log-cache/jobs/name=log-cache/properties/health_addr?
+  value: "::6060"
+
+# Add quarks properties for log-cache.
+- type: replace
+  path: /instance_groups/name=log-cache/jobs/name=log-cache/properties/quarks?
+  value:
+    ports:
+    - name: log-cache-metrics
```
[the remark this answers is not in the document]
f0rmiga: We used to expose on SCF, so I'll leave as is and create an issue to investigate this.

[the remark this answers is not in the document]
f0rmiga: https://github.com/cloudfoundry-incubator/kubecf/issues/635 created.

deploy/helm/kubecf/assets/operations/instance_groups/log-cache.yaml
```diff
@@ -0,0 +1,132 @@
...
+              - https://localhost:6060/debug/pprof/cmdline
+
+# Add quarks properties for log-cache-gateway.
+- type: replace
+  path: /instance_groups/name=log-cache/jobs/name=log-cache-gateway/properties/quarks?
+  value:
+    ports:
+    - name: log-cache-gateway-metrics
+      protocol: TCP
+      internal: 6063
```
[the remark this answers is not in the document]
f0rmiga: Should be tackled here as well: https://github.com/cloudfoundry-incubator/kubecf/issues/635.

deploy/helm/kubecf/job_moving.bzl
f0rmiga: I can see that. These python functions on `.bzl` files are known as macros in Bazel. They are sometimes useful for creating a bunch of targets. I was able to do the same job with a list comprehension on the `.build` file. It should be better to read now.

f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/kubecf/pull/619/commits/52ebe7224e48656c3986a2d5d109f9083a56d106.

deploy/helm/kubecf/templates/bosh_deployment.yaml
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/kubecf/pull/619/commits/49c5c179b823834d7b592b1d83725f079dc3312f.

deploy/helm/kubecf/templates/ops.yaml
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/kubecf/pull/619/commits/49c5c179b823834d7b592b1d83725f079dc3312f.

PR #643, opened 2020-04-09
Revert "Revert "Update cf-operator dependency.""

viovanov:
Reverts cloudfoundry-incubator/kubecf#641


viovanov:
This operator version has a regression, fixed by 
https://github.com/cloudfoundry-incubator/cf-operator/pull/886

PR #651, opened 2020-04-09
Consume eirini release

jimmykarily:
## Description
This PR combines (and in doing so, also supersedes) the changes from these two:

https://github.com/cloudfoundry-incubator/kubecf/pull/477
https://github.com/cloudfoundry-incubator/kubecf/pull/465

## Motivation and Context

The 2 PRs above are both needed in order to use the Eirini release chart instead of the eirini bosh release. It makes sense to have all the changes in one branch which we can test and merge.

## How Has This Been Tested?

Will deploy on kind and test (PR is still a draft, will update the description when testing is done)

## Types of changes

## Checklist:


jimmykarily:
Fails because the bits tls certificate files are don't match what the template expects. Draft PR to fix it: https://github.com/cloudfoundry-incubator/bits-service-release/pull/39/files

deploy/helm/kubecf/templates/ops-remove-diego.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if .Values.features.eirini.enabled }}
```
jimmykarily, line 1: This seem to have moved here: https://github.com/cloudfoundry-incubator/kubecf/blob/master/deploy/helm/kubecf/assets/operations/instance_groups/eirini.yaml#L332

jimmykarily: Sorry, my mistake. Not "have moved" but "also exist". The condition here is only for eirini to be enabled. The condition on the eirini.yaml file is for eirini to be enabled and use_helm_release to be false. That means, when we don't use the helm release for eirini, boths ops will be applied. I think we should better change the eirini.yaml file so that it works for both cases.

deploy/helm/kubecf/requirements.lock
```diff
@@ -0,0 +1,9 @@
+dependencies:
+- name: eirini
+  repository: https://cloudfoundry-incubator.github.io/eirini-release
+  version: 1.4.0
+- name: bits
+  repository: https://cloudfoundry-incubator.github.io/bits-service-release/helm
+  version: 2.36.0
```
jimmykarily, line 7: This should be `2.37.0` like it is below.

Issue #575, opened 2020-03-31
Consume latest eirini-loggregator-bridge


mudler:
**Is your feature request related to a problem? Please describe.**
The eirini-loggregator-bridge fix is merged in the eirini-bosh-release https://github.com/SUSE/eirini-loggregator-bridge/issues/4 , we should at least bump to the latest to avoid pod crash loops https://github.com/cloudfoundry-community/eirini-bosh-release/pull/74

**Describe the solution you'd like**
Bump eirini-bosh-release in the meantime that we cannot consume the Eirini native chart

**Describe alternatives you've considered**
Not bumping it is not an option, as otherwise we stuck with https://github.com/SUSE/eirini-loggregator-bridge/issues/4

f0rmiga:
I guess we need to bump the whole eirini-bosh-release, right?

mudler:
Yup! but if we do https://github.com/cloudfoundry-incubator/kubecf/issues/578 first, we don't need to use the bosh release necessarily. We could just avoid to consume the jobs from the bosh release and create a separate pod for the extensions (ssh, persi, persi-broker and loggregator-bridge)

fargozhu:
I would prefer to wait and avoid the bosh release...

mudler:
Closing, as it was fixed when https://github.com/cloudfoundry-incubator/kubecf/pull/681 was merged, but not valid anymore since we don't consume anymore the eirini bosh release.

Issue #576, opened 2020-03-31
 Eirini-loggregator-bridge is missing logs


mudler:
**Describe the bug**
https://github.com/SUSE/eirini-loggregator-bridge/issues/6

**To Reproduce**
Deploy Kubecf with latest eirini-loggregator-bridge code

**Expected behavior**
Logs shouldn't be lost, but at most be delayed because with the EiriniX bump, the watcher implementation is bookmarked.

gaktive:
Looking for an update @mudler when you get a chance :)

mudler:
I think we can close this, I've bumped the dependencies in EiriniX here: https://github.com/SUSE/eirinix/pull/31 which includes a bunch of fixes also regarding watchers. The fix isn't in KubeCF yet as the bosh-release was removed as part of the Eirini native chart integration, so we have to wait for https://github.com/cloudfoundry-incubator/kubecf/pull/771 to be in

PR #681, opened 2020-04-15
Bump eirini-bosh release to 0.0.27

mudler:
## Description
Bumps to latest bosh release to consume the loggregator-bridge including fixes for https://github.com/SUSE/eirini-loggregator-bridge/issues/4 

## Motivation and Context
Couldn't reproduce https://github.com/cloudfoundry-incubator/kubecf/issues/576 anymore, see https://github.com/SUSE/eirini-loggregator-bridge/issues/6 see https://github.com/cloudfoundry-incubator/kubecf/issues/575  for the bump card.

## How Has This Been Tested?
Locally with kind

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/values.yaml
+++ b/deploy/helm/kubecf/values.yaml
@@ -60,6 +60,9 @@ releases:
       tag: 0.9.4
   eirini:
     version: 0.0.27
+    stemcell:
+      os: SLE_15_SP1
+      version: 23.21-7.0.0_374.gb8e8e6af
   postgres:
     version: "39"
   sle15:
--- a/deploy/helm/kubecf/values.yaml
+++ b/deploy/helm/kubecf/values.yaml
@@ -59,7 +59,7 @@ releases:
       repository: docker.io/cfcontainerization/pxc
       tag: 0.9.4
   eirini:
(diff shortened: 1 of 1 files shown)


f0rmiga:
So, we should close #575 and #576 after we merge this?

mudler:
This fixes #575.  

I couldn't reproduce #576 anymore but I wouldn't close it as it seems better to consume the latest changes in kubernetes/client-go regarding `RetryWatchers` (see https://github.com/SUSE/eirini-loggregator-bridge/issues/6#issuecomment-614068411 ) It looks like retrywatchers being consumed by EiriniX right now are not taking in consideration when there are network timeouts because it doesn't include that commit - which could likely happen in an environment with network issues (DNS delays and such). 

It requires a bump in kubernetes/client-go in EiriniX, i'm trying to see if this is feasible , but that requires a controller-runtime bump. Since we depend on cf-operator code, that is not possible at the moment, as cf-operator is behind that commit too. 

My theory is that probably I was connected inside VPN when doing testing, and that actually makes sometimes my dns resolving flaky locally, as the resolv.conf is inherited to the kubernetes cluster in kind, hence the issue.

PR #694, opened 2020-04-15
fix: simplify the output of external_binary rule

f0rmiga:
## Description

The structure of the files exported by the repository_rule is simpler, allowing for easier dependency consumption from other repository rules.

## Motivation and Context

Repository rules dislike symlinks for binary dependencies that we want to consume. E.g. passing the Helm binary to a repository rule requires the use of the real path, rather than the `:binary` target that returns a symlink to the real path.

With this change, the Helm binary can also be consumed via `@helm//:helm`, which happens to be the real path.

The `:binary` symlink is still useful for regular rules and should not be impacted by these changes.

## How Has This Been Tested?

Ran the `--help` for each of the binary dependencies we currently have using the `:binary` targets, as well as tried out the explicit binary location for helm and kubectl, which asserts that a `.tgz` and a binary src work well with the rule.

## Types of changes

## Checklist:

### DIFF
--- a/def.bzl
+++ b/def.bzl
@@ -8,128 +8,136 @@ project = struct(
 
     # External binaries; see external_binary() invocation in WORKSPACE.
     external_binaries = {
-        "bosh_cli": struct(
-            sha256 = {
+        "bosh_cli": {
+            "sha256": {
                 "darwin": "1d2ced5edc7a9406616df7ad9c0d4e3ade10d66d33e753885ab8e245c037e280",
                 "linux": "ca7580008abfd4942dcb1dd6218bde04d35f727717a7d08a2bc9f7d346bce0f6",
                 "windows": "77c736c15001b1eb320ae61042fb6c72a1addde143e0a9af703ddda35b2c5bce",
             },
-            url = {
+            "url": {
                 "darwin": "https://github.com/cloudfoundry/bosh-cli/releases/download/v{version}/bosh-cli-{version}-darwin-amd64",
                 "linux": "https://github.com/cloudfoundry/bosh-cli/releases/download/v{version}/bosh-cli-{version}-linux-amd64",
                 "windows": "https://github.com/cloudfoundry/bosh-cli/releases/download/v{version}/bosh-cli-{version}-windows-amd64.exe",
             },
-            version = "6.2.1",
-        ),
-        "docker": struct(
-            sha256 = {
-                "linux":   "50cdf38749642ec43d6ac50f4a3f1f7f6ac688e8d8b4e1c5b7be06e1a82f06e9",
+            "version": "6.2.1",
+        },
+        "docker": {
+            "sha256": {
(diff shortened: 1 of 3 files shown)


def.bzl
```diff
@@ -8,128 +8,136 @@ project = struct(
 
     # External binaries; see external_binary() invocation in WORKSPACE.
     external_binaries = {
-        "bosh_cli": struct(
-            sha256 = {
+        "bosh_cli": {
```
mudler, line 11: I guess this is still related to #545 . That's more a post-merge comment: since we (were) here, couldn't we have a yaml file that `bazel` is reading instead? Is Bazel implying `struct` in this case?

rules/external_binaries/def.bzl
```diff
@@ -12,32 +12,52 @@ def _external_binary_impl(ctx):
...
         "sha256": ctx.attr.sha256[os],
     }
+
+    if ctx.attr.strip_prefix.get(os, "") != "":
+        args["stripPrefix"] = ctx.attr.strip_prefix[os]
+
     if any([url.endswith(suffix) for suffix in [".zip", ".tar.gz", ".tgz", ".tar.bz2", ".tar.xz"]]):
-        ctx.download_and_extract(output="{name}/{name}_out".format(name = ctx.attr.name), **args)
+        ctx.download_and_extract(output = ".", **args)
+        build_contents = """
```
mudler, line 21: I'm a bit confused.. is this a generated file? wouldn't make it sense to carry out in in the codebase instead to be in stubbed in the function?

rules/external_binaries/def.bzl
```diff
@@ -56,15 +76,17 @@ _external_binary = repository_rule(
             doc = "Version of the binary",
             mandatory = False,
         ),
+        "strip_prefix": attr.string_dict(
+            allow_empty = True,
+            doc = "Directory prefixex to strip from the extracted files",
```
mudler, line 81: nitpick: small typo ~prefixex~ prefixes

def.bzl
f0rmiga: Done with https://github.com/cloudfoundry-incubator/kubecf/pull/712. :)

rules/external_binaries/def.bzl
f0rmiga: A Bazel repository rule generates another Bazel repository that we can load with `@<repository_name>`. So, for the downloaded files, we generate a `BUILD.bazel` that creates a symlink and exports those real files.

f0rmiga: When we use `@helm//:binary`, it's using the symlink. When we use `@helm//:helm` it's using the real path.

For using the real path, it's less predictable, this is why we create the `:binary` target, for easy discoverability of the binary from the downloaded files. It's useful when consuming `tgz` files that we don't know the exact directory structure.

rules/external_binaries/def.bzl
f0rmiga: Oh yeah, I caught that as well when I moved it to its own repo, so it's fixed there already.

PR #697, opened 2020-04-16
Remove tiller check from dev/kube/kube-ready-state-check.sh

viccuad:
## Description

Checking for Tiller is not needed anymore, we require helm >= 3.

## How Has This Been Tested?

Run `kube-ready-state-check.sh` by hand.

## Types of changes

## Checklist:


mudler:
I think in general we should advertize what `def.bzl` pins (https://github.com/cloudfoundry-incubator/kubecf/blob/master/def.bzl#L47 , 3.0.3 in such case ) but Catapult tests with the version you updated in the README, so we should be good :+1:

viccuad:
Thanks; that's better, updated.

PR #707, opened 2020-04-16
feat: helm dependencies rules

f0rmiga:
## Description

The dependencies repository rule is suited for fetching and caching Helm dependencies declared in the requirements.yaml file. Those dependencies are also known as sub-charts and are fetched as .tgz files under the charts/ directory. The rule exports a filegroup that selects all the charts to be used with the package rule.

## Motivation and Context

This PR is to support the Eirini native Helm charts PR: https://github.com/cloudfoundry-incubator/kubecf/pull/651.

We want to include the Helm sub-charts in the final package, but still keep the fetching as part of a repository rule so it is aggressively cached to avoid network interaction during the build phase.

## How Has This Been Tested?

Added a few sub-charts to KubeCF and used the repository rule implemented in this PR:

```
# WORKSPACE

load("//rules/helm:def.bzl", helm_dependencies = "dependencies")

helm_dependencies(
    name = "kubecf_helm_dependencies",
    chart_yaml = "//deploy/helm/kubecf:Chart.yaml",
    requirements = "//deploy/helm/kubecf:requirements.yaml",
    requirements_lock = "//deploy/helm/kubecf:requirements.lock",
)
```

Then added the `subcharts` property to the `kubecf` target:
```
# //deploy/helm/kubecf/BUILD.bazel

# Packages the KubeCF Helm chart.
helm_package(
    name = "kubecf",
    srcs = [
        ":chart_files_static",
    ],
    generated = [
        ":metadata",
    ],
    tars = [
        "//bosh/releases:pre_render_scripts",
        ":cf_deployment",
        ":extracted_jobs",
    ],
    subcharts = ["@kubecf_helm_dependencies//charts"],
)
```

## Types of changes

## Checklist:

### DIFF
--- a/rules/helm/def.bzl
+++ b/rules/helm/def.bzl
@@ -16,6 +16,7 @@ def _package_impl(ctx):
             "[[tars]]": multipath_sep.join([f.path for f in ctx.files.tars]),
             "[[generated]]": multipath_sep.join([f.path for f in ctx.files.generated]),
             "[[version]]": ctx.attr.version,
+            "[[subcharts]]": multipath_sep.join([f.path for f in ctx.files.subcharts]),
             "[[helm]]": ctx.executable._helm.path,
             "[[output_tgz]]": output_tgz.path,
         },
@@ -23,7 +24,11 @@ def _package_impl(ctx):
     )
     ctx.actions.run_shell(
         command = "ruby {}".format(package_script.path),
-        inputs = [package_script] + ctx.files.srcs + ctx.files.tars + ctx.files.generated,
+        inputs = [package_script] +
+            ctx.files.srcs +
+            ctx.files.tars +
+            ctx.files.generated +
+            ctx.files.subcharts,
         outputs = outputs,
         tools = [ctx.executable._helm],
     )
@@ -44,6 +49,7 @@ _package = rule(
             mandatory = False,
             default = "",
         ),
+        "subcharts": attr.label_list(),
         "_helm": attr.label(
             allow_single_file = True,
             cfg = "host",
--- a/rules/helm/def.bzl
+++ b/rules/helm/def.bzl
@@ -2,6 +2,13 @@
 The definitions and implementations of the Bazel rules for dealing with Helm.
 """
 
+_helm_attr = attr.label(
+    allow_single_file = True,
+    cfg = "host",
+    default = "@helm//:helm",
+    executable = True,
+)
+
 def _package_impl(ctx):
     output_filename = "{}.tgz".format(ctx.attr.name)
     output_tgz = ctx.actions.declare_file(output_filename)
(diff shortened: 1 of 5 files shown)


rules/helm/def.bzl
```diff
@@ -63,6 +71,67 @@ def package(**kwargs):
         **kwargs
     )
 
+def _dependencies_impl(ctx):
```
mudler, line 74: This is a bit hard to follow, is it generating a BUILD.bazel in place? I cannot see where actually it is consumed, could you please explain me how this works?

rules/helm/def.bzl
```diff
@@ -63,6 +71,67 @@ def package(**kwargs):
...
+
+    # Fetch the dependencies.
+    ctx.execute([helm, "dep", "up"])
+
+    # Create the workspace root BUILD.bazel.
+    ctx.file("BUILD.bazel", 'package(default_visibility = ["//visibility:public"])\n')
+
+    # Create the charts/BUILD.bazel exporting the fetched charts.
+    charts_build = """
+    package(default_visibility = ["//visibility:public"])
```
mudler, line 94: If this ends up in the generated `BUILD.bazel`, I would prefer a template or something that is imported by this rule, at least is more clear and you carry the logic out in the code separately

f0rmiga: Hmm, so, I went to take a better look at the docs for the `repository_ctx`, and it turns out we can do what you asked: https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#read.

This file ends up being `@<workspace>//charts:BUILD.bazel`. So, if we use this rule like in the example I gave in the description:

```
helm_dependencies(
    name = "kubecf_helm_dependencies",
    chart_yaml = "//deploy/helm/kubecf:Chart.yaml",
    requirements = "//deploy/helm/kubecf:requirements.yaml",
    requirements_lock = "//deploy/helm/kubecf:requirements.lock",
)
```

Then, this file is `@kubecf_helm_dependencies//charts:BUILD.bazel`. It will export the `.tgz` charts downloaded with the `helm dep up` command.

rules/helm/def.bzl
f0rmiga: This rule is not consumed in their very same PR. I will be first used in the Eirini PR that is introducing some sub-charts.

f0rmiga: The workspace created with the repository_rule can then be used in the kubecf package with `subcharts = ["@kubecf_helm_dependencies//charts"],`.

rules/helm/def.bzl
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/kubecf/pull/707/commits/b733db8d722e8d910de3934ec3de54cd8ed107ca.

PR #708, opened 2020-04-17
feat: move external_binaries to its own home

f0rmiga:
## Description

Moving the `external_binaries` rules to its own home at https://github.com/SUSE/rules_binaries.

## Motivation and Context

We want to move generic Bazel rules to their own repositories to simplify the KubeCF repository and to enable other Bazel projects to reuse these rules.

## How Has This Been Tested?

Locally with the usual development flow using external binary dependencies.

## Types of changes

## Checklist:


mudler:
I guess this is somehow related to #545, but a naive question (as I am not a Bazel expert): couldn't we have at least "one" repo with the utils that we have for bazel? Or we will have multiple repositories for each functionality? 

I fear that in this way, we will split way too much, and have to handle a release for each change in those modules (if we end up with a lot of multiple repos)

f0rmiga:
@mudler We will rename the `rules_binaries` to something more obvious and that can take more utility rules! Thanks for the input.

viovanov:
I renamed the repos to `bazel_rules_helm` and `bazel_rules_utils`.
The helm ones should still live in their own repo. Things like binary downloading and other generic stuff can live in the utils one.

f0rmiga:
This PR needs to be worked on now that the repository was renamed.

viovanov:
With #1011 this looks like it's no longer necessary.

Issue #476, opened 2020-03-11
Use the SUSE eirinifs


f0rmiga:
**Is your feature request related to a problem? Please describe.**

In order to have SSH with Eirini, we need to use https://github.com/SUSE/eirinifs-sle15-release.

**Describe the solution you'd like**

Consume the `eirinifs-sle15-release` and set the appropriate properties to make it work. `cf ssh` should work with Eirini apps.

**Additional context**

This is a follow-up from https://github.com/cloudfoundry-incubator/kubecf/pull/464.

f0rmiga:
The commit on SCF for reference on adding it: https://github.com/SUSE/scf/commit/d9edcc41a48d6a2a500ab5791df062bbb47d83b7.

viovanov:
This needs to work with the helm chart.
We shouldn't need the rootfs patcher.

fargozhu:
https://github.com/cloudfoundry-incubator/bits-service-release/blob/master/docker/eirini-rootfs-downloader/eirini-rootfs-downloader.sh#L62

jimmykarily:
This is how upstream is downloading the eirinifs: https://github.com/cloudfoundry-incubator/bits-service-release/blob/master/docker/eirini-rootfs-downloader/eirini-rootfs-downloader.sh#L62 (just for reference)

jimmykarily:
PR to make the eirinifs downloading optional: https://github.com/cloudfoundry-incubator/bits-service-release/pull/41/files

We used a bits pod which had the eirinifs vendored in and it works fine.

viovanov:
Current state:

- [ ] opi
- [x] route_collector
- [x] route_pod_informer_image
- [x] route_statefulset_informer_image
- [x] staging_reporter_image
- [x] recipe-downloader
- [x] recipe-executor
- [x] recipe-uploader
- [x] event_reporter_image
- [x] event-reporter
- [x] metrics_collector_image
- [ ] eirinifs
- [ ] bits

We still need to build the the opi image, and inject the eirinifs in the bits image

PR #771, opened 2020-04-24
feat: support eirinix as external helm chart

mook-as:
This makes eirinix work using an external helm chart instead of as part of eirini-bosh-release.

The various eirini-extensions pre-render scripts are removed as those BOSH jobs no longer exist, and therefore patching them will fail.

We also need to configure CC and UAA to allow access as their service names (`api` and `uaa`), so that the ssh-proxy pod can talk to it without having access to BOSH DNS name resolution.

Closes #771

## Motivation and Context
Put eirini extensions back in.

## How Has This Been Tested?
* Deployed locally with https://github.com/SUSE/eirinix-helm-release/pull/3
* Logs (staging & running) works
* SSH does _not_ work
* Persi not tested

## Types of changes

## Checklist:

## Additional information:
- We will need a follow-up (or same PR?) for making it a subchart.


mudler:
I think we also need to disable fluentd from the Eirini native chart when we enable Eirini extension chart otherwise you probably we see the logs streamed by fluentd. Might be handy a feature flag for the extensions, that disables also fluentd, wdyt?

mook-as:
Rebased, and fluentd disabled.

Current status of the pieces:
- ssh: blocked by #476 
- persi: blocked by https://github.com/SUSE/eirini-persi/issues/24
- logging: works, except staging logs are broken.

I think that this is ready for review; I'll make a follow-up to actually use it as a subchart.

Note that for minikube, we need to override `persi-broker.service-plans.(id == default).kube_storage_class` to `standard`.

viovanov:
@mook-as how are staging logs broken? can you please open an issue?

Issue #739, opened 2020-04-22
DB seeder tries to seed the routing-api database even when routing-api is disabled


f0rmiga:
**Describe the bug**

```
database-seeder-4d3bb95ad30f499b-28qlw seeder /bin/bash: line 38: /passwords/routing-api/password: No such file or directory
```

**To Reproduce**

Disable `routing-api` by setting `features.routing_api.enabled=false`.

**Expected behavior**

The database seeder should not try to create the `routing-api` database.

f0rmiga:
Will be fixed with https://github.com/cloudfoundry-incubator/kubecf/pull/775 by the commit https://github.com/cloudfoundry-incubator/kubecf/pull/775/commits/90081488fceea44c0d6280072a979c37d4ed0f1e.

Issue #549, opened 2020-03-25
database should use utf8 charset instead of latin1


jandubois:
The PXC database has been configured to use the default `latin1` charset.

I'm not sure if this has practical implication, but I think it could affect e.g. using non-latin1 user names in UAA. I have not verified this.

I've found this issue by comparing `mysqldump` output for the UAA database between an SCF and a kubecf instance, to investigate potential migration issues, e.g.

```diff
CREATE TABLE `SPRING_SESSION` (
-  `PRIMARY_ID` char(36) COLLATE utf8_unicode_ci NOT NULL,
+  `PRIMARY_ID` char(36) NOT NULL,
[...]
-) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci ROW_FORMAT=DYNAMIC;
+) ENGINE=InnoDB DEFAULT CHARSET=latin1 ROW_FORMAT=DYNAMIC;
```

jandubois:
There are a whole bunch of pxc config settings beyond the character set that pxc-release sets explicitly, but we currently leave at default values:

https://github.com/cloudfoundry-incubator/pxc-release/blob/master/jobs/pxc-mysql/templates/my.cnf.erb

I've created a diff between the pxc-release values (exported from SCF) and the current kubecf settings and uploaded it to https://gist.github.com/jandubois/6f3faf17ed95adefb321c108f08da33a

We need to determine which of those settings we want to update as well, and which we consider inconsequential.

jandubois:
Assigned to @fargozhu to consider adding it to the icebox/backlog. (I thought this assignment was  already done by a bot, or maybe this issue is too old?).

PR #775, opened 2020-04-27
fix: database charset and collate

f0rmiga:
## Description

Fixes https://github.com/cloudfoundry-incubator/kubecf/issues/549.
Fixes https://github.com/cloudfoundry-incubator/kubecf/issues/739.

Use `utf8` instead of `latin1` for the charset and `utf8_unicode_ci` instead of `latin1_swedish_ci` for the collate.

## Motivation and Context

Sync with what pxc-release uses.

## How Has This Been Tested?

By checking that `show variables like 'char%';` returns the expected values:

```
mysql> show variables like 'char%';
+--------------------------+----------------------+
| Variable_name            | Value                |
+--------------------------+----------------------+
| character_set_client     | utf8                 |
| character_set_connection | utf8                 |
| character_set_database   | utf8                 |
| character_set_filesystem | binary               |
| character_set_results    | utf8                 |
| character_set_server     | utf8                 |
| character_set_system     | utf8                 |
| character_sets_dir       | /usr/share/charsets/ |
+--------------------------+----------------------+
```

#### Migrating from KubeCF 2.1

1. Deployed KubeCF 2.1 which uses `latin1`.
2. Deployed the 12factor app.
3. Upgraded cf-operator.
4. Upgraded KubeCF to the version on this PR.
5. Ran the charset migration job:
```
kubectl patch qjobs database-migrate-charset --namespace kubecf --type merge --patch '{ "spec": { "trigger": { "strategy": "now" } } }'
```
6. Asserted that the databases are using `utf8`:
```
SELECT SCHEMA_NAME 'database', default_character_set_name 'charset', DEFAULT_COLLATION_NAME 'collation' FROM information_schema.SCHEMATA;
```
7. Ran all the test suites.

## Types of changes

## Checklist:


f0rmiga:
Ping @fargozhu to make sure we don't miss the instructions on the release on how to migrate to the new charset.

jandubois:
> OCD, but I have to mention it: you disable/re-enable foreign key checks around each `ALTER` statement. Can't it be disabled at the beginning, and then re-enabled at the very end?

Not just OCD, the MySQL manual also states:

>When the foreign_key_checks system variable is enabled, which is the default setting, character set conversion is not permitted on tables that include a character string column used in a foreign key constraint. The workaround is to disable foreign_key_checks before performing the character set conversion. You must perform the conversion on both tables involved in the foreign key constraint before re-enabling foreign_key_checks.

viccuad:
Started looking into this, summarizing to see if I understood correctly.

We don't want to change the upstream types for the columns.
So in `migrate_charset.sh` we either dump the DB, modify, and commit, or we edit each column of type `varchar`, `*text`, `char` of each table of each database. I will go with the second one, taking care of not re-enabling `foreign_key_checks` until all tables of the database are processed, as:

> If you re-enable foreign_key_checks after converting only one of the tables, an ON DELETE CASCADE or ON UPDATE CASCADE operation could corrupt data in the referencing table due to implicit conversion that occurs during these operations

deploy/helm/kubecf/templates/database.yaml
```diff
@@ -387,4 +409,67 @@ spec:
...
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/name: {{ include "kubecf.fullname" . }}
+    app.kubernetes.io/version: {{ default .Chart.Version .Chart.AppVersion | quote }}
+    helm.sh/chart: {{ include "kubecf.chart" . }}
+spec:
+  trigger:
+    strategy: manual
```
viovanov, line 426: So you run this manually based on docs?

deploy/helm/kubecf/assets/scripts/jobs/pxc/migrate_charset.sh
```diff
@@ -5,41 +5,100 @@
...
+EOF
+
 echo "Waiting for database to be ready..."
-until echo "SELECT 'Ready!'" | mysql --host="${DATABASE_HOST}" --user=root --password="${DATABASE_ROOT_PASSWORD}" --connect-timeout=3 2> /dev/null; do
+until echo "SELECT 'Ready!'" | mysql --connect-timeout=3 1> /dev/null 2> /dev/null; do
   sleep 1
 done
 
-function alter_database() {
+function show_tables() {
```
viccuad, line 20: Wouldn't it be possible to use SQL queries to obtain the info from the tables instead of show tables and awk?
Something like:
```
USE information_schema;

SELECT concat("ALTER TABLE `",table_schema,"`.`",table_name, "` CHANGE `",column_name,"` `",column_name,"` ",data_type,"(",character_maximum_length,")
                  CHARACTER SET utf8 COLLATE utf8_unicode_ci",IF(is_nullable="YES"," NULL"," NOT NULL"),";") AS _sqlquery
FROM `COLUMNS` WHERE table_schema LIKE "${database}" AND data_type IN ('varchar','char');

SELECT concat("ALTER TABLE `",table_schema,"`.`",table_name, "` CHANGE `",column_name,"` `",column_name,"` ",data_type," CHARACTER SET utf8 COLLATE utf8_unicode_ci",IF(is_nullable="YES"," NULL"," NOT NULL"),";") as _sqlquery
FROM `COLUMNS` WHERE table_schema LIKE "${database}" AND data_type IN ('text','tinytext','mediumtext','longtext');
```

deploy/helm/kubecf/templates/database.yaml
f0rmiga: Yep!

f0rmiga: I plan on putting this directly into the release notes.

deploy/helm/kubecf/assets/scripts/jobs/pxc/migrate_charset.sh
f0rmiga: I'd love to take this suggestion now, but this would involve a lot of testing again. `awk` was my friend to experiment a lot and compare the dumps from pre-migration and post-migration.

PR #798, opened 2020-04-29
fix: check for stemcell semver

bikramnehra:
## Description

This PR adds functionality to only update stemcell version if its higher than existing. 

## Motivation and Context

We recently saw issues in `fissile-stemcell-sle` versions not appearing in order in : https://hub.docker.com/repository/docker/splatform/fissile-stemcell-sle

## How Has This Been Tested?

Test pipelines deployed at concourse.suse.dev.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/suse-buildpacks-ci/buildpack-version-bump/tasks/create_pr.sh
+++ b/.concourse/suse-buildpacks-ci/buildpack-version-bump/tasks/create_pr.sh
@@ -41,21 +41,26 @@ yaml.preserve_quotes = True
 yaml.representer.add_representer(type(None), represent_none)
 
 # Breaking down the BUILT_IMAGE to retrieve individual values.
-BUILT_IMAGE_LIST = "${BUILT_IMAGE}".split("/", 2)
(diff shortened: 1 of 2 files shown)


f0rmiga:
Any chance we could use https://pypi.org/project/semver/ instead?

bikramnehra:
> Any chance we could use https://pypi.org/project/semver/ instead?

I don't think the comparison we are doing is following proper `MAJOR.MINOR.PATCH` versioning which is required by [semver](https://pypi.org/project/semver/).

```
>>> semver.VersionInfo.isvalid("23.20")
False
```

Though we can override it but it might be an overkill.

jandubois:
How about using `Gem::Version` then?

```
$ ruby -e "exit Gem::Version.new('20.33') < Gem::Version.new('20.222')"; echo $?
0
$ ruby -e "exit Gem::Version.new('20.33') < Gem::Version.new('20.2')"; echo $?
1
```

I also generally prefer to keep all scripts to either shell or ruby and not add more and more languages to the mix.

bikramnehra:
I was able to use `semver` with a little trick.

PR #814, opened 2020-05-06
feat: Implement credentials and variables settings in values.yaml

jandubois:
Credentials can be specified either with dot notation, or as a map:

```
credentials:
  cf_admin_password: changeme
  credhub_tls.ca: credhub-ca
  credhub_tls.certificate: the-cert
  credhub_tls.private_key: the-key
  credhub_tls.is_ca: false
  credhub_ca:
    certificate: cert
    is_ca: true
    private_key: key
```

Variables are a simple list of key/value pairs:

```
variables:
  system_domain: example.com
```

Note that setting the `system_domain` variable doesn't actually work because it is treated specially by the templates and has to be defined as `.Values.system_domain` instead.

See cloudfoundry-incubator/quarks-operator#895 for more details.

### DIFF
--- a/deploy/helm/kubecf/values.yaml
+++ b/deploy/helm/kubecf/values.yaml
@@ -25,6 +25,10 @@ system_domain: ~
 #             default_size: "1Gi"
 properties: {}
 
+credentials: {}
+
+variables: {}
+
 kube:
   # The storage class to be used for the instance groups that need it (e.g. bits, database and
   # singleton-blobstore). If it's not set, the default storage class will be used.
--- a/deploy/helm/kubecf/templates/bosh_deployment.yaml
+++ b/deploy/helm/kubecf/templates/bosh_deployment.yaml
@@ -15,6 +15,16 @@ spec:
   manifest:
     name: cf-deployment
     type: configmap
+  vars:
+{{- $creds := dict }}
+{{- range $name, $value := .Values.credentials }}
+  {{- $parts := splitList "." $name }}
+  {{- $_ := set $creds (first $parts) nil }}
+{{- end }}
+{{- range $name, $value := $creds }}
+  - name: {{ $name | quote }}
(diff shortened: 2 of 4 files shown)


jandubois:
@viovanov 

I decided to use `credentials` as the top level for explicit variables, and `variables` for the implicit ones, similar to your suggestion on Rocketchat.

Secrets are restricted to `map[string]string`, so I simplified your prototype to only support a single dot in the credential names, to separate the variable name from the field name.

I've extended the logic to allow either dotted notation, or a proper YAML map, and will loop over all keys to convert boolean or numeric values to proper strings.

The sample values specified above will render something like this:

```
---
# Source: cf/templates/secret.yaml
name: "cf_admin_password"
stringData: {"password":"changeme"}
---
# Source: cf/templates/secret.yaml
name: "credhub_ca"
stringData: {"certificate":"cert","is_ca":"true","private_key":"key"}
---
# Source: cf/templates/secret.yaml
name: "credhub_tls"
stringData: {"ca":"credhub-ca","certificate":"the-cert","is_ca":"false","private_key":"the-key"}
```

All the secrets seem to be created correctly, but I did not observe that pods are restarted when the settings are changed. I either misunderstand how the mechanism is supposed to work, or there is an issue with the operator.

There should probably be some comments in the implementation, and I would (maybe) also like to implement some additional logic so that you could set `system_domain` via the `variables` section, but not sure, if it is worth the effort.

deploy/helm/kubecf/values.yaml
```diff
@@ -25,6 +25,10 @@ system_domain: ~
 #             default_size: "1Gi"
 properties: {}
 
+credentials: {}
```
viovanov, line 28: Your description in the PR would fit well here I think.

viovanov:
@jandubois since we can have both options, how does it work when using the cli?
`--set credentials.foo.ca` - what does that do?

PR #818, opened 2020-05-06
Eirini and EiriniX integration

viovanov:


deploy/helm/kubecf/requirements.lock
```diff
@@ -7,6 +7,6 @@ dependencies:
   version: 1.0.14
 - name: eirini-extensions
   repository: https://opensource.suse.com/eirinix-helm-release/
-  version: '>= 0-0'
-digest: sha256:ada05d269e97901296f3e67922b59b02fa4a7ba732c74f7187a5de466e436213
-generated: "2020-05-06T11:53:34.20461864+03:00"
+  version: 0.0.0-9+g6b37279
```
f0rmiga, line 10: Thanks for pinning to a specific version.

deploy/helm/kubecf/values.yaml
```diff
@@ -483,11 +483,12 @@ eirini:
       enable: false
 
 bits:
+  download_eirinifs: false
   global:
     labels: {}
     annotations: {}
     images:
-      bits_service: eirini/bits-service:2.36.0@sha256:4cf84e13890890f5d8443a5e6e129b701d524f51d35c9c4295a0562ed8bb1bb2
+      bits_service: registry.suse.com/cap-staging/bits-service:bits-1.0.15-15.1.6.2.220-24.2
```
mook-as, line 491: We probably want a non-staging image eventually right? (But then that's probably more "production" than the old value anyway)

andreas-kupries: The `download_eirinifs` flag controls the `bits` service init container `download-eirini-rootfs`. False disables that container (exclusion from template result).

deploy/helm/kubecf/values.yaml
```diff
@@ -515,3 +516,13 @@ bits:
...
   tls_key_name: private_key
   tls_ca_name: ca
+
+eirinix:  
+  persi-broker:
+    service-plans:
+    - id: default
+      name: "default"
+      description: "Existing default storage class"
+      kube_storage_class: "standard"
```
mook-as, line 526: This does not belong here: `standard` is for minikube, but this will get deployed with the chart we ship to customers.

(The proper fix, I think, is to drop it in the upstream (also our) chart, and use the default storage class.)

viovanov: @mook-as I will change it back to "default"

viovanov: @mook-as I also agree with the "proper" fix - it **should** use the `default` storage class.
Can you please add an issue for that?

PR #853, opened 2020-05-11
fix: Use correct encryption key labels

jandubois:
To import existing data from an SCF export, it must be possible to specify the encryption key label exactly, e.g.

```yaml
ccdb:
  encryption:
    rotation:
      key_labels:
      - NEW_KEY
      current_key_label: NEW_KEY

credentials:
  ccdb_key_label_new_key: abcdef
```

This means the label must be used verbatim, without the `ccdb_key_label_` prefix, and it must support uppercase characters.

Unfortunately this means that users upgrading kubecf from prior releases will have to manually add the hard-coded prefixes in their config file.

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -96,7 +96,7 @@
   value: 3600
 
 {{- range $keyLabel := .Values.ccdb.encryption.rotation.key_labels }}
-{{- if not (regexMatch "^[a-z]+[a-z0-9_]*[a-z0-9]+$" $keyLabel) }}
+{{- if not (regexMatch "^[a-z]+[a-z0-9_]*[a-z0-9]+$" (lower $keyLabel)) }}
 {{ fail "One or more items in ccdb.encryption.rotation.key_labels don't match the regex /^[a-z]+[a-z0-9_]*[a-z0-9]+$/" }}
 {{- end }}
 {{- if gt (len $keyLabel) 240 }}
@@ -105,7 +105,7 @@
 - type: replace
   path: /variables/-
   value:
-    name: "ccdb_key_label_{{ $keyLabel }}"
+    name: "ccdb_key_label_{{ $keyLabel | lower }}"
     type: password
 {{- end }}
 - type: replace
@@ -113,9 +113,9 @@
   value: &encryption_info
     keys:
       {{- range $keyLabel := .Values.ccdb.encryption.rotation.key_labels }}
-      ccdb_key_label_{{ $keyLabel }}: "((ccdb_key_label_{{ $keyLabel }}))"
+      "{{ $keyLabel }}": "((ccdb_key_label_{{ $keyLabel | lower }}))"
       {{- end }}
-    current_key_label: "ccdb_key_label_{{ .Values.ccdb.encryption.rotation.current_key_label }}"
+    current_key_label: "{{ .Values.ccdb.encryption.rotation.current_key_label }}"
 - type: replace
   path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/properties/cc/database_encryption?
   value: *encryption_info
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -113,9 +113,9 @@
   value: &encryption_info
     keys:
       {{- range $keyLabel := .Values.ccdb.encryption.rotation.key_labels }}
-      {{ $keyLabel }}: "((ccdb_key_label_{{ $keyLabel | lower }}))"
+      "{{ $keyLabel }}": "((ccdb_key_label_{{ $keyLabel | lower }}))"
       {{- end }}
-    current_key_label: {{ .Values.ccdb.encryption.rotation.current_key_label }}
+    current_key_label: "{{ .Values.ccdb.encryption.rotation.current_key_label }}"
 - type: replace
   path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/properties/cc/database_encryption?
   value: *encryption_info
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -96,8 +96,8 @@
   value: 3600
 
 {{- range $keyLabel := .Values.ccdb.encryption.rotation.key_labels }}
-{{- if not (regexMatch "^[a-z]+[a-z0-9_]*[a-z0-9]+$" (lower $keyLabel)) }}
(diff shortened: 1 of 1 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -96,7 +96,7 @@
   value: 3600
 
 {{- range $keyLabel := .Values.ccdb.encryption.rotation.key_labels }}
-{{- if not (regexMatch "^[a-z]+[a-z0-9_]*[a-z0-9]+$" $keyLabel) }}
+{{- if not (regexMatch "^[a-z]+[a-z0-9_]*[a-z0-9]+$" (lower $keyLabel)) }}
 {{ fail "One or more items in ccdb.encryption.rotation.key_labels don't match the regex /^[a-z]+[a-z0-9_]*[a-z0-9]+$/" }}
```
jandubois, line 100: Maybe the regexp should be extended to include uppercase letters (instead of using `lower` in the test), so the error message stays correct.

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -105,17 +105,17 @@
...
+    name: "ccdb_key_label_{{ $keyLabel | lower }}"
     type: password
 {{- end }}
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/database_encryption?
   value: &encryption_info
     keys:
       {{- range $keyLabel := .Values.ccdb.encryption.rotation.key_labels }}
-      ccdb_key_label_{{ $keyLabel }}: "((ccdb_key_label_{{ $keyLabel }}))"
+      "{{ $keyLabel }}": "((ccdb_key_label_{{ $keyLabel | lower }}))"
```
f0rmiga, line 116: You don't need to quote anything before the `:`.
```suggestion
      {{ $keyLabel }}: "((ccdb_key_label_{{ $keyLabel | lower }}))"
```

f0rmiga: Ping @mook-as. This was my first suggestion when I first opened the PR to add these keys. Mark explicitly asked to add the prefix. Mark, any concerns about doing this now?

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
f0rmiga: Yes, please. Just add the upper-case to the regex.

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
jandubois: > You don't need to quote anything before the `:`

@f0rmiga I know, but then you don't need to quote the right hand side either. I suspect all the quotes are unnecessary, but if we quote `$keylabel` on the right side, then we should quote it on the left for consistency as well.

mook-as: I asked for the prefix in https://github.com/cloudfoundry-incubator/kubecf/pull/298#discussion_r365026734 because it was just a top level variable: when looking at an arbitrary kubecf deployment, seeing a secret named `var-foo` would be rather confusing (at least, more so than `var-ccdb-key-label-foo`, IMHO).  But that only applied to the _variable name_; it doesn't apply to the hash key here. So this change should be fine.

f0rmiga: @jandubois Not really, on the right side, the quote is intended to make the data a string. On the left side it's always a string.

jandubois: @f0rmiga You will be surprised by this then:

```
$ ruby -ryaml -e 'puts YAML.load("not: true")'
{"not"=>true}

$ ruby -ryaml -e 'puts YAML.load("true: true")'
{true=>true}

$ ruby -ryaml -e 'puts YAML.load("\"true\": true")'
{"true"=>true}
```

So without the quotes on the left side, the value is not guaranteed to be a string either...

jandubois: Or this:

```
$ ruby -ryaml -e 'puts YAML.load("0x7a: true")'
{122=>true}
```

jandubois: Tempted to put all the quotes back in now 😄

f0rmiga: image

jandubois:
> Hmmm, I'm thinking that this may break KubeCF upgrades.

@f0rmiga Yes, it will break upgrades; thanks for catching this!

Please review dd07908 as a potential fix (which still needs manual action from the user during the upgrade).

jandubois:
From d9c643d:

>The release notes for the next kubecf release must mention that the `ccdb.encryption.rotation` labels must be manually prefixed for the upgrade to succeed. This is the case even if the default settings from values.yaml are being used.
>
>Furthermore, you cannot use both a prefixed and unprefixed key label at the same time, like `ccdb_key_label_my_key` and `my_key` because they would both store the value in the same secret. So don't do that!

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -96,26 +96,29 @@
...
 {{- if gt (len $keyLabel) 240 }}
 {{ fail "One or more items in ccdb.encryption.rotation.key_labels exceed the maximum length of 240 characters" }}
 {{- end }}
 - type: replace
   path: /variables/-
   value:
-    name: "ccdb_key_label_{{ $keyLabel }}"
+    {{- /* Don't double-prefix labels that already have the prefix because they are
+           imported from an earlier version of kubecf.  This allows upgrades to the
+           new (verbatim) label naming scheme while retaining access to old key values. */ }}
```
mook-as, line 110: I believe this is failing `helm lint`: `unexpected comment ends before closing delimiter in input`
It expects either `{{- /*` / `*/ -}}` or `{{/*` / `*/}}` (spaces are significant here).

jandubois: So `helm lint` requires you to violate the [best practices](https://helm.sh/docs/chart_best_practices/templates/#formatting-templates) for formatting templates from its own documentation:

>Template directives should have whitespace after the opening braces and before the closing braces

So yeah, will change to `{{- /* ... */}}` even though I think this is wrong.

jandubois:
> The release notes for the next kubecf release must mention that the `ccdb.encryption.rotation` labels must be manually prefixed for the upgrade to succeed. This is the case even if the default settings from values.yaml are being used.

I've tested the upgrade and confirmed that to make the upgrade from `v2.2.0` work I had to add these settings:

```
ccdb:
  encryption:
    rotation:
      key_labels:
      - ccdb_key_label_encryption_key_0
      current_key_label: ccdb_key_label_encryption_key_0
```

I did however run an upgrade first without the correct encryption keys, and it left my running apps in a `started` state, but without actual running instances. This was not fixed automatically by upgrading again with the fixed encryption key; I had to explicitly `cf restart` the app to get the instance back to `running` state.

PR #857, opened 2020-05-12
Bundle sample values

### LINKED ISSUE #711: Bundles should contain a sample values.yaml with commented defaults
**Is your feature request related to a problem?**

Users should be able to use a values.yaml without actually overriding anything.
It's tempting to grab values.yaml from the chart and start editing, but using the same file in a future release wouldn't work well.
A user could grab a values.yaml from the bundle that has all values commented out, except for the mandatory ones.

- each bundle should contain a sample-values.yaml with the optional keys commented out.
- the yaml file contains comments for each key, that contain:
  - a description of the key
  - a markup that indicates whether the key is mandatory, optional, or invisible
- the file is linted so that all keys have a comment like this

Documentation, but that doesn't help anyone.

**Additional context**

mook-as:
## Description
This adds a `sample-values.yaml` in the chart bundle, where most values are commented out.  Only values marked as `REQUIRED` will be kept uncommented.  Values that are marked as `HIDDEN` will not be present.

## Motivation and Context
Fixes #711 (Mostly; I have not implemented the check for exhaustive commenting, as that requires us to clean up `values.yaml` first.)

## How Has This Been Tested?
Generated a bundle locally, and examined the output.
The terrible ruby script also includes tests (run it with no arguments).

## Types of changes

As this creates a new file only, it is not breaking.

## Checklist:


deploy/helm/kubecf/values.yaml
```diff
@@ -10,19 +11,6 @@ system_domain: ~
 #            logs:
 #              addr: kubecf-log-api:8082
 #
-#  Eirini Persistence Broker setup example:
```
mook-as, line 13: I'm removing this chunk because we're not using BOSH releases for this anymore, and we _do_ have defaults at the bottom of the file.

rules/kubecf/create_sample_values.rb
```diff
@@ -0,0 +1,635 @@
...
+    def has_children?
+        !(children.nil? || children.empty?)
+    end
+
+    def comment
+        [head_comment, line_comment].compact.reject(&:empty?).join("\n")
+    end
+
+    # These two functions are used to assert preconditions:
+    # state(sym) flags a node has having had an operation done.
```
andreas-kupries, line 76: -`has having had`
+`as having had` 
?

rules/kubecf/create_sample_values.rb
```diff
@@ -0,0 +1,635 @@
...
+        node.state :depths
+    end
+    nodes.each do |node|
+        node.state? :parent_nodes
+        next unless node.parent_node.nil?
+        set_node_depth node, 1
+    end
+end
+
+# Determine the node structrually before and after each node.
```
andreas-kupries, line 164: -> `structurally`

rules/kubecf/create_sample_values.rb
mook-as: Fixed in f1e098498638e42df8e9e2999d17f98289a728c3.

rules/kubecf/create_sample_values.rb
mook-as: Fixed in 5063ac6e.

deploy/helm/kubecf/values.yaml
viovanov: 👍

viovanov:
> @mook-as very cool
> how do you plan to run the tests?
> i don't see the `REQUIRED` keyword used - shouldn't `system_domain` be `REQUIRED`?

Just noticed it is `REQUIRED` :)

andreas-kupries:
How to run tests, see PR description: `The terrible ruby script also includes tests (run it with no arguments).`

andreas-kupries:
Note: @viovanov 's requested change is to add the new script's tests to the bazel or other things where CI will run them automatically.

PR #859, opened 2020-05-12
Refactor pipeline to use gke

jimmykarily:
## Description

We want to try to use gke for ephemeral test clusters instead of EKCP. We hope that this will lead to more stable CI, faster tests and hopefully not cost too much.

## How Has This Been Tested?

Not yet tested. It's a draft PR. Will deploy the pipeline and test it out.

## Types of changes

## Checklist:

- [x] My code has security implications. (need to make sure we don't expose any public cloud secrets in the concourse output)

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -79,13 +79,16 @@ resource_types:
     tag: release
 
 resources:
-- name: kind-environments
-  type: pool
+- name: semver.gke-cluster
+  type: semver
   source:
-    uri: git@github.com:SUSE/cf-ci-pools.git
-    branch: kubecf-kind-pools
-    pool: kind
-    private_key: ((github-private-key))
+    driver: s3
+    bucket: {{ if has . "s3_bucket" }}{{ .s3_bucket }}{{ else }}kubecf-ci{{ end }}
+    key: gke-cluster
+    initial_version: 0.0.1
+    access_key_id: ((aws-access-key))
+    secret_access_key: ((aws-secret-key))
+    region_name: {{ if has . "s3_bucket_region" }}{{ .s3_bucket_region }}{{ else }}"eu-central-1"{{ end }}
 
 - name: kubecf-github-release
   type: github-release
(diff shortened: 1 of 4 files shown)


.concourse/pipeline.yaml.gomplate
```diff
@@ -179,51 +182,109 @@ resources:
...
   # https://github.com/SUSE/catapult/wiki/Build-and-run-SCF#build-and-run-kubecf
   make kubeconfig scf
 
+
+  # TODO: Write the name of the cluster on an S3 text file so it can be shared
+  # with the next jobs (to be able to connect to it)
+
 test_args: &test_args
-- -xce
+- -ce
```
jimmykarily, line 242: This should have never been enabled. We tend to copy and paste this everywhere which could easily lead in leaked credentials. Not sure if there is a way to prevent this once and for all. Maybe this can help? https://concourse-ci.org/creds-redacting.html (although I'm not sure if it's gonna work for a big json file). The last sentence says it all :D `Think of this as an airbag, not a seatbelt!`

.concourse/tasks/upgrade.sh
```diff
@@ -15,10 +15,40 @@ export FORCE_DELETE=true
...
+  --enable-ip-alias \
+  --network "projects/${GKE_PROJECT}/global/networks/default" \
+  --subnetwork "projects/${GKE_PROJECT}/regions/$(echo ${GKE_ZONE} | sed 's/-.$//')/subnetworks/default" \
+  --default-max-pods-per-node "110" \
+  --no-enable-master-authorized-networks \
+  --addons HorizontalPodAutoscaling,HttpLoadBalancing \
+  --no-enable-autorepair
+
+# Get a kubeconfig
+gcloud container clusters get-credentials "${GKE_CLUSTER_NAME}" --zone "${GKE_CLUSTER_ZONE}" --project "${GKE_PROJECT}"
```
jimmykarily, line 51: Catapult is doing some strange things here: https://github.com/SUSE/catapult/blob/master/backend/gke/kubeconfig.sh#L6

it requires a kubeconfig but then [gets one itself](https://github.com/SUSE/catapult/blob/master/backend/gke/kubeconfig.sh#L35). We fetch the kubeconfig here too to make catapult happy but this needs some clarification on how catapult is supposed to work with gke.

viccuad: This seems to come from https://github.com/SUSE/catapult/pull/183 , which is a fix as it seems that `gcloud` creates configs with expiry, but I actually think that doesn't hold true anymore. @prabalsharma , are we sure about that bit?

.concourse/pipeline.yaml.gomplate
viccuad: It's a bit troubling that it's everywhere in this pipeline. Could we please `s/-xce/ce`?

jimmykarily: agreed , done

mook-as:
Hi! This PR broke CI because it doesn't pass lint.  Because it's been merged to `master`, none of the other PRs can pass now either.

Please avoid merging PRs before passing CI, thanks!

PR #885, opened 2020-05-18
Remove not needed "privileged" tasks from pipelines

jimmykarily:
because there seems to be a performance penatly with overlay
baggageclaim driver in concourse and privileged containers.

https://github.com/concourse/concourse/issues/4332

With `privileged` on, it took 20 minutes to even start the `deploy` task. Without it it takes 10-20 seconds: https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-diego-master/builds/2

### DIFF
--- a/.concourse/kubecf-pool-reconciler.yaml.gomplate
+++ b/.concourse/kubecf-pool-reconciler.yaml.gomplate
@@ -45,7 +45,6 @@ jobs:
     trigger: true
   - get: catapult
   - task: deploy
-    privileged: true
     timeout: 30m
     config:
       platform: linux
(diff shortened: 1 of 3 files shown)


.concourse/kubecf-pool-reconciler.yaml.gomplate
```diff
@@ -45,7 +45,6 @@ jobs:
     trigger: true
   - get: catapult
   - task: deploy
-    privileged: true
```
jimmykarily, line 48: SInce I'm at it I thought about removing it from here too. Will need to redeploy this pipeline too.

PR #896, opened 2020-05-19
fix: Cherry pick packaging fix

jandubois:
This fix was applied after the `v2.2.0` tag, but used to build the `v2.2.0` release, but the `release-2.2` branch was based of the original tag.


gaktive:
@mook-as as in me? :)

jandubois:
> I assume we'd want an actual release manager to look at it too :D

I don't see why. This commit was used to build the 2.2.0 release, and is required to get the release branch building again. If we don't merge it, all other changes will be blocked.

PR #901, opened 2020-05-20
Don't skip on_failure hooks when one of them fails

jimmykarily:
E.g. when cleanup failed, we didn't update the GH status

## How Has This Been Tested?
Deployed the pipeline. It's an edge case so a bit difficult to test it.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -485,69 +485,75 @@ jobs:
         state: success
 {{- end }}
   on_failure:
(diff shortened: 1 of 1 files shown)


jimmykarily:
I deployed the pipeline with this change. If we decide we like it, I will add the same `try` on all hooks.

viccuad:
I haven't had time for a review, but I wonder if we want to commit this on the long run. I feel using a pool would make things clearer.

PR #908, opened 2020-05-21
fix: deploy: rename sample values file

mook-as:
## Description
Make it clear that the sample values are for kubecf.

## Motivation and Context
Vlad mentioned in chat:
> should they be in the kubecf tarball? they are called "sample-values", but don't apply to the operator for example

## How Has This Been Tested?
Built locally, and inspected the output `tgz`.

```
$ tar tf bazel-bin/deploy/bundle/kubecf-bundle.tgz
./
./cf-operator.tgz
./kubecf_release.tgz
./kubecf-sample-values.yaml
```

## Types of changes

## Checklist:

### DIFF
--- a/deploy/bundle/BUILD.bazel
+++ b/deploy/bundle/BUILD.bazel
@@ -11,7 +11,7 @@ copy_file(
 create_sample_values_binary(
     name = "sample_values",
     input = "//deploy/helm/kubecf:values.yaml",
-    output = ":sample-values.yaml",
+    output = ":kubecf-sample-values.yaml",
 )
 
 pkg_tar(
@@ -21,6 +21,6 @@ pkg_tar(
     srcs = [
         ":cf_operator_chart",
         "//deploy/helm/kubecf:release_chart",
-        ":sample-values.yaml",
+        ":kubecf-sample-values.yaml",
     ]
 )


jandubois:
I think this file belongs inside the kubecf chart, right next to `values.yaml`, just like `imagelist.txt`.

mook-as:
But if this goes in the kubecf chart, nobody will ever see it (because helm can install from the tgz directly).

Of course, the _saner_ thing would be for helm to have a more developed concept where sample values.yaml files (and also how that interacts with subcharts, but that's a different discussion altogether…)

jandubois:
One thing I'm wondering is if we shouldn't have commented out most values inside `values.yaml` anyways, and provide them inside the templates with

```
{{ default .Variables.my_setting "default_value" }}
```

Unfortunately this doesn't work for configuring subcharts unless we are willing to modify the subcharts themselves. helm is unfortunately quite limited in this area.

jandubois:
> Unfortunately this doesn't work for configuring subcharts unless we are willing to modify the subcharts themselves.

We don't actually need to modify the charts; we can just rewrite the `values.yaml` files of the subcharts when we package them up.

I'm now convinced that we should try first to slim down the top-level `values.yaml` and move out defaults to the templates. We might find that we don't need a separate `sample-values.yaml` anymore at that time.

Bonus feature: `helm show values kubecf` will show something sensible.

mook-as:
Closing since we want to do #916 today :)

Issue #907, opened 2020-05-21
Improve sample values output: Block-oriented comments


mook-as:
**Is your feature request related to a problem? Please describe.**
We currently have a [sample values generator](https://github.com/cloudfoundry-incubator/kubecf/blob/master/rules/kubecf/create_sample_values.rb) from #857 that comments out things that are not required for the user:

```yaml
features:
  eirini:
    enabled: false
  # ingress:
    # enabled: false
    # tls:
      # crt: ~
      # key: ~
    # annotations: {}
    # labels: {}
```

**Describe the solution you'd like**

It would be nice if the comment characters (`#`) were lined up in one column so that on column-editing-capable editors they can be uncommented easily.  Something like:

```yaml
features:
  eirini:
    enabled: false
  # ingress:
  #   enabled: false
  #   tls:
  #     crt: ~
  #     key: ~
  #   annotations: {}
  #   labels: {}
```
**Describe alternatives you've considered**

The easy path would be to just make all the `#`s be on the first column; that doesn't look as nice though (especially if there's a single deeply-nested required value).

We probably want to think about how to deal with actual comments in there too. Do we double-comment? Leave it along (and have jagged comment starts)? Move the comment mark around (and make uncommenting annoying)?

The current behaviour was the easiest to implement quickly.

jandubois:
One problem I see with putting the `#` char in a lower column is that it becomes more ambiguous if the commented lines are indented by the comments, or if the `#` just replaced a space, i.e. does

```yaml
features:
  eirini:
    enabled: false
  # ingress:
  #   enabled: false
```
mean

```yaml
features:
  eirini:
    enabled: false
  ingress:
    enabled: false
```
or

```yaml
features:
  eirini:
    enabled: false
    ingress:
      enabled: false
```

Hmm, I guess it is ambiguous whatever you do. 😦 Suggestions welcome!

PR #911, opened 2020-05-22
feat: Revert 'Cherry Pick "Bundle sample values #857" #898'

jandubois:
This feature was creating packaging problems for SUSE, and the general concept is still controversial. See #907 and #908.


jandubois:
We decided not to revert, but to move the `sample-values.yaml` file from the bundle to inside the kubecf helm chart: #917

PR #919, opened 2020-05-22
feat: sample values: render markdown config help

mook-as:
## Description
This lets the sample values script render a markdown table with the values that have descriptions.  Kind of fragile right now and lists things we don't really want in the table, but it's a start.

## Motivation and Context
We want to be able to have a README in the helm chart with the values for the user.

## How Has This Been Tested?
Manually ran the script:
```bash
MODE=help ruby rules/kubecf/create_sample_values.rb deploy/helm/kubecf/values.yaml
```
Pasted the output into a GitHub issue comment box to test markdown rendering.

Parameter | Description
--- | ---
`system_domain` | REQUIRED: the domain that the deployment will be visible to the user.
`properties` | Set or override job properties. The first level of the map is the instance group name. The second  level of the map is the job name. E.g.:  properties:  adapter:  adapter:  scalablesyslog:  adapter:  logs:  addr: kubecf-log-api:8082
`kube.storage_class` | The storage class to be used for the instance groups that need it (e.g. bits, database and  singleton-blobstore). If it's not set, the default storage class will be used.
`sizing.diego_cell.ephemeral_disk.size` | Size of the ephemeral disk used to store applications in MB
`sizing.diego_cell.ephemeral_disk.storage_class` | The name of the storage class used for the ephemeral disk PVC.
`settings.router.tls` | tls sets up the public TLS for the router. The tls keys:  crt: the certificate in the PEM format. Required.  key: the private key in the PEM format. Required.
`features.blobstore.provider` | Possible values for provider: singleton and s3.
`features.blobstore.s3.app_package_directory_key` | The following values are used as S3 bucket names.
`operations.custom` | A list of configmap names that should be applied to the BOSH manifest.
`operations.inline` | Inlined operations that get into generated ConfigMaps. E.g. adding a password variable:  operations:  inline:  - type: replace  path: /variables/-  value:  name: my_password  type: password

## Types of changes

## Checklist:

### DIFF
--- a/rules/kubecf/create_sample_values.rb
+++ b/rules/kubecf/create_sample_values.rb
@@ -14,9 +14,12 @@
 
 # If the output file is not given, the result is printed on standard out.
 
-# If the input file is not given, tests are run instead.  If the environment
-# variable `DEBUG` is set, and the input file is given, internal debugging
-# information is printed (to standard output) instead of the adjusted result.
+# If the input file is not given, tests are run instead.
+
+# The output (when not runnning tests) depends on the environment variable
+# `MODE`.  If it is set to `DEBUG`, internal debugging information is printed
+# (to standard output).  If it is set to `HELP`, a markdown table is printed
+# for documentation.  Otherwise, it prints out the sample values.yaml.
 
 require 'yaml'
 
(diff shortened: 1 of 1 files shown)


mook-as:
Closing because #1056 landed, and that was based on this branch.

PR #943, opened 2020-05-27
feat: updating kube ready state check

bikramnehra:
Following changes have been added:
- adding kernel version check.
- adding max user namespaces check.
- removing privileged pod check.

## Description

The kube ready state check script must be updated to reflect recent changes.

## Motivation and Context

The idea is to introduce new required checks and remove unnecessary checks.

## How Has This Been Tested?

Local execution by pointing to a `kubeconfig`.

## Types of changes

## Checklist:

### DIFF
--- a/dev/kube/kube-ready-state-check.sh
+++ b/dev/kube/kube-ready-state-check.sh
@@ -99,13 +99,13 @@ if having_category node; then
     fi
 fi
 
-# max_user_namespaces must be >= 10000, see https://github.com/cloudfoundry-incubator/kubecf/issues/484
+# max_user_namespaces must be >= 0, see https://github.com/cloudfoundry-incubator/kubecf/issues/484
 if having_category node ; then
     ns=$(cat /proc/sys/user/max_user_namespaces)
-    if [ $((ns)) -ge 10000 ] ; then
-        status "user_max_namespaces set to greater than 10000"
+    if [ $((ns)) -ge 0 ] ; then
+        status "user_max_namespaces set to greater than 0"
     else 
-        trouble "user_max_namespaces should be set to more than 10000"
+        trouble "user_max_namespaces should be set to more than 0"
     fi
 fi
 
--- a/dev/kube/kube-ready-state-check.sh
+++ b/dev/kube/kube-ready-state-check.sh
@@ -149,8 +149,9 @@ if having_category kube ; then
 fi
 
 # privileged pods are enabled in K8s
-if having_category api ; then	
-    pgrep -ax 'hyperkube|(kube-)?apiserver' | grep apiserver | grep --silent -- --allow-privileged	
+if having_category api ; then
+    # checks for `--allow-privileged` negating cases like `--allow-priviledged=false` or `--allow-priviledged false`
+    pgrep -ax 'hyperkube|(kube-)?apiserver' | grep --silent -vE -- "--allow-privileged((=| +)false)"
     status "Privileged must be enabled in 'kube-apiserver'"	
 fi
 
--- a/dev/kube/kube-ready-state-check.sh
+++ b/dev/kube/kube-ready-state-check.sh
@@ -88,6 +88,27 @@ function having_category() {
 
 echo "Testing $(green "${category}")"
 
+# kernel version should be >= 3.19
+if having_category node; then
+    desired_kernel_version=3.19
+    version=$(uname -r | cut -c1-4)
+    if (( $(awk 'BEGIN {print ("'"$version"'" >= "'$desired_kernel_version'")}') )); then
+        status "kernel version must be equal to or greater than 3.19"
+    else
+        trouble "kernel version should not be lower than 3.19"
+    fi
+fi
+
+# max_user_namespaces must be >= 10000
+if having_category node ; then
+    ns=$(cat /proc/sys/user/max_user_namespaces)
+    if [ $((ns)) -ge 10000 ] ; then
+        status "user_max_namespaces set to greater than 10000"
+    else 
+        trouble "user_max_namespaces should be set to more than 10000"
+    fi
+fi
+
 # swap should be accounted
 if having_category node ; then
     # https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt - section 2.4.
@@ -109,6 +130,7 @@ if having_category kube ; then
 fi
 
 # kube-dns shows all pods ready
+# This check should handle core-dns pods as well,see https://github.com/coredns/deployment/issues/116
 if having_category kube ; then
     kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns 2> /dev/null | grep -Eq '([0-9])/\1 *Running'
     status "all kube-dns pods should be running (show N/N ready)"
@@ -124,17 +146,6 @@ fi
 if having_category kube ; then
     test ! "$(kubectl get storageclasses 2>&1 | grep -e "No resources found." -e "Unable to connect to the server")"
     status "A storage class should exist in K8s"
-fi
-
-# privileged pods are enabled in K8s
-if having_category api ; then
-    pgrep -ax 'hyperkube|apiserver' | grep apiserver | grep --silent -- --allow-privileged
-    status "Privileged must be enabled in 'kube-apiserver'"
-fi
-
-if having_category node ; then
-    pgrep -ax 'hyperkube|kubelet' | grep kubelet | grep --silent -- --allow-privileged
-    status "Privileged must be enabled in 'kubelet'"
 fi
 
 # override tasks infinity in systemd configuration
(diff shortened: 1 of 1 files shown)


dev/kube/kube-ready-state-check.sh
```diff
@@ -126,17 +148,6 @@ if having_category kube ; then
     status "A storage class should exist in K8s"
 fi
 
-# privileged pods are enabled in K8s
```
satadruroy, line 129: You also removed this check for the privileged container (`apiserver --allow-privileged`) - that needs to be put back in, I think? Not sure what to do with the kubelet privileged check as it only applies to kube versions < 1.15

jandubois: I'm not sure how robust the checks using `pgrep -ax 'hyperkube|apiserver'` really are.

Some failure modes:

* the process is not called `apiserver`, but e.g. `kube-apiserver`
* the flag is listed, but set to `false` via e.g. `--allow-privileged=false`

The first is true on CaaSP 4. The second isn't, but you would not catch it if it was:

```
vagrant@caasp4-master-1:~> pgrep -ax kube-apiserver
7463 kube-apiserver --advertise-address=192.168.121.120 --allow-privileged=true --authorization-mode=Node,RBAC --client-ca-file=/etc/kubernetes/pki/ca.crt --enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,TaintNodesByCondition,Priority,DefaultTolerationSeconds,DefaultStorageClass,PersistentVolumeClaimResize,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota,StorageObjectInUseProtection,RuntimeClass,NodeRestriction,PodSecurityPolicy --enable-bootstrap-token-auth=true --etcd-cafile=/etc/kubernetes/pki/etcd/ca.crt --etcd-certfile=/etc/kubernetes/pki/apiserver-etcd-client.crt --etcd-keyfile=/etc/kubernetes/pki/apiserver-etcd-client.key --etcd-servers=https://127.0.0.1:2379 --insecure-port=0 --kubelet-client-certificate=/etc/kubernetes/pki/apiserver-kubelet-client.crt --kubelet-client-key=/etc/kubernetes/pki/apiserver-kubelet-client.key --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname --oidc-ca-file=/etc/kubernetes/pki/ca.crt --oidc-client-id=oidc --oidc-groups-claim=groups --oidc-issuer-url=https://caasp4-lb-1:32000 --oidc-username-claim=email --proxy-client-cert-file=/etc/kubernetes/pki/front-proxy-client.crt --proxy-client-key-file=/etc/kubernetes/pki/front-proxy-client.key --requestheader-allowed-names=front-proxy-client --requestheader-client-ca-file=/etc/kubernetes/pki/front-proxy-ca.crt --requestheader-extra-headers-prefix=X-Remote-Extra- --requestheader-group-headers=X-Remote-Group --requestheader-username-headers=X-Remote-User --secure-port=6443 --service-account-key-file=/etc/kubernetes/pki/sa.pub --service-cluster-ip-range=10.96.0.0/12 --tls-cert-file=/etc/kubernetes/pki/apiserver.crt --tls-private-key-file=/etc/kubernetes/pki/apiserver.key
```

On CaaSP `kube-apiserver`, `kube-scheduler` etc. all run inside pods started by `kubelet`:

```
vagrant@caasp4-master-1:~> sudo grep -r allow-privileged /etc/kubernetes/manifests
/etc/kubernetes/manifests/kube-apiserver.yaml:    - --allow-privileged=true
```

I assume the objection did not extend to the `kubelet` check, as it's `--allow-privileged` option was removed in kube 1.15.

jandubois: > * the process is not called `apiserver`, but e.g. `kube-apiserver`

This could be fixed by `pgrep -ax 'hyperkube|(kube-)?apiserver'`, but it remains a heuristic, which means it sometimes will still not do the right thing.

bikramnehra:
> I just realized we use `docker info` to check the storage driver but with a CRI-O container runtime docker would not be installed so that will be a false error - we may have to a `crictl info` check (I haven't run it to see what it shows).
> 
> https://github.com/cloudfoundry-incubator/kubecf/blob/f4abc43728ff4aaac813f89c3e21150f1b499db1/dev/kube/kube-ready-state-check.sh#L122

This is what I saw from `crictl info`:

```
docker@minikube:~$ sudo crictl info
{
  "status": {
    "conditions": [
      {
        "type": "RuntimeReady",
        "status": true,
        "reason": "",
        "message": ""
      },
      {
        "type": "NetworkReady",
        "status": true,
        "reason": "",
        "message": ""
      }
    ]
  }
}
```
Also the fact that it requires root privileges to run:

`docker@minikube:~$ crictl info
FATA[0002] failed to connect: failed to connect, make sure you are running as root and the runtime has been started: context deadline exceeded`

Thus far I have not been able to find a reliable way to fetch `storage_class` for `crio`. May be we can use `/etc/crio/crio.conf` but I am not sure if it would account any overrides at runtime.

@jandubois feel free to suggest in case you have any good options.

dev/kube/kube-ready-state-check.sh
bikramnehra: I agree with your first point - changes pushed already.
However, the second point might be bit of a stretch as we cannot account for all the possible configurations on different platforms.

jandubois:
> Thus far I have not been able to find a reliable way to fetch `storage_class` for `crio`. May be we can use `/etc/crio/crio.conf` but I am not sure if it would account any overrides at runtime.

I think it may be `/etc/containers/storage.conf`, but that is also just an override for `/usr/share/...` file etc. I think trying to parse config files is futile; you need to query the running system, if possible (or give up checking, if you can't).

> @jandubois feel free to suggest in case you have any good options.

I have no idea; I would suggest asking the SUSE CaaSP team, as they have at least one CRIO committer on their team, iirc.

dev/kube/kube-ready-state-check.sh
jandubois: > the second point might be bit of a stretch as we cannot account for all the possible configurations on different platforms.

I don't know what the second point is that you refer to here, but the whole point of this script is checking any arbitrary configuration for compatibility with kubecf. If you don't mind detecting the values correctly, why bother checking for them at all.

Here is a simple test that will treat `--allow-privileged=false` correctly:

```
# echo '--allow-privileged' | perl -ne 'exit(!/--allow-privileged(?!(=| +)false)/)'; echo $?
0
# echo '--allow-privileged --foo' | perl -ne 'exit(!/--allow-privileged(?!(=| +)false)/)'; echo $?
0
# echo '--allow-privileged=true' | perl -ne 'exit(!/--allow-privileged(?!(=| +)false)/)'; echo $?
0
# echo '--allow-privileged=false' | perl -ne 'exit(!/--allow-privileged(?!(=| +)false)/)'; echo $?
1
# echo '--allow-privileged  false' | perl -ne 'exit(!/--allow-privileged(?!(=| +)false)/)'; echo $?
1
```

It uses a negative look-ahead assertion to make sure `--allow-privileged` is not followed by `=` or one or more spaces, followed by the string `false`.

bikramnehra: updated check.

satadruroy:
> > I just realized we use `docker info` to check the storage driver but with a CRI-O container runtime docker would not be installed so that will be a false error - we may have to a `crictl info` check (I haven't run it to see what it shows).
> > https://github.com/cloudfoundry-incubator/kubecf/blob/f4abc43728ff4aaac813f89c3e21150f1b499db1/dev/kube/kube-ready-state-check.sh#L122
> 
> This is what I saw from `crictl info`:
> 
> ```
> docker@minikube:~$ sudo crictl info
> {
>   "status": {
>     "conditions": [
>       {
>         "type": "RuntimeReady",
>         "status": true,
>         "reason": "",
>         "message": ""
>       },
>       {
>         "type": "NetworkReady",
>         "status": true,
>         "reason": "",
>         "message": ""
>       }
>     ]
>   }
> }
> ```
> 
> Also the fact that it requires root privileges to run:
> 
> `docker@minikube:~$ crictl info FATA[0002] failed to connect: failed to connect, make sure you are running as root and the runtime has been started: context deadline exceeded`
> 
> Thus far I have not been able to find a reliable way to fetch `storage_class` for `crio`. May be we can use `/etc/crio/crio.conf` but I am not sure if it would account any overrides at runtime.
> 
> @jandubois feel free to suggest in case you have any good options.

`podman info` gives about the same info as `docker info` does including the FS storage driver but I'm not sure we can expect `podman` to be installed either in a CRI-O setup (`crictl` yes but not necessarily `podman`) 

https://www.openshift.com/blog/crictl-vs-podman

PR #945, opened 2020-05-28
When suse_buildpacks are enabled (the default) use sle15 as the default stack

jimmykarily:
## Description

Right now the default stack for diego is cflinuxfs3. For Eirini we already change to sle15 (https://github.com/cloudfoundry-incubator/kubecf/blob/master/deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml#L6)

## Motivation and Context

Consistency

## How Has This Been Tested?

Deploy with default values and test what the default stack is.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -1,7 +1,4 @@
 {{- if .Values.features.suse_default_stack.enabled }}
-{{- if not .Values.features.suse_buildpacks.enabled }}
-{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is true" }}
-{{- end }}
 {{- $cc_jobs := list "cloud_controller_ng" }}
 {{- $cc_jobs = append $cc_jobs "cloud_controller_worker" }}
 {{- $cc_jobs = append $cc_jobs "cloud_controller_clock" }}
@@ -10,6 +7,9 @@
 - type: replace
    path: /instance_groups/name=api/jobs/name={{ $job }}/properties/cc/default_stack?
    value: sle15
+{{- end }}
+{{- if not .Values.features.suse_buildpacks.enabled }}
+{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is set" }}
 {{- end }}
 {{- end }}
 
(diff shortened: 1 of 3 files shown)


jandubois:
Using `--set features.suse_buildpacks.enabled=true` to change the default stack feels like a hack to me. I think it should either have a separate feature flag, so you can still install SUSE buildpacks even when `cflinuxfs3` is the default, or we should rename the feature flag to something like `features.suse_stack`.

jandubois:
## Suggestion

After in-person discussions we decided that we can't change the name of the existing feature flag because it would break backwards compatibility. Adding a second feature flag to use `sle15` instead of `cflinuxfs3` as the default stack would therefore be required, e.g. `features.suse_default_stack.enabled=true`.

In v2.2.x the `suse_default_stack` default must be `false` for backwards compatibility, but should be changed to `true` in v2.3+ because it will be required for Eirini support.

Proposed functionality for all different combinations:

### Diego

#### `suse_buildpacks=false` and `suse_default_stack=false`

However, buildpacks can be added back later via `cf create-buildpack`, but stacks cannot, so maybe both stacks should always be available.

#### `suse_buildpacks=true` and `suse_default_stack=false` (current default)

SUSE buildpacks would be installed, but not used by default because the default stack remains `cflinuxfs3`.

#### `suse_buildpacks=false` and `suse_default_stack=true`

#### `suse_buildpacks=true` and `suse_default_stack=true` (default in v2.3+)

### Eirini

#### `suse_buildpacks=false` and `suse_default_stack=false`

#### `suse_buildpacks=true` and `suse_default_stack=false` (current default)

#### `suse_buildpacks=false` and `suse_default_stack=true`

#### `suse_buildpacks=true` and `suse_default_stack=true` (default in v2.3+)

@viovanov please review and comment!

jandubois:
Has been replaced by #966

jandubois:
> #### `suse_buildpacks=false` and `suse_default_stack=false`
>
> SUSE buildpacks are not installed, but the SUSE rootfs still is. We can discuss if the rootfs should be removed as well in the future, but in v2.2.x it has to stay for backwards compatibility.
> 
> However, buildpacks can be added back later via `cf create-buildpack`, but stacks cannot, so maybe both stacks should always be available.

This turns out to be not the case. The SUSE stack is not configured when `suse_buildpacks=false`. So it is not possible to add buildpacks later.

I think this is fine, but just wanted to mention it here that the previous claims were incorrect.

cc @troytop @gaktive

PR #958, opened 2020-05-29
CI: fix filename for suse-java-buildpack.

bikramnehra:
## Description

This PR adds an exception for `filepath` of `suse-java-buildpack` while bumping its release image. Since [packaging](https://github.com/SUSE/cf-java-buildpack-release/blob/master/packages/java-buildpack-sle15/packaging) for java buildpack is different from others, we can't simply rely on filepaths from the buildpack package.

Also see: https://github.com/cloudfoundry-incubator/kubecf/pull/952

## Motivation and Context

This creates a problem while bumping buildpacks and creates a commit like this https://github.com/cloudfoundry-incubator/kubecf/pull/851/files which is wrong.

## How Has This Been Tested?

https://concourse.suse.dev/

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/suse-buildpacks-ci/buildpack-version-bump/pipeline.yaml.gomplate
+++ b/.concourse/suse-buildpacks-ci/buildpack-version-bump/pipeline.yaml.gomplate
@@ -43,7 +43,7 @@ resources:
   source:
     owner: SUSE
     repository: {{ $release.repo }}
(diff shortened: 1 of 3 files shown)


jimmykarily:
I just wonder if it would be better to change the name of the zip file in the bosh release to make it consistent. This way we wouldn't need any special handling here. I don't have a strong opinion though. Approving.

PR #960, opened 2020-05-29
Add suse_default_stack feature flag

### LINKED ISSUE #948: Incompatible stack/buildpack settings should throw an error in the helm chart
It's easy for the user to install kubecf using incompatible settings for buildpacks, stacks and scheduler.

We should implement @jandubois' suggestions in the link below, that the helm chart should error when incompatible combinations are used.

n/a

**Additional context**
https://github.com/cloudfoundry-incubator/kubecf/pull/945#issuecomment-635624549

jandubois:
It switches the default_stack from cflinuxfs3 to sle15.

Fixes #948.

https://github.com/cloudfoundry-incubator/kubecf/pull/945#issuecomment-635624549
specifies the expected functionality:

### Diego

* `suse_buildpacks=false` and `suse_default_stack=false`

However, buildpacks can be added back later via cf create-buildpack, but stacks cannot, so maybe both stacks should always be available.

* `suse_buildpacks=true` and `suse_default_stack=false` (current default)

SUSE buildpacks would be installed, but not used by default because the default stack remains cflinuxfs3.

* `suse_buildpacks=false` and `suse_default_stack=true`

* `suse_buildpacks=true` and `suse_default_stack=true` (default in v2.3+)

### Eirini

* `suse_buildpacks=false` and `suse_default_stack=false`

* `suse_buildpacks=true` and `suse_default_stack=false` (current default)

* `suse_buildpacks=false` and `suse_default_stack=true`

* `suse_buildpacks=true` and `suse_default_stack=true` (default in v2.3+)

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -1,3 +1,18 @@
+{{- if .Values.features.suse_default_stack.enabled }}
+{{- $cc_jobs := list "cloud_controller_ng" }}
+{{- $cc_jobs = append $cc_jobs "cloud_controller_worker" }}
+{{- $cc_jobs = append $cc_jobs "cloud_controller_clock" }}
+{{- $cc_jobs = append $cc_jobs "cc_deployment_updater" }}
+{{- range $job := $cc_jobs }}
+- type: replace
+   path: /instance_groups/name=api/jobs/name={{ $job }}/properties/cc/default_stack?
+   value: sle15
+{{- end }}
+{{- if not .Values.features.suse_buildpacks.enabled }}
+{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is set" }}
+{{- end }}
+{{- end }}
+
 {{- if .Values.features.suse_buildpacks.enabled }}
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/stacks/-
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -1,4 +1,7 @@
 {{- if .Values.features.suse_default_stack.enabled }}
+{{- if not .Values.features.suse_buildpacks.enabled }}
+{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is true" }}
+{{- end }}
 {{- $cc_jobs := list "cloud_controller_ng" }}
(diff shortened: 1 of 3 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -1,3 +1,18 @@
...
+{{- $cc_jobs = append $cc_jobs "cloud_controller_clock" }}
+{{- $cc_jobs = append $cc_jobs "cc_deployment_updater" }}
+{{- range $job := $cc_jobs }}
+- type: replace
+   path: /instance_groups/name=api/jobs/name={{ $job }}/properties/cc/default_stack?
+   value: sle15
+{{- end }}
+{{- if not .Values.features.suse_buildpacks.enabled }}
+{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is set" }}
+{{- end }}
```
f0rmiga, line 13: Could you move this block to the top, right after the check on the `suse_default_stack.enabled`? It reads better.

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -1,3 +1,18 @@
...
+{{- $cc_jobs = append $cc_jobs "cloud_controller_worker" }}
+{{- $cc_jobs = append $cc_jobs "cloud_controller_clock" }}
+{{- $cc_jobs = append $cc_jobs "cc_deployment_updater" }}
+{{- range $job := $cc_jobs }}
+- type: replace
+   path: /instance_groups/name=api/jobs/name={{ $job }}/properties/cc/default_stack?
+   value: sle15
+{{- end }}
+{{- if not .Values.features.suse_buildpacks.enabled }}
+{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is set" }}
```
f0rmiga, line 12:
```suggestion
{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is true" }}
```

deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml
```diff
@@ -2,13 +2,13 @@
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/opi?/enabled?
   value: true
-# XXX This check is disabled until v2.3.0 for backwards compatibility reasons.
+# TODO(jandubois) This check is disabled until v2.3.0 for backwards compatibility reasons.
```
f0rmiga, line 5: You didn't need to do that on every single line, but, fine!

PR #965, opened 2020-05-30
fix: image_list.rb only needs to check 2 feature combinations

jandubois:
The script already takes almost 10 minutes to run, and runtime will scale exponentially as more feature flags are added. Testing shows that all possible image names are detected by just expanding the templates twice: once with all features disabled, and once with all of them enabled.

Future changes are also expected to throw errors for certain feature combinations, which the current code also doesn't deal with.

### DIFF
--- a/rules/kubecf/image_list.tmpl.rb
+++ b/rules/kubecf/image_list.tmpl.rb
@@ -251,10 +251,12 @@ class ConfigMap < Resource
   end
 end
 
-# Create all permutations for enabling and disabling the features.
+# Running through all permutations takes very long, and makes it impractical to
+# add additional feature flags. Some flag combinations also could throw errors.
+# So far running with features either all enabled or all disabled generates
+# the complete set of used images and is nearly instantaneous.
 features = values['features'].keys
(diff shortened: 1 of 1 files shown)


rules/kubecf/image_list.tmpl.rb
```diff
@@ -256,6 +256,12 @@ def images
 permutations = [true, false].repeated_permutation(features.size)
                             .map { |v| features.zip(v).to_h }
 
+# Running through all permutations takes very long, and makes it impractical to
+# add additional feature flags. Some flag combinations also could throw errors.
+# So far running with features either all enabled or all disabled generates
+# the complete set of used images and is nearly instantaneous.
+permutations = [ features.to_h{ |x| [x, false] }, features.to_h{ |x| [x, true] } ]
+
```
andreas-kupries, line 264: Is line 256 (just before the added block) still needed ? I do not believe so, as the new block overrides it.

jandubois: It isn't needed, and I've removed it now; the implementation remains accessible in the git history...

PR #966, opened 2020-05-30
feat: Implement features.suse_default_stack.enabled

### LINKED ISSUE #948: Incompatible stack/buildpack settings should throw an error in the helm chart

jandubois:
This PR uses #965 as the base, so that PR should be reviewed/merged first.

It switches the default_stack from cflinuxfs3 to sle15.

Fixes #948.

This is a new version of #960, which was merged without testing, and later reverted.

https://github.com/cloudfoundry-incubator/kubecf/pull/945#issuecomment-635624549 specifies the expected functionality:

### Diego

* `suse_buildpacks=false` and `suse_default_stack=false`

However, buildpacks can be added back later via cf create-buildpack, but stacks cannot, so maybe both stacks should always be available.

* `suse_buildpacks=true` and `suse_default_stack=false` (current default)

SUSE buildpacks would be installed, but not used by default because the default stack remains cflinuxfs3.

* `suse_buildpacks=false` and `suse_default_stack=true`

* `suse_buildpacks=true` and `suse_default_stack=true` (default in v2.3+)

### Eirini

* `suse_buildpacks=false` and `suse_default_stack=false`

* `suse_buildpacks=true` and `suse_default_stack=false` (current default)

* `suse_buildpacks=false` and `suse_default_stack=true`

* `suse_buildpacks=true` and `suse_default_stack=true` (default in v2.3+)

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
@@ -1,3 +1,24 @@
+# TODO(jandubois) Starting with v2.3.0 enabling eirini will no longer automatically switch to the suse stack.
+{{- if or .Values.features.suse_default_stack.enabled .Values.features.eirini.enabled }}
+
+{{- if not .Values.features.suse_buildpacks.enabled }}
+{{- fail "feature.suse_buildpacks.enabled must be true when features.suse_default_stack.enabled is true" }}
+{{- end }}
+
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/default_stack?
+  value: sle15
+- type: replace
+  path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/properties/cc/default_stack?
+  value: sle15
+- type: replace
+  path: /instance_groups/name=scheduler/jobs/name=cloud_controller_clock/properties/cc/default_stack?
+  value: sle15
+- type: replace
+  path: /instance_groups/name=scheduler/jobs/name=cc_deployment_updater/properties/cc/default_stack?
+  value: sle15
+{{- end }}
+
 {{- if .Values.features.suse_buildpacks.enabled }}
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/stacks/-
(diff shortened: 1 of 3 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
```diff
@@ -1,3 +1,24 @@
+# TODO(jandubois) Starting with v2.3.0 enabing eirini will no longer automatically switch to the suse stack.
```
andreas-kupries, line 1: Trivial typo: `enab(l)ing`.

jandubois:
> I don't see `suse_default_stack` in values.yaml.

Thanks for catching that. It was in the original PR https://github.com/cloudfoundry-incubator/kubecf/pull/960 but must have gotten lost in the rebasing. I've put it back and amended the commit.

> Also, I don't think we should call it `suse_default_stack`.
> It should be `default_stack` with a set of acceptable values - some folks want to add a red hat stack (can't find the slack conversation anymore).

I do agree with that. Unfortunately we only support 2 hard-coded stacks right now, and adding additional stacks is not possible without some complicated ops files.

I do want us to parameterize additional stack support in the future, but it will require reorganizing the current structure of defining releases in `values.yaml` and making things like `feature.suse_buildpacks` more generic as well.

So for the current point in time I think `features.suse_default_stack` works well as a companion to `features.suse_buildpacks`. Without it being SUSE stack specific, the particular failure checks would also not make much sense.

We can migrate `suse_default_stack` and `suse_buildpacks` to a more generic configuration at a later point in time.

deploy/helm/kubecf/assets/operations/instance_groups/api.yaml
jandubois: Fixed in amended commit.

gaktive:
I believe the main conditions of @viovanov's review have been met. We'll need a feature item to expand past 2 stacks since there will be substantial changes needed in quarks to prepare for it, which shouldn't hold up the current needs for kubecf.

Issue #928, opened 2020-05-25
Autoscaler goes into a CLBO if we run helm upgrade to enable it (after dns setup)


prabalsharma:
**Describe the bug**
Autoscaler goes into a CLBO if we run helm upgrade to enable it on a running cluster with DNS setup.
`{"API":"https://cloud-controller-ng.service.cf.internal:9024","error":"request client credential grant failed: https://uaa.prabal-gke.ci.kubecf.charmedquarks.me/oauth/token [401] 401 Unauthorized"`

Similar to https://github.com/cloudfoundry-incubator/kubecf/issues/897

**To Reproduce**

- Deploy HA without autoscaler.
- Set up DNS ( cf login should work)
- Do a helm upgrade to turn autoscaler on.

**Expected behavior**

```
scf           asactors-0                                                        5/6     CrashLoopBackOff   26         71m
scf           asapi-0                                                           0/6     Init:6/12          2          71m
scf           asdatabase-0                                                      4/4     Running            0          71m
scf           asmetrics-0                                                       0/5     Init:5/10          2          72m
scf           asnozzle-0                                                        0/4     Init:5/9           2          70m
```

```
2020/05/25 21:15:34, /var/vcap/sys/log/scalingengine_ctl.log, [2020-05-25 21:15:33+0000] Removing stale pidfile...
2020/05/25 21:15:34, /var/vcap/sys/log/scalingengine/scalingengine.stdout.log, {"data":{"form":{"client_id":["app_autoscaler"],"client_secret":"*REDACTED*","grant_type":["client_credentials"]},"session":"1","tokenURL":"*REDACTED*"},"log_level":1,"log_time":"2020-05-25T21:15:34Z","message":"scalingengine.cf.request-client-credential-grant","source":"scalingengine","timestamp":"1590441334.403366089"}
2020/05/25 21:15:34, /var/vcap/sys/log/scalingengine/scalingengine.stdout.log, {"data":{"error":"request client credential grant failed: https://uaa.prabal-gke.ci.kubecf.charmedquarks.me/oauth/token [401] 401 Unauthorized","session":"1"},"log_level":2,"log_time":"2020-05-25T21:15:34Z","message":"scalingengine.cf.request-client-credential-grant-response","source":"scalingengine","timestamp":"1590441334.771143198"}
[...]

PR #971, opened 2020-06-01
Fix CLBO in autoscaler after transient failure to connect to UAA

andreas-kupries:
## Description

Added autoscaler patches to render `pid_guard` in scalingengine and operator inert.

Prevents a platform from failing to properly restart these jobs at container level after a transient failure (uaa connectivity, for example) due to the new pid for the control/main process being the same as in the failed container, and thus aborting the restart in the mistaken belief of the previous process still running (while actually looking at the new process).

A side change is an extension of the `klog.sh` tool with a new option `-r` to retrieve not just logs from a CF cluster, but also the rendered scripts and configuration files.

## Motivation and Context

See #928 

## How Has This Been Tested?

Manually via catapult, see ticket comment https://github.com/cloudfoundry-incubator/kubecf/issues/928#issuecomment-634309394 Main difference from the reproduction is the source of the bundle containing the patched chart. This was generated via `bazel build //deploy/bundle:kubecf-bundle` and catapult then pointed to this new chart.

## Types of changes

## Checklist:

### DIFF
--- a/bosh/releases/pre_render_scripts/asactors/operator/jobs/patch_operator_ctl.sh
+++ b/bosh/releases/pre_render_scripts/asactors/operator/jobs/patch_operator_ctl.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+target="/var/vcap/all-releases/jobs-src/app-autoscaler/operator/templates/operator_ctl"
+sentinel="${target}.patch_sentinel"
+if [[ -f "${sentinel}" ]]; then
+  if sha256sum --check "${sentinel}" ; then
+    echo "Patch already applied. Skipping"
+    exit 0
+  fi
+  echo "Sentinel mismatch, re-patching"
+fi
+
+sed -i "s/pid_guard/#pid_guard/g" "${target}"
+
+sha256sum "${target}" > "${sentinel}"
--- a/bosh/releases/pre_render_scripts/asactors/scalingengine/jobs/patch_scalingengine_ctl.sh
+++ b/bosh/releases/pre_render_scripts/asactors/scalingengine/jobs/patch_scalingengine_ctl.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+target="/var/vcap/all-releases/jobs-src/app-autoscaler/scalingengine/templates/scalingengine_ctl"
+sentinel="${target}.patch_sentinel"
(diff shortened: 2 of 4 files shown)


bosh/releases/pre_render_scripts/asactors/operator/jobs/patch_operator_ctl.sh
```diff
@@ -0,0 +1,17 @@
...
+sentinel="${target}.patch_sentinel"
+if [[ -f "${sentinel}" ]]; then
+  if sha256sum --check "${sentinel}" ; then
+    echo "Patch already applied. Skipping"
+    exit 0
+  fi
+  echo "Sentinel mismatch, re-patching"
+fi
+
+sed -i "s/pid_guard/#pid_guard/g" "${target}"
```
mook-as, line 15: For context, this is patching https://github.com/cloudfoundry/app-autoscaler-release/blob/3ae61ae696e0e067b3253a692a4be1566f7e2e0e/jobs/operator/templates/operator_ctl#L19

Where `pid_guard` is implemented in https://github.com/cloudfoundry/app-autoscaler-release/blob/master/src/common/utils.sh

Ultimately, then, our issue is that `/var/vcap/sys/run/` isn't expected to be reused between containers.  It looks like we just have a mount for [all of `/var/vcap/sys`](https://github.com/cloudfoundry-incubator/quarks-operator/blob/v4.5.6/pkg/bosh/manifest/job.go#L17) and nothing for `/var/vap/sys/run` in particular, so it gets reused.

Problem is that all volume mounts are attached to the pod (not the container), so they would get re-used across container restarts…  Not sure if there's a good way to go other than just patch the script.

andreas-kupries:
> but I can't think of a good way to do it that is even more hacky…

Neither could I

> Have it check that the executable has the correct leaf name

Oh, you mean distinguish between the `_ctl` process vs the process `exec`d from it to see that we are looking at the startup, and not the main.

> I mean, I guess since it's busted anyway, maybe patch /var/vcap/packages/common/utils.sh directly

I feared here that changing the common part could affect things which are working, and somehow breaking that. Bout of paranoia, maybe.

bosh/releases/pre_render_scripts/asactors/operator/jobs/patch_operator_ctl.sh
andreas-kupries: Yep, and the same in `scalingengine_ctl`.
Yep for `pid_guard` location.
Other way than patching the scripts - I did not see any either.

PR #972, opened 2020-06-01
feat: Cherry Pick "Implement features.suse_default_stack.enabled #966"

jandubois:
See #966 

This PR is based on the branch for #970 which should be merged first!


gaktive:
CATS are failing due to some CI flakiness. Let's see if we can the subsequent CI to have better luck.

PR #974, opened 2020-06-02
Use a separate semver resource for each deploy job

jimmykarily:
because the creation of a version is not an atomic operation and it
could happen that 2 deploy jobs use the same version. When one of them
fails it will trigger a deletion of the cluster although the other job
is still running. This results if failed jobs which are hard to explain.

Also add missing inputs in job

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -50,23 +50,39 @@ resource_types:
     tag: release
 
 resources:
-- name: semver.gke-cluster
-  type: semver
-  source:
-    driver: s3
-    bucket: {{ if has . "s3_bucket" }}{{ .s3_bucket }}{{ else }}kubecf-ci{{ end }}
-    key: gke-cluster
-    initial_version: 0.0.1
-    access_key_id: ((aws-access-key))
-    secret_access_key: ((aws-secret-key))
-    region_name: {{ if has . "s3_bucket_region" }}{{ .s3_bucket_region }}{{ else }}"eu-central-1"{{ end }}
-
 - name: kubecf-github-release
   type: github-release
   source:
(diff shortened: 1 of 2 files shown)


jimmykarily:
There is still a chance that the same cluster name is used in 2 jobs that run at the same time. That may happen if 2 deploy jobs for the same branch and scheduler trigger almost simultaneously. I think that is very unlikely to happen unless Concourse all of a sudden discovers many versions for which it hadn't run before (e.g. because we added `version: every` on some resource). I think this PR fixes 99% of the cases. After all the conflict was already rare as it was, I guess we will never see the edge case I just described.

jimmykarily:
@mook-as that would be an improvement, true, but we will again hijack some irrelevant resource (this time the pool resource) to achieve something a lot simpler. I think it would be better to generate one artifact at the beginning of the pipeline that will identify the whole build from the beginning to the end. I'm imagining a json file that holds all the information about the build. E.g.

```
scheduler: eirini
branch: master
timestamp: blahblah
... (more)
```
then the name of the file could be something like what you describe. 
This is pretty much what [this resource](https://github.com/suse/concourse-git-queue) was doing but we can limit the responsibility and skip the GitHub status update.

This solution has some benefits:
- We have one place where all the information about a build is kept.
- Each step will trigger the next one automatically (even on manual triggers) because there will always be a new resource version connecting the various jobs.
- There are no name collisions (the original issue here).
- A custom resource can provide any metadata we find useful while debugging (e.g. links to the various used git repos, operator release, command to copy/past and get a kubeconfig, anything).

How about we keep this PRs solution for now until we have a decision on the gke vs ekcp first? A lot of refactoring will happen if we switch back to ekcp so let's not open 2 big stories at the same time.

Issue #738, opened 2020-04-22
cfdot job does not work on kubecf


jandubois:
**Describe the bug**

[cfdot](https://github.com/cloudfoundry/cfdot) is a commandline interface to the Diego BBS API. I wanted to use it to investigate why an app remained in a `starting` state, but no instances got scheduled.

`cfdot` is supposed to be installed by the cfdot job from diego-release into at least the `diego-cell` instance group, but potentially all other diego related ones too.

It uses a [pre-start](https://github.com/cloudfoundry/diego-release/blob/develop/jobs/cfdot/templates/pre-start.erb) script to copy [shell setup](https://github.com/cloudfoundry/diego-release/blob/develop/jobs/cfdot/templates/setup.erb) code to `/etc/profile.d`, which in turn sets up some config variables and puts the `/var/vcap/packages/cfdot/bin` directory on the `PATH`.

**To Reproduce**

Connect to the `diego-cell` pod and verify that neither `/etc/profile.d/cfdot.sh` nor `/var/vcap/packages/cfdot/bin/cfdot` exist.

**Expected behavior**

I don't see a way for the init container running the pre-start script to push the setup script into `/etc/profile.d`.

I would suggest that both (an updated) `cfdot.sh` script as well as the `cfdot` tool itself be copied to `/var/vcap/data/cfdot/bin`. The user will then have to run `source /var/vcap/data/cfdot/bin/cfdot.sh` to get a configured `cfdot` on the `PATH`.

**Additional context**

`cfdot` is purely a debugging tool used by developers (and admins) to investigate problems with the Diego scheduler. It is not used during normal operation, and is not accessible to regular CF users.

andreas-kupries:
Looking through the `cf-deployment.yaml` the `cfdot` job is run in three instance groups:
  - scheduler
  - diego-api
  - diego-cell

andreas-kupries:
PR = #975

PR #975, opened 2020-06-02
Make cfdot available in a properly shared location.

andreas-kupries:
## Description

Patched cfdot job prestart to a shared location accessible in all the container of an instance group. The patches are all the same for the three affected instance groups. (As the job itself is the same).

Added comments to the diego-cell patch setup as a reminder to maintainers as to the intent of the complexity.

## Motivation and Context

See ticket #738 

## How Has This Been Tested?

Testing was done in a local minikube deployment. After the cluster became :heavy_check_mark: all affected instance groups were entered into via `kubectl exec`, the `cfdot.sh` was sourced from its new location and then a few `cfdot` commands invoked (`--help`, `locks`, `task`, `domains`, `cells`, `presences`.).

## Types of changes

## Checklist:

### DIFF
--- a/bosh/releases/pre_render_scripts/diego-api/cfdot/jobs/patch_pre-start.sh
+++ b/bosh/releases/pre_render_scripts/diego-api/cfdot/jobs/patch_pre-start.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+##
+# ATTENTION: This is part of a set of six interconnected patches.
+#            Two files spread over three instance groups.
+# See
+# - bosh/releases/pre_render_scripts/diego-cell/cfdot/jobs
+# - bosh/releases/pre_render_scripts/diego-api/cfdot/jobs
+# - bosh/releases/pre_render_scripts/scheduler/cfdot/jobs
+
+set -o errexit -o nounset
+
+target="/var/vcap/all-releases/jobs-src/diego/cfdot/templates/pre-start.erb"
+sentinel="${target}.patch_sentinel"
+if [[ -f "${sentinel}" ]]; then
+  if sha256sum --check "${sentinel}" ; then
+    echo "Patch already applied. Skipping"
+    exit 0
+  fi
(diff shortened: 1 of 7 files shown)


deploy/helm/kubecf/assets/operations/instance_groups/diego-cell.yaml
```diff
@@ -225,10 +225,12 @@
   value: true
 
 {{- if .Values.features.suse_buildpacks.enabled }}
+# For SUSE buildpacks take all patches
```
jandubois, line 228: The check really should be against `features.suse_default_stack.enabled`, but that is outside the scope of this PR, and besides there are many other locations with this issue.

bosh/releases/pre_render_scripts/diego-api/cfdot/jobs/patch_pre-start.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+##
+# ATTENTION: This is part of a set of six interconnected patches.
+#            Two files spread over three instance groups.
```
jandubois, line 4: I wish these patches were somehow implemented with a template helper function, so we don't have almost identical code in 3 different places. But this is also true for other patches we have, e.g. the buildpacks, that move files to `/var/vcap/data/shared-packages`. So maybe we can consolidate all this later (hah) in a future cleanup PR.

deploy/helm/kubecf/assets/operations/instance_groups/diego-cell.yaml
andreas-kupries: Leaving it to separate PR

bosh/releases/pre_render_scripts/diego-api/cfdot/jobs/patch_pre-start.sh
andreas-kupries: Leaving it to later PR.

PR #982, opened 2020-06-04
Remove version: "every" from publish step artifacts

jimmykarily:
because this together with `passed:` is confusing to Concourse and
doesn't let the step trigger when it should.

See also: https://github.com/concourse/concourse/issues/736#issuecomment-479557411

## Types of changes

## Checklist:


jimmykarily:
According to this though, it should work (?): https://concourse-ci.org/jobs.html#schema.step.get-step.version

```
If set to every, builds will walk through all available versions of the resource. Note that if passed is also configured, it will only step through the versions satisfying the constraints.
```

PR #988, opened 2020-06-05
Delete all unused disks

jimmykarily:
## Description

This PR makes sure we delete not used disks that were created dynamically for PVCs. The reason is that when preemptible nodes are used the disk don't get deleted automatically (https://cloud.google.com/compute/docs/instances/preemptible#understanding_the_preemption_process). There is one more disk for the master node but that one gets deleted automatically (checked).

Another solution would be to not use preemptible nodes.

## Motivation and Context

Disks are left behind and they cost money. Also at some point we hit the disk quota (because we don't cleanup).

## How Has This Been Tested?

Tested the commands locally.

## Types of changes

## Checklist:


jimmykarily:
Works but shows a warning: https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/smoke-tests-diego-master/builds/38

```error: Missing or incomplete configuration info.  Please point to an existing, complete config file:``` not sure why. This command should setup the kubeconfig: ```              gcloud auth activate-service-account --key-file $PWD/gke-key.json
```
and apparently it does (since the kubectl command worked).

PR #951, opened 2020-05-29
Revert "Add cats-internetless suite in experimental jobs"

viccuad:
Reverts cloudfoundry-incubator/kubecf#913

PR #913, opened 2020-05-22
Add cats-internetless suite in experimental jobs

jimmykarily:
## Description
Completes #144 

## Motivation and Context

This is one more test we need to run in our pipeline

## How Has This Been Tested?
TODO: Deploy the pipeline under a different name and let the job run.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code has security implications.
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #995, opened 2020-06-09
Cats internetless

jimmykarily:
## Description
Re-applying https://github.com/cloudfoundry-incubator/kubecf/pull/913 .
The reason [we reverted that PR](https://github.com/cloudfoundry-incubator/kubecf/pull/951/files) was that CATs became slow. We found that the catapult bump was causing it. We fixed catapult in the meantime so the PR should be ok to be tried again.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -816,7 +816,6 @@ jobs:
     passed:
     - {{ $previousTest | quote }}
     trigger: true
-    version: "every"
   - get: s3.kubecf-ci
     passed:
     - {{ $previousTest | quote }}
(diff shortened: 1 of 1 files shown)


jimmykarily:
This was tested here: https://concourse.suse.dev/teams/main/pipelines/dk-kubecf (a bit modified to run cats-internetless first). The internetless test passes and so does cats afterwards which means the cleanup code that brings internet access back also works.

.concourse/pipeline.yaml.gomplate
```diff
@@ -808,6 +809,108 @@ jobs:
...
 {{ $previousTest = (printf "cf-acceptance-tests-%s-%s" $cfScheduler $branch) }}
 
+- name: cats-internetless-{{ $cfScheduler }}-{{ $branch }}
+  public: true
+  plan:
+  - get: kubecf-{{ $branch }}
+    passed:
+    - {{ $previousTest | quote }}
+    trigger: true
+    version: "every"
```
viccuad, line 819: did we decide to remove all `version: every`, of was that for a different resource?

jimmykarily: it was a different one but makes sense to remove it here too

PR #998, opened 2020-06-09
Refactor pipelines and config files 

viccuad:
## Description

Refactor kubecf and reconciler pipeline template to accept explicit configs `kubecf.yaml` & `kubecf-pool-reconciler.yaml` files. 
Thanks to this, remove unneeded templating options, remove old EKCP documentation, and add 
a `trigger_publish: true` to the config.

## Motivation and Context

This way there's no way to mistakenly operate against a pipeline you don't intend to. One always needs to specify the name of the pipeline they are operating, and by copying the config `kubecf.yaml` to their own `my-kubecf.yaml` they don't operate against the production pipeline.

## How Has This Been Tested?

Tested by:
- Deploying my own test copy of the pipeline
- Running the script against production kubecf & kubecf-pool-reconciler pipelines, seeing no relevant changes.

## Types of changes

## Checklist:


viccuad:
It is missing the upgrade task, which is in `.concourse/tasks/*`. This needs to be templated too with the config `kubecf.yaml` as a datasource.

jimmykarily:
1. `trigger_publish` will still not work but we can fix it in a separate PR
2. With the pool resource now removed, do we need the reconciler pipeline around? If yes can we make it clear what it should be used for?

viccuad:
>     1. `trigger_publish` will still not work but we can fix it in a separate PR

`trigger_publish` option should work, but it's just minimal. My thought process was that it is needed for pipeline copies coming from SUSE contributors and running on concourse.suse.dev, as those would have the creds to succeed in publishing.

>     2. With the pool resource now removed, do we need the reconciler pipeline around? If yes can we make it clear what it should be used for?

We don't need it around, but I felt it doesn't hurt to have it for a bit. It's undocumented and out of the way.

jimmykarily:
True "trigger_publish" will work. I though it was the flag that disables all push actions (e.g. update GitHub statuses). That's the one that needs fixing (maybe in a different PR).

Having a non-documented pipeline around is what worries me. It will only confuse people. Especially when it's the same script that deploys that one and the kubecf one. Anyway, we can wait a bit until we decide if we stick to gke or not.

PR #992, opened 2020-06-05
feat: Add NOTES.txt to helm chart

viccuad:
## Description

Add a helm `templates/NOTES.txt` which gets shown on `helm install`.

The NOTES.txt points to https://kubecf.suse.dev/docs for documentation, which seems like a good starting point.

## Motivation and Context

Closes https://github.com/cloudfoundry-incubator/kubecf/issues/855.

## How Has This Been Tested?

Successfully tested by deploying the resulting chart and checking that the NOTES.txt are displayed correctly.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code has security implications.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #999, opened 2020-06-09
add notes.txt to release helmchart

dragonchaser:
This PR adds a `NOTES.txt`  to the templates folder of the helmchart. 
It is output as soon as the cluster is set up and proviedes useful information about the cluster.

## Checklist:

### DIFF
--- a/rules/kubecf/release_chart.sh
+++ b/rules/kubecf/release_chart.sh
@@ -25,6 +25,37 @@ tar xf "${KUBECF_CHART}"
   < "${KUBECF_IMAGE_LIST_JSON_FILE}" \
   > "${KUBECF_IMAGE_LIST_TXT_FILE}"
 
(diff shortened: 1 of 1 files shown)


andreas-kupries:
Closing this as a near duplicate of #992. Will copy the information about the eirini CSR over to that one.

andreas-kupries:
Additional comment: The eirini CSR templating was faulty, and according to @jandubois also irrrelevant.
Quoting From the discussion in :rocket: 

> Anyways, quarks-operator always wants to get all the permissions it can, so it there is no variable to explicitly allow it to approve CSRs.

> I think you can drop the whole block because you never have to approve certs manually with kubecf; they are always auto-approved because quarks has good connections to the authorities.

As such the copying of this part of the PR has been undone. And technically it makes this PR a full duplicate of #992.

Issue #949, opened 2020-05-28
CVE catchup: bump cflinuxfs3 to 0.189.0+


gaktive:
**Describe the bug**
Currently in kubecf, it looks like have cflinuxfs3 `0.167.0` (2020 March). We normally build on sle15 but we should ensure we keep up-to-date since CVEs do keep rolling in. Once I get a moment, I'll scan for any scary ones.

Latest cflinuxfs3: 0.189.0 that addresses other CVEs though I haven't vetted them yet: https://github.com/cloudfoundry/cflinuxfs3/releases/tag/0.189.0
scf has already done its own test with it based on older automation so we should be able to repurpose existing code to have something similar for kubecf: SUSE/scf#3472

gaktive:
Crosslinked for reference, the cf-deployment bump to `13.2` gets us to cflinuxfs3 `0.178.0` so we'd need to review higher than that once we do that bump.

thardeck:
The latest cflinuxfs3 is already automatically built and uploaded to Dockerhub in the flintstone Concourse: https://ci.flintstone.cf.cloud.ibm.com/teams/quarks/pipelines/release-images/jobs/build-cflinuxfs3/builds/76

But it uses always the latest stemcell which is not part yet of kubecf unless we merge the current stemcell pr.

@gaktive 
Should we use this image or create an additional pipeline to build it?
Or is it only about overriding the cflinuxfs3 version (from cf-deployment) for kubecf - but then we could have different stemcells in the release.

gaktive:
@thardeck upon discussion, the rule of thumb here is to look at the cflinuxfs3 versions as they become available and evaluate them based on CVEs. Whenever possible and avoiding breakage, we should bump cflinuxfs3 to the latest so long as we don't break cf-deployment.

Unfortunately, I don't have access to Flintstone to see the latest image but I can infer that we do build the latest. The proposal then is to use that image and put it into kubecf and ensure that tests pass. I'm not sure what the logic to use in case that there is breakage but when in doubt, use the cf-deployment bundled cflinuxfs3 instead (unless there's a downgrade).

Hope that helps.

gaktive:
@viovanov & @troytop any further insight?

troytop:
@gaktive I think that covers it.

thardeck:
@gaktive 
[...]

PR #1001, opened 2020-06-10
Override cflinuxfs3 to incorporate security fixes

thardeck:
## Description
Override cflinuxfs3 to incorporate security fixes - #949

## Motivation and Context
Our current cf-deployment does provide only an older cflinuxfs3 which is missing certain security fixes.
To get these fixes as soon as possible we are overriding the cflinuxfs3 version to the latest one.

## How Has This Been Tested?
Locally with catapult.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/values.yaml
+++ b/deploy/helm/kubecf/values.yaml
@@ -53,6 +53,11 @@ releases:
     stemcell:
       os: SLE_15_SP1
       version: 25.2-7.0.0_374.gb8e8e6af
+  cflinuxfs3:
+    version: 0.192.0
+    stemcell:
+      os: SLE_15_SP1
+      version: 26.3-7.0.0_374.gb8e8e6af
   # pxc is not a BOSH release.
   pxc:
     image:


deploy/helm/kubecf/values.yaml
```diff
@@ -53,6 +53,11 @@ releases:
     stemcell:
       os: SLE_15_SP1
       version: 25.2-7.0.0_374.gb8e8e6af
```
jimmykarily, line 55: If we are bumping the stemcell to be consistent in all images, then it should happen [here](https://github.com/cloudfoundry-incubator/kubecf/pull/1001/files#diff-8fc2de7ef5a3d25ddfce0cd023890d0cR36). Otherwise we will still use images with different stemcells so there is no benefit in bumping only some of them here.

thardeck: True, we should change the automation for it. Why are there so many stemcell overrides when all images are anyway rebuild against the latest stemcell?

Issue #772, opened 2020-04-24
TLS secrets for the embedded database should be plain k8s resources


f0rmiga:
**Describe the bug**

Currently, the TLS secrets are created via the BOSH variables, which doesn't make sense in our current setup that the database is a standalone PXC Helm chart.

**To Reproduce**

Go to the chart templates and assert that we don't create the TLS secrets there.

**Expected behavior**

The TLS secrets should be created together with the other database k8s resources.

viovanov:
These should be `QuarksSecrets`

andreas-kupries:
@f0rmiga @viovanov With regard to the TLS variables we want to be gone, are we talking about

```yaml
# Add certificates for PXC.
- type: replace
  path: /variables/-
  value:
    name: pxc_ca
    type: certificate
    options:
      common_name: pxc_ca
      is_ca: true
- type: replace
  path: /variables/-
  value:
    name: pxc_tls
    type: certificate
    options:
      ca: pxc_ca
      common_name: &pxc-cluster-address {{ printf "database.%s" .Release.Namespace }}
```

in file `kubecf/assets/operations/instance_groups/database.yaml` ?

Musings ...
- I notice that the above contains the anchor `&pxc-cluster-address`, and that further down in the same file this anchor is used. This seems to be fixable with a helper key in the ops file.
- Further however is the use of the BOSH variable `((pxc_tls.ca))` ... Ah, that likely maps to the value of `options.ca`, i.e. `pxc_ca`, and that is then used via anchor, so just another helper key.
- Hm. ... Ah, the eirini template has example of a `QS` for a certificate. And should be able to derive CA cert as well.

viovanov:
These secrets are not actually native.
I guess @f0rmiga thought they were because we use the `pxc_*` prefix everywhere.
But really, we just change the name of an existing variable: https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L505

We can use native->bosh entanglements for this.
https://quarks.suse.dev/docs/features/entanglements/#native-bosh

Please note that [this story](https://www.pivotaltracker.com/story/show/172993878) fixes some entanglement issues. This is likely blocked until that's fixed.

PR #1004, opened 2020-06-10
Switch embedded db from explicit bosh variables to implicit

andreas-kupries:
## Description

Added quarks secrets for the embedded database, matching the BOSH variables.
Dropped the explicit variable definition for same.
The implicit variable ref `((pxc_tls.ca))` should now be resolved from
the quarks secret.

## Motivation and Context

See #772 

## How Has This Been Tested?

This is currently failing.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/database.yaml
@@ -16,28 +16,11 @@
 
 {{- if .Values.features.embedded_database.enabled }}
 
-# Add certificates for PXC.
-- type: replace
-  path: /variables/-
-  value:
-    name: pxc_ca
-    type: certificate
-    options:
-      common_name: pxc_ca
-      is_ca: true
-- type: replace
-  path: /variables/-
-  value:
-    name: pxc_tls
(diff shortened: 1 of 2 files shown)


andreas-kupries:
I begin to strongly suspect that this setup requires explicit bosh variables for the proper sequencing of the secrets generation and use.

jandubois:
> It is my understanding that helm doesn't ensure any such thing. It rather evaluates the templates of the main chart and all subcharts together, as evidenced by this part of the documentation:

I just found that it is [explicitly documented](https://helm.sh/docs/topics/charts/#operational-aspects-of-using-dependencies):

>This is because when Helm installs/upgrades charts, the Kubernetes objects from the charts and all its dependencies are
>
> * aggregrated into a single set; then
> * sorted by type followed by name; and then
> * created/updated in that order.
>
> Hence a single release is created with all the objects for the chart and its dependencies.

jandubois:
Closing the PR because it is blocked and can't become unblocked in the foreseeable future.

It remains linked to issue #772, and the branch can be restored, and the PR re-opened via the Github UI, if this becomes desirable.

PR #1005, opened 2020-06-11
Store subcharts inside the kubecf repo

### LINKED ISSUE #987: Store subcharts inside the kubecf repo
Right now we reference the bundled subcharts via `requirements.yaml`:

```yaml
dependencies:
- name: eirini
  version: 1.0.2
  repository: http://opensource.suse.com/eirini-release/
...
```

and we only fetch the actual charts via `helm dep up` when we build the kubecf helm chart.

I would like to propose that we run `helm dep up` every time we bump the dependency and check the downloaded subcharts into the kubecf repo. This has the following advantages (in decreasing order of importance to me):

* It becomes possible to quickly look up something in the subchart without having to fetch it manually, e.g. to see how something is configurable.

* When a chart is bumped, you can see the diff of what has changed, both on Github and locally with git. It can provide context to understand why some subchart settings are being changed in kubecf.

* We can now guarantee that the subchart doesn't change between rebuilds. `requirements.yaml` only uses a version number, not a shasum, so currently we rely on charts not being re-published to the external repo without version number updates.

* The build becomes slightly faster, although only trivially so.

Are there any reasons why we should **not** do this?

jandubois:
This will make it possible to inspect changes during version bumps using regular git tools and also allows for text search against the charts.

Fixes #987


jandubois:
Please review the 2 commits separately; only https://github.com/cloudfoundry-incubator/kubecf/commit/fee86a4 contains manual changes, and except for deletion of a single line of code and the removal of `.gitignore` is solely documentation update.

The second commit then executes the commands documented in the first commit to add the expanded charts to this repo.

The most important part of the review would be to double-check that these commands does cover all possible scenarios in the future: add/removed/updated subcharts, add/removed/changed files belonging to one of the subcharts.

Tagging @mook-as for review because he has commented on #987. "No good deed shall go unpunished!"

doc/Contribute.md
```diff
@@ -100,6 +101,27 @@ associated documentation, if we have any.
...
+
+## Updating subcharts
+
+The kubecf helm chart includes a number of subcharts. They are declared in
+[requirements.yaml](../deploy/helm/kubecf/requirements.yaml). For the
+convenience of development they are included in unpacked form directly in
+this repo, so version changes can be inspected with regular `git` tools,
+and the subcharts can be searched with `grep` etc.
+
+The procedure to update the version of a subchart is:
```
mook-as, line 113: Can we not have a `bazel` target for this to ensure we don't screw this up when we copy & paste? (Yes, we want to get rid of bazel, but until the replacement _starts_ to exist it's still our build system.)

doc/Contribute.md
```diff
@@ -100,6 +101,27 @@ associated documentation, if we have any.
...
+
+```
+cd deploy/helm/kubecf
+vi requirements.yaml
+rm -rf charts/*
+helm dep up
+for CHART in charts/*tgz; do tar xfz $CHART -C charts; rm $CHART; done
```
mook-as, line 120: Of course my first thought is "what if the file name has a space in it".  `C:\Program Files` has traumatized me forever.

doc/Contribute.md
jandubois: I agree that we should, but personally I'm unable to create a `bazel` target. I can create a shell script that you can run from the `bazel` target, but that is practically already part of the docs I wrote.

So, we can:

1. Have somebody else contribute the `bazel` target for this PR, or
2. Delay merging this PR until we have switched build systems, and then add the shell script, or
3. Be careful for the next month or two to make sure we follow the docs when bumping subcharts.

Personally I find (3) acceptable, given that all PRs go through code review, but I'm fine with the other options too.

The advantage of merging this earlier is that we will already have diffs available for any subchart bumps we do before switching build systems.

doc/Contribute.md
jandubois: I doubt this will happen, but agree that the code should be robust enough to handle that, so will make a change.

jandubois:
> * so this should work.

I've tested that this change works before making the PR. 😄 

And for the test it deployed with Eirini, so I was worried that using expanded charts were somehow incompatible with conditions in `requirements.yaml`. But then I realized that it was a proper Eirini deployment, with no `diego-cell` pods, and eventually found the hidden config file in my tree that had `features.eirini.enabled: true`.

doc/Contribute.md
jandubois: Fixed in a262c5e

doc/Contribute.md
mook-as: I think we _do_ want at least a shell script; even if we don't run it from bazel, we want to make sure we don't make dumb typos / copy & paste errors (I always assume I can find ways to mess up).

We can tack on the bazel target later.

jandubois: Script added in 81541c2

dev/linters/shellcheck.sh
```diff
@@ -6,4 +6,7 @@ workspace=$(bazel info workspace)
 
 # shellcheck disable=SC2046
 # We want word splitting with find.
-bazel run @shellcheck//:binary -- $(find "${workspace}" -name '*.sh')
+bazel run @shellcheck//:binary -- \
+      $(find "${workspace}" \
+             -not \( -path "${workspace}/deploy/helm/kubecf/charts" -prune \) \
+             -name '*.sh')
```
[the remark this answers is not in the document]
jandubois: Sure, we could do that.

Or should we store the list of files in a variable `shell_scripts=$(find ...)`? Is there a size limit on variables?

Either way, do you want this as part of this PR, or should it just go into the bazel replacement?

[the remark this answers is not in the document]
mook-as: I originally was thinking `shell_scripts=( $(find…) )` and then realized `shellcheck` will want `mapfile -t shell_scripts < <(find …)` and then it'll be long again and whatnot, but I'm perfectly happy with that as long as it works.

I think I prefer that as part of this PR, just to ensure things are readable.

dev/linters/yamllint.sh
```diff
@@ -6,9 +6,13 @@ workspace=$(bazel info workspace)
 
 find_args=(
     -not \( -path "${workspace}/deploy/helm/kubecf/charts" -prune \)
-    -path "${workspace}/deploy/helm/kubecf/values.*"
-    -or -not -path "${workspace}/deploy/helm/kubecf/*"
-    -name '*.yaml' -or -name '*.yml'
+    -and
+    \( -path "${workspace}/deploy/helm/kubecf/values.*"
+       -or
+       -not \( -path "${workspace}/deploy/helm/kubecf" -prune \)
```
mook-as, line 12: If we pruned `"${workspace}/deploy/helm/kubecf"` how will we ever find `"${workspace}/deploy/helm/kubecf/values.*"`? Isn't that redundant? And wouldn't that no longer find `…/deploy/helm/kubecf/assets/operations/*`?

Note that the old expression did _not_ prune anything.

jandubois: The expression `-path "./deploy/helm/kubecf/values.*" -or -not -path "./deploy/helm/kubecf/*"` does effectively prune all those files too; it just works a bit harder by traversing the whole tree. But the right side of the `-or` will reject all of them.

```
$ find . -path "./deploy/helm/kubecf/values.*" -or -not -path "./deploy/helm/kubecf/*" -name '*.yaml' | grep helm
./deploy/helm/kubecf/values.yaml
```

And this makes sense because the files still have unresolved templating code in them, which makes them invalid YAML. You could only lint the chart if you push it through `helm template` first.

But you are right that the pruning will remove the `values.yaml` from the list, so I have to revert that part.

dev/linters/helmlint.sh
```diff
@@ -2,7 +2,7 @@
 
 set -o errexit -o nounset
 
-if grep --exclude="helmlint.sh" -r "{{-.*-}}" .; then
+if git grep --quiet '{{-.*-}}' -- . ':!/deploy/helm/kubecf/charts' ':!/dev/linters/helmlint.sh'; then
```
mook-as, line 5: Could we either user `git -c core.pager=cat` or `PAGER=cat` or something instead? Not being able to see _where_ the issue is will be very annoying.

Issue #736, opened 2020-04-22
Detailed documentation for each value in values.yaml


viovanov:
**Is your feature request related to a problem? Please describe.**
Lack of documentation in `values.yaml`

**Describe the solution you'd like**
Just like in SCF, each value has a comment.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

mook-as:
https://github.com/cloudfoundry-incubator/kubecf/pull/1056 added to start on the documentation (and to ensure we don't add any new ones lacking comments).  Will still leave this open though, as that doesn't include all the sub-chart knobs.

mook-as:
Closing this as good enough: if we want more, we could consider writing a schema for the file instead.

PR #1015, opened 2020-06-12
Cleanup pods after tests

jimmykarily:
## Description

When you delete a job with `kubectl` the created pods get deleted too. But when quarks-job deletes a job it doesn't delete the pod unless the `delete: pod` label is set on the job.

See also:
- https://github.com/cloudfoundry-incubator/quarks-job/blob/master/docs/quarksjob.md#job-controller
- https://github.com/cloudfoundry-incubator/quarks-job/blob/b16b2a026f867b8e4c8acc070735864dc5c3f6d8/pkg/kube/controllers/quarksjob/job_reconciler.go#L96

For the kubecf pipeline, we will cleanup the pod manually in catapult because we need the pod to be around for a while to be able to write the logs on a file. If quarks job deletes it, then we may not get the chance to do that (https://github.com/SUSE/catapult/pull/232).

I opened this PR in case it is useful for some other pipeline or use case. I know people had issues with testing pods that were left behind. The default is still to keep the pod around so this shouldn't break anything.

## How Has This Been Tested?

Run smoke tests with this setting on and pod was deleted after completion.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
@@ -40,4 +40,9 @@
       serial: true
     vm_type: minimal
 
+
+- type: replace
+  path: /instance_groups/name=acceptance-tests/env?/bosh/agent/settings/labels/delete?
+  value: "pod"
+
 {{- end }}
--- a/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
@@ -41,8 +41,11 @@
     vm_type: minimal
 
 
+{{- if .Values.testing.cf_acceptance_tests.delete_pod }}
 - type: replace
   path: /instance_groups/name=acceptance-tests/env?/bosh/agent/settings/labels/delete?
   value: "pod"
+{{- end }}
+
 
 {{- end }}
--- a/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
@@ -43,7 +43,7 @@
 
 {{- if .Values.testing.cf_acceptance_tests.delete_pod }}
 - type: replace
-  path: /instance_groups/name=acceptance-tests/env?/bosh/agent/settings/labels/delete?
+  path: /instance_groups/name=acceptance-tests/env?/bosh/agent/settings/labels/delete
   value: "pod"
 {{- end }}
 
--- a/deploy/helm/kubecf/assets/operations/instance_groups/brain-tests.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/brain-tests.yaml
@@ -56,8 +56,10 @@
       serial: true
     vm_type: minimal
 
+{{- if .Values.testing.brain_tests.delete_pod }}
 - type: replace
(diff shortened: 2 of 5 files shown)


jimmykarily:
Not sure how useful this feature is @bikramnehra. Feel free to close this if it's not. I don't know how much of a problem this is for pipelines other than the kubecf one.

deploy/helm/kubecf/assets/operations/instance_groups/acceptance-tests.yaml
```diff
@@ -40,4 +40,12 @@
       serial: true
     vm_type: minimal
 
+
+{{- if .Values.testing.cf_acceptance_tests.delete_pod }}
+- type: replace
+  path: /instance_groups/name=acceptance-tests/env?/bosh/agent/settings/labels/delete?
```
mook-as, line 46:
```suggestion
  path: /instance_groups/name=acceptance-tests/env?/bosh/agent/settings/labels/delete
```
The `?` applies to the key and any subsequent keys; no need to repeat it.

deploy/helm/kubecf/assets/operations/instance_groups/brain-tests.yaml
```diff
@@ -56,4 +56,10 @@
       serial: true
     vm_type: minimal
 
+{{- if .Values.testing.brain_tests.delete_pod }}
+- type: replace
+  path: /instance_groups/name=brain-tests/env?/bosh/agent/settings/labels/delete?
```
mook-as, line 61:
```suggestion
  path: /instance_groups/name=brain-tests/env?/bosh/agent/settings/labels/delete
```

deploy/helm/kubecf/assets/operations/instance_groups/smoke-tests.yaml
```diff
@@ -1,5 +1,11 @@
 {{- if .Values.testing.smoke_tests.enabled }}
 
+{{- if .Values.testing.smoke_tests.delete_pod }}
+- type: replace
+  path: /instance_groups/name=smoke-tests/env?/bosh/agent/settings/labels/delete?
```
mook-as, line 5:
```suggestion
  path: /instance_groups/name=smoke-tests/env?/bosh/agent/settings/labels/delete
```

deploy/helm/kubecf/assets/operations/instance_groups/sync-integration-tests.yaml
```diff
@@ -49,4 +49,10 @@
       serial: true
     vm_type: minimal
 
+{{- if .Values.testing.sync_integration_tests.delete_pod }}
+- type: replace
+  path: /instance_groups/name=sync-integration-tests/env?/bosh/agent/settings/labels/delete?
```
mook-as, line 54:
```suggestion
  path: /instance_groups/name=sync-integration-tests/env?/bosh/agent/settings/labels/delete
```

deploy/helm/kubecf/values.yaml
```diff
@@ -331,12 +331,16 @@ features:
 testing:
   brain_tests:
     enabled: false
+    delete_pod: false
```
mook-as, line 334: If possible, please add documentation for this (see #736).

Why do we want this to be false by default? Not that I have a preference either way; I was just curious as to the choice.

jimmykarily: I made it false to avoid breaking any existing pipelines. Right now the pod is not deleted. Deleting it after completion could break a pipeline that expect the pod to be around in order to fetch the logs (e.g. the kubecf one). This also addresses @bikramnehra 's comment above.

PR #1028, opened 2020-06-17
Pipeline: try to cf login right after deployment

viccuad:
## Description

Try to cf login right after deployment.

## Motivation and Context

This catches glaring errors in deployment.

## How Has This Been Tested?

This catapult target is already being used in SUSE/cap.

## Types of changes

## Checklist:


viccuad:
Fixed order of login, needs to happen after setting DNS.

viccuad:
Closing, DNS entries are private and can only be accessed from inside the cluster. Which means that running `kubecf-login` from the Concourse worker to do `cf-login` will fail, as the Concourse cluster cannot access the internal k8s cluster.

PR #1031, opened 2020-06-17
Remove unneeded vars in the upgrade task of the kubecf pipeline

viccuad:
## Description
Remove unneeded vars in the upgrade pipeline task, use defaults already present.
Set Catapult vars to not download and install gcloud inline and provide plaintext input.
Perform `cf login` after `helm install` and afer `helm upgrade`, as sanity check.

## Motivation and Context
Speeds up the test because gcloud cli is not downloaded and installed, makes clearer what is happening.

## How Has This Been Tested?
Not tested. Because of line https://github.com/cloudfoundry-incubator/kubecf/blob/viccuad-upgrade-clobber/.concourse/pipeline.yaml.gomplate#L1401, one needs to deploy a pipeline that targets this `viccuad-upgrade-clobber` branch, and that is blocked by https://github.com/cloudfoundry-incubator/kubecf/issues/1032.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/tasks/upgrade.sh
+++ b/.concourse/tasks/upgrade.sh
@@ -40,6 +40,7 @@ export GKE_CLUSTER_ZONE="$GKE_ZONE"
 KUBECF_LATEST_RELEASE="$(cat kubecf-github-release/version)"
 export KUBECF_LATEST_RELEASE
 export SCF_CHART="https://github.com/cloudfoundry-incubator/kubecf/releases/download/v${KUBECF_LATEST_RELEASE}/kubecf-bundle-v${KUBECF_LATEST_RELEASE}.tgz"
+export ENABLE_EIRINI=false
 
 export BACKEND=gke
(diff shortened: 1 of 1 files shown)


.concourse/tasks/upgrade.sh
```diff
@@ -42,13 +41,9 @@ KUBECF_LATEST_RELEASE="$(cat kubecf-github-release/version)"
 export KUBECF_LATEST_RELEASE
 export SCF_CHART="https://github.com/cloudfoundry-incubator/kubecf/releases/download/v${KUBECF_LATEST_RELEASE}/kubecf-bundle-v${KUBECF_LATEST_RELEASE}.tgz"
 
-export ENABLE_EIRINI=false
```
jimmykarily, line 45: I think being explicit is good in this case (so people don't wonder what the default is)

viccuad: Fixed

Issue #1036, opened 2020-06-18
 Have documentation on "How do I deploy a copy of the pipeline?"


viovanov:
**Is your feature request related to a problem? Please describe.**
It's hard to deploy a copy of the pipeline without knowing all the settings and how they affect "production" artifacts.

**Describe the solution you'd like**
A clear and concise description of how to deploy the pipeline without adversely affecting the production pipeline.

**Additional context**
#1032

PR #1041, opened 2020-06-19
Expand pipeline docs for copies of the pipeline

viccuad:
## Description

- Expand docs for deploying copies of the kubecf pipeline.
- Explain how to target a specific branch on a pipeline copy.
- Add 2 new s3 buckets (`kubecf-ci-draft`, `kubecf-draft`) for use on pipeline copies. These buckets are copies of the existing ones (public, etc), yet s3 will only keep 1 historical version of each file.

## Motivation and Context

Closes https://github.com/cloudfoundry-incubator/kubecf/issues/1036.

## How Has This Been Tested?

Deploy a pipeline copy, following the instructions https://concourse.suse.dev/teams/main/pipelines/kubecf-viccuad.
It successfully uses the `-draft` buckets: https://concourse.suse.dev/teams/main/pipelines/kubecf-viccuad/jobs/build-master/builds/4.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/kubecf.yaml
+++ b/.concourse/kubecf.yaml
@@ -1,16 +1,12 @@
 ---
 
+# NOTE use your own bucket when developing
 s3_bucket: kubecf-ci
 s3_bucket_region: eu-central-1
-# NOTE use the following if developing
-# s3_bucket: kubecf-ci-draft
-# s3_bucket_region: eu-central-1
 
+# NOTE use your own bucket when developing
 s3_final_bucket: kubecf
 s3_final_bucket_region: us-west-2
-# NOTE use the following if developing
-# s3_final_bucket: kubecf-draft
-# s3_final_bucket_region: us-west-2
 
 kubecf_repository: cloudfoundry-incubator/kubecf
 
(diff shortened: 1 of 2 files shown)


.concourse/kubecf.yaml
```diff
@@ -2,8 +2,16 @@
 
 s3_bucket: kubecf-ci
 s3_bucket_region: eu-central-1
+# NOTE use the following if developing
```
jimmykarily, line 5: If many people use the same buckets then they will end up messing each other's clusters/artifacts. I would prefer if anyone deploying a copy of the pipeline, used their own bucket. Buckets are easy to create and cheap.

viccuad: Done.
I would be happier appending commit hashes to intermediate artifacts. That feels saner, even if it's not bulletproof, since the same artifact can be built differently as pipeline definitions change.

jimmykarily: we do have commit hashes in the artifact names I think. As you said, 2 different pipeline might build the same commit but using different pipeline configuration, thus one overwriting the other pipelines artifacts. Separate buckets are safer imho.

Issue #1032, opened 2020-06-17
Kubecf pipeline should allow for development copies of itself


viccuad:
**Is your feature request related to a problem? Please describe.**

Right now it's not possible to deploy a copy of the production pipeline that doesn't interfere with the production one. Also, by disabling some jobs on the copy of the pipeline, it is possible that the cleanup of GKE cluster doesn't get triggered.

Copies of the pipeline are needed, for example if you want to improve the KubeCF production pipeline without testing in production, or want to deploy your own that tests a specific branch. 

**Describe the solution you'd like**

The KubeCF pipeline should:

- [x] Have options to disable publishing artifacts to Github or the release S3 bucket, and to update Github status. (https://github.com/cloudfoundry-incubator/kubecf/pull/1041)
- [x] Work if the list of branches to be tested  is changed (see `$branches` definition in pipeline.yaml.gomplate).
    - Right now, selecting a somewhat normal branch name such as `viccuad-upgrade-clobber`   breaks GKE cluster creation or deletion because the cluster name is too long. (https://github.com/cloudfoundry-incubator/kubecf/pull/1108)
- [ ]  ~Allow partial deployments of the kubecf pipeline, making sure that cluster deletion gets triggered correctly to not forget destroying clusters.~
- [ ] Allow pausing the first triggering job by default, to not waste resources (https://github.com/cloudfoundry-incubator/kubecf/pull/1139)
- [x] Consume and publish intermediate build artifacts to a different s3 bucket than the production one
- [x] #1036 Have documentation on "How do I deploy a copy of the pipeline?"

**Describe alternatives you've considered**
Make the needed changes (chop code, rename vars) to the pipeline definition by hand, which is impractical.

**Additional context**

This arose just in a try to test https://github.com/cloudfoundry-incubator/kubecf/pull/1031

viovanov:
Closing because of #1144

PR #1043, opened 2020-06-19
Pipeline: Fix upgrade test by running smoke-tests instead of trying to cf login

viccuad:
## Description

* Run smoke-tests as sanity check instead of trying to cf login. `cf login` happens outside of the GKE cluster and from the concourse worker; and the DNS zone is set to private (only accessible from inside the cluster).
* Update the GKE code in the upgrade job to match the rest of the pipeline.

## Motivation and Context

Make upgrade job green.

## How Has This Been Tested?

Not tested. Because of line https://github.com/cloudfoundry-incubator/kubecf/blob/viccuad-upgrade-clobber/.concourse/pipeline.yaml.gomplate#L1401, one needs to deploy a pipeline that targets this viccuad-upgrade-clobber branch, and that is blocked by #1032.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/tasks/upgrade.sh
+++ b/.concourse/tasks/upgrade.sh
@@ -104,17 +104,17 @@ make kubeconfig
 make kubecf kubecf-login
 
 # Setup dns
-tcp_router_ip=$(kubectl  get svc -n scf tcp-router-public -o json | jq -r .status.loadBalancer.ingress[].ip | head -n 1)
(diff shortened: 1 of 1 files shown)


viccuad:
Closing without merging.

In the beginning I tried to do cf login after upgrade, but it's not possible with private DNS. Then tried to run smoke-tests, but we are not waiting for the private DNS entries to be up, and a wait on CI from the concourse worker is not going to work.

PR #1049, opened 2020-06-19

cfcontainerizationbot:

### DIFF
--- a/doc/dev/general.md
+++ b/doc/dev/general.md
@@ -54,8 +54,8 @@ with Helm Tiller pre-installed.
 ```shell
 helm install --name cf-operator \
      --namespace kubecf \
-     --set "global.operator.watchNamespace=kubecf" \
-     https://s3.amazonaws.com/cf-operators/helm-charts/cf-operator-v0.4.2-147.gb88e4296.tgz
+     --set "global.singleNamespace.name=kubecf" \
+     https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-5.0.0%2B0.gd7ac12bc.tgz
 ```
 
 In the example above, version 0.4.1 of the operator was used. Look
--- a/doc/dev/general.md
+++ b/doc/dev/general.md
@@ -58,7 +58,7 @@ helm install --name cf-operator \
      https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-5.0.0%2B0.gd7ac12bc.tgz
 ```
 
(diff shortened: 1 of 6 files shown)


mook-as:
Note that we'll need catapult to be updated (https://github.com/SUSE/catapult/pull/240) to be able to deploy this in CI.

jandubois:
> Note that we'll need catapult to be updated ([SUSE/catapult#240](https://github.com/SUSE/catapult/pull/240)) to be able to deploy this in CI.

It is my understanding that @troytop and @manno agreed to update the operator to make the settings backwards compatible again to make upgrades easier for users (old option will be added back, will be the default, and may even default the watch namespace to `kubecf`).

I don't know the timeline for this, but maybe we should not bump the operator until that change has landed?

manno:
@jandubois we added a story to add support for the old argument and add a deprecation notice (https://www.pivotaltracker.com/story/show/173486608) when using it.

Beyond that, I'd suggest not to use the arg (or the new singleNamespace one) in demos and documentation. Since the namespace is created by helm, it's my understanding that helm will also delete it after uninstalling the operator. So there is little value in users choosing a name for that namespace.

I find the 'kubecf' default a bit too specific, but it's used it in kubecf documentation, I guess?

doc/dev/general.md
```diff
@@ -54,8 +54,8 @@ with Helm Tiller pre-installed.
 ```shell
 helm install --name cf-operator \
      --namespace kubecf \
-     --set "global.operator.watchNamespace=kubecf" \
-     https://s3.amazonaws.com/cf-operators/helm-charts/cf-operator-v0.4.2-147.gb88e4296.tgz
+     --set "global.singleNamespace.name=kubecf" \
+     https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-5.0.0%2B0.gd7ac12bc.tgz
 ```
 
 In the example above, version 0.4.1 of the operator was used. Look
```
manno, line 61: The version number is outdated

troytop:
> I find the 'kubecf' default a bit too specific, but it's used it in kubecf documentation, I guess?

I get that we want to make Quarks useful for _any_ BOSH release (and indeed other Kubernetes applications), but KubeCF is the only "customer" right now. If this default becomes a problem (and it will be an great problem to have) we can change the default.

doc/dev/general.md
mook-as: Nice catch! Fixed in 0227a410

PR #1074, opened 2020-06-25
ADR for upgrade testing

viovanov:
## Description
Architecture Decision Record for testing upgrades.

## Motivation and Context
We have to ensure that users can upgrade KubeCF deployments.

### DIFF
--- a/doc/architecture/decisions/0004-upgrade-testing.md
+++ b/doc/architecture/decisions/0004-upgrade-testing.md
@@ -1,0 +1,40 @@
+# 3. upgrade testing methodology when releasing
+
+Date: 2020-06-25
+
+## Status
+
+Accepted
+
+## Context
+
+One of the goals of the KubeCF project is to offer a smooth experience when it comes to upgrading Cloud Foundry deployments.
+That means any user operating a KubeCF deployment should:
+
+- be able to easily identify which versions he can upgrade to
+- upgrade using documented procedures
+- have confidence that the exact version upgrade he's doing has been tested by the KubeCF team 
+
+## Decision
+
+- not allowed to have minor releases for version X.y.z if X+1 has been released (only patch releases)
+- for a version X.y.z, the upgrade path must be tested for:
+  - the previous X.*.* release
+  - version max(X-1.*.*)
+
+For downstream projects, assuming they don't release as often as KubeCF, the following rules should be in place:
+
+- there should be a release for each major KubeCF release
+- for a version X.y.z, the upgrade path must be tested for:
+  - the previous X.*.* release
+- if a version X.y.z cannot be upgraded from X.y-1.t, a new version X.y-1.t+1 must be created from a known upgradeable path in KubeCF.
+
+![diagram](https://docs.google.com/drawings/d/e/2PACX-1vSK_9XqNiLbpzrQGnJ9BISSQq8DKTeE3yDjszyJfC7BdPuABO0QbAMMZruEoMnTFwhhtzCEGeXowqmh/pub?w=1037&h=918)
+
+## Consequences
+
+- because we only allow patch releases for old major releases
+- more scenarios must be tested in our pipelines to ensure the upgrade paths work.
+- accurately using semver for KubeCF is important
+- should enforce [this ADR](https://github.com/cloudfoundry-incubator/kubecf/blob/master/doc/architecture/decisions/0002-git-commit-messages.md)
+
--- a/doc/architecture/decisions/0004-upgrade-testing.md
+++ b/doc/architecture/decisions/0004-upgrade-testing.md
@@ -17,7 +17,7 @@ That means any user operating a KubeCF deployment should:
 
 ## Decision
 
-- not allowed to have minor releases for version `X.y.z` if `X+1` has been released (only patch releases)
+- KubeCF is not allowed to have minor releases for version `X.y.z` if `X+1` has been released (only patch releases)
 - for a version `X.y.z`, the upgrade path must be tested for:
   - the previous `X.*.*` release
   - the latest `X-1.*.*` release
--- a/doc/architecture/decisions/0004-upgrade-testing.md
+++ b/doc/architecture/decisions/0004-upgrade-testing.md
@@ -17,17 +17,17 @@ That means any user operating a KubeCF deployment should:
 
 ## Decision
 
-- not allowed to have minor releases for version X.y.z if X+1 has been released (only patch releases)
-- for a version X.y.z, the upgrade path must be tested for:
-  - the previous X.*.* release
-  - version max(X-1.*.*)
+- not allowed to have minor releases for version `X.y.z` if `X+1` has been released (only patch releases)
+- for a version `X.y.z`, the upgrade path must be tested for:
+  - the previous `X.*.*` release
+  - version max(`X-1.*.*`)
 
 For downstream projects, assuming they don't release as often as KubeCF, the following rules should be in place:
 
 - there should be a release for each major KubeCF release
-- for a version X.y.z, the upgrade path must be tested for:
-  - the previous X.*.* release
-- if a version X.y.z cannot be upgraded from X.y-1.t, a new version X.y-1.t+1 must be created from a known upgradeable path in KubeCF.
+- for a version `X.y.z`, the upgrade path must be tested for:
+  - the previous `X.*.*` release
+- if a version `X.y.z` cannot be upgraded from `X.y-1.t`, a new version `X.y-1.t+1` must be created from a known upgradeable path in KubeCF.
 
 ![diagram](https://docs.google.com/drawings/d/e/2PACX-1vSK_9XqNiLbpzrQGnJ9BISSQq8DKTeE3yDjszyJfC7BdPuABO0QbAMMZruEoMnTFwhhtzCEGeXowqmh/pub?w=1037&h=918)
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,40 @@
...
+- be able to easily identify which versions he can upgrade to
+- upgrade using documented procedures
+- have confidence that the exact version upgrade he's doing has been tested by the KubeCF team 
+
+## Decision
+
+- not allowed to have minor releases for version X.y.z if X+1 has been released (only patch releases)
+- for a version X.y.z, the upgrade path must be tested for:
+  - the previous X.*.* release
+  - version max(X-1.*.*)
```
viccuad, line 23: We explicitly don't support upgrading from `1.0` to `2.0`, I think it is worth to note it in this document.

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,40 @@
...
+
+For downstream projects, assuming they don't release as often as KubeCF, the following rules should be in place:
+
+- there should be a release for each major KubeCF release
+- for a version X.y.z, the upgrade path must be tested for:
+  - the previous X.*.* release
+- if a version X.y.z cannot be upgraded from X.y-1.t, a new version X.y-1.t+1 must be created from a known upgradeable path in KubeCF.
+
+![diagram](https://docs.google.com/drawings/d/e/2PACX-1vSK_9XqNiLbpzrQGnJ9BISSQq8DKTeE3yDjszyJfC7BdPuABO0QbAMMZruEoMnTFwhhtzCEGeXowqmh/pub?w=1037&h=918)
```
viccuad, line 32: Nitpicking:
We are going to remove the 'v' out of our version strings, to comply with semver. May we worth it to remove it from the diagram too. Aso, maybe it is good to have the diagram as an asset file of the repo.

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Consequences
+
+- because we only allow patch releases for old major releases
```
viccuad, line 36: I don't understand the phrase as it is, maybe it was intended to be as follows?

```suggestion
- we only allow patch releases for old major releases
```

doc/architecture/decisions/0004-upgrade-testing.md
mook-as: I don't understand this graph:
- What's all the coloured boxes?
- Why is there a disconnected graph on the right?
- Why is v3.2.1 not upgradable from v2.2.0?

doc/architecture/decisions/0004-upgrade-testing.md
viccuad: Maybe with a wording such as "Starting from X.Y" could be enough.

doc/architecture/decisions/0004-upgrade-testing.md
andreas-kupries: There is also the direction of the arrows. Right now they seem to point from version to predecessor to upgrade from ?
With upgrades I would have expected arrows from version to descendant to upgrade to.

doc/architecture/decisions/0004-upgrade-testing.md
viovanov: Can you please give an example? If `1.0` the the max for the `1.*` series, then `2.0` should upgrade from `1.0`.

doc/architecture/decisions/0004-upgrade-testing.md
viovanov: Removed the v, added titles to the 2 graphs. Hopefully the colors make sense now @mook-as 
I changed the direction of the arrows too, I agree it's more intuitive this way @andreas-kupries

mook-as: I _think_ the colours basically mean "downstream 2.0.0 packages KubeCF 1.0.0", etc? If that's the case, it would be much better to have explicit text saying so (a dashed arrow pointing from KubeCF 1.0.0 to 2.0.0 with a caption, perhaps).

It might also be a good idea to embed text-representation of the graph, instead of linking to a Google doc that not everybody can edit (esp. for forks that want to PR changes).  Perhaps something like [this](https://github.com/TLmaK0/gravizo) would help.

doc/architecture/decisions/0004-upgrade-testing.md
viccuad: But on the `2.0` release notes we have:
> 

    WARNING: this major version introduces some changes required to consume Eirini as a native component
    by forcing the usage of kubecf name throughout the deployment resources (pod, namespace, ...)
    and so, upgrading from 1.x to 2.x will not be possible!
    More technical details can be found on #477

Hence, saying "starting from 2.0". Not that it matters much, though.

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,39 @@
...
+- be able to easily identify which versions he can upgrade to
+- upgrade using documented procedures
+- have confidence that the exact version upgrade he's doing has been tested by the KubeCF team 
+
+## Decision
+
+- not allowed to have minor releases for version `X.y.z` if `X+1` has been released (only patch releases)
+- for a version `X.y.z`, the upgrade path must be tested for:
+  - the previous `X.*.*` release
+  - version max(`X-1.*.*`)
```
viccuad, line 23: Nitpicking:
```suggestion
  - the latest `X-1.*.*`
```

doc/architecture/decisions/0004-upgrade-testing.md
viccuad: I agree with the caption, I end up making my own assumptions here when reading the chart.

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,39 @@
...
+## Status
+
+Accepted
+
+## Context
+
+One of the goals of the KubeCF project is to offer a smooth experience when it comes to upgrading Cloud Foundry deployments.
+That means any user operating a KubeCF deployment should:
+
+- be able to easily identify which versions he can upgrade to
```
viccuad, line 14:
```suggestion
- be able to easily identify which versions they can upgrade to
```

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,39 @@
...
+- upgrade using documented procedures
+- have confidence that the exact version upgrade he's doing has been tested by the KubeCF team
```
viccuad, line 16:
```suggestion
- have confidence that the exact version upgrade they are performing has been tested by the KubeCF team 
```

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,39 @@
...
+One of the goals of the KubeCF project is to offer a smooth experience when it comes to upgrading Cloud Foundry deployments.
+That means any user operating a KubeCF deployment should:
+
+- be able to easily identify which versions he can upgrade to
+- upgrade using documented procedures
+- have confidence that the exact version upgrade he's doing has been tested by the KubeCF team 
+
+## Decision
+
+- not allowed to have minor releases for version `X.y.z` if `X+1` has been released (only patch releases)
```
viccuad, line 20:
```suggestion
- KubeCF is not allowed to have minor releases for version `X.y.z` if `X+1` has been released (only patch releases)
```

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,39 @@
...
+- for a version `X.y.z`, the upgrade path must be tested for:
+  - the previous `X.*.*` release
+  - version max(`X-1.*.*`)
+
+For downstream projects, assuming they don't release as often as KubeCF, the following rules should be in place:
+
+- there should be a release for each major KubeCF release
```
viccuad, line 27: Nitpicking:

```suggestion
- there should be a release for each major KubeCF release (as upgrades skipping a major KubeCF release are not supported)
```

doc/architecture/decisions/0004-upgrade-testing.md
```diff
@@ -0,0 +1,39 @@
...
+- have confidence that the exact version upgrade he's doing has been tested by the KubeCF team 
+
+## Decision
+
+- not allowed to have minor releases for version `X.y.z` if `X+1` has been released (only patch releases)
+- for a version `X.y.z`, the upgrade path must be tested for:
+  - the previous `X.*.*` release
+  - version max(`X-1.*.*`)
+
+For downstream projects, assuming they don't release as often as KubeCF, the following rules should be in place:
```
viccuad, line 25: Nitpicking:
```suggestion
Therefore, for downstream projects, assuming they don't release as often as KubeCF, we recommend that the following rules should be in place:
```

As downstream actors are free to consume KubeCF as they see fit.

doc/architecture/decisions/0004-upgrade-testing.md
viovanov: Correct, this ADR can't be applied retroactively.

Issue #1095, opened 2020-07-01
Ensure we can deploy working pipelines from the code of any branch


viccuad:
**Is your feature request related to a problem? Please describe.**

Right now, we deploy pipeline definitions always from master.

Problem:
To test a specific branch (`release-2.2`, `release-2.3`), we need to deploy a pipeline from the CI code of said branch, and not from master. Since:
- master may not use Bazel in the pipelines but Bash to lint/build (yet the branches only have Bazel)
- production and experimental jobs may differ from master
- We need to only test those PRs that target the specific branch (eg, `release-2.3`), instead of re-testing all PRs to the repo.
- ...

(corollary: CI is tightly-coupled with the code it is testing)

**Describe the solution you'd like**

- Contribute to master all needed bits to be able to deploy a pipeline not tied to master.
  - [x] Add config options to specify which branch(s) does the pipeline monitor, list of prod/experimental jobs (https://github.com/cloudfoundry-incubator/kubecf/pull/1059)
  - [x] add configuration on which base_branch to monitor on PRs (https://github.com/cloudfoundry-incubator/kubecf/pull/1098)
- - [x] Remove release-2.X tabs from the kubecf pipeline. Reconfigure and redeploy kubecf master pipeline, to test only PRs against `master`, and not duplicate test PRs against release branches (because those tests will fail anyways!) ( https://github.com/cloudfoundry-incubator/kubecf/pull/1105)
- - [ ] Ensure resources between pipeline copies don't collide (https://github.com/cloudfoundry-incubator/kubecf/pull/1108)
- - [x] Backport all straight fixes to the pipeline from master and deploy corresponding pipeline for:
  - [x] `release-2.2`:  doesn't need backport, but may need to resurrect EKCP
  - [x] `release-2.Y-1` (2.3 https://github.com/cloudfoundry-incubator/kubecf/pull/1103, backport of https://github.com/cloudfoundry-incubator/kubecf/pull/1108 can be skipped, since master will have it and therefore it will not collide)
- - [x] document in [release guidelines](https://github.com/cloudfoundry-incubator/kubecf/blob/master/doc/release/release-guidelines.md) that one needs to add a new pipeline for the new release branch. (https://github.com/cloudfoundry-incubator/kubecf/pull/1104)

Caveats:
[...]

PR #1098, opened 2020-07-02
CI: Add pr_base_branch option to filter PRs by which branch(es) they target

viccuad:
## Description

Add pr_base_branch option to filter PRs by which branch(es) they target.

The resource either allows for the parameter to be set to 1 branch, or not set,
for all branches.
See: https://github.com/telia-oss/github-pr-resource#source-configuration

## Motivation and Context

Needed for https://github.com/cloudfoundry-incubator/kubecf/issues/1095.

## How Has This Been Tested?

Manually set the new option `pr_base_branch` to `~`, `release-2.3`, `foo` and verified the gomplate works as intended. Deployed kubecf-viccuad pipeline with those options and verified that the kubecf-pr resource lists the correct PRs to test.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/kubecf.yaml
+++ b/.concourse/kubecf.yaml
@@ -31,6 +31,7 @@ branches: # Repository branches to track
 - release-2.3
 pr_resources:
 - pr
+pr_base_branch: ~ # filter PRs depending on the branch they target. `~` for all branches
 
 # Stable Jobs
 stable_jobs:
--- a/.concourse/README.md
+++ b/.concourse/README.md
@@ -30,8 +30,8 @@ If you wish to deploy a custom pipeline:
    `<your-pipeline-name>.yaml`
 2. Edit the yaml and disable production options as said by the NOTEs (publishing
    artifacts, updating github status, s3 buckets to consume, etc)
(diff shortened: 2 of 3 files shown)


.concourse/kubecf.yaml
```diff
@@ -31,6 +31,7 @@ branches: # Repository branches to track
 - release-2.3
 pr_resources:
 - pr
+pr_base_branch: ~ # filter PRs depending on the branch they target. `~` for all branches
```
jimmykarily, line 34: based on the comment below, the null value here doesn't mean "all branches", so this comment is wrong?

viccuad: the `# filter PRs depending on the branch they target. ~ for all branches` is a comment of the option `pr_base_branch`. The implementation is different.

I prefer to have something like `~` to denote "test all branches", rather than using `all` or `all-branches` or whatever. With our luck, we would get a branch named `all` to make a mess of it.

viccuad: Another option would be to act if the `pr_base_branch` key is present, and leave it commented in the docs. But people tend to overlook commented code.

PR #1099, opened 2020-07-02
Add `workertags` option to pipeline config

viccuad:
## Description

Add `workertags` option to pipeline config.
It can be null for no tags, or an array of tags.

## Motivation and Context

Allows to deploy and run the pipeline only in some specific concourse worker.

## How Has This Been Tested?

https://concourse.suse.dev/teams/main/pipelines/kubecf-viccuad

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/kubecf.yaml
+++ b/.concourse/kubecf.yaml
@@ -1,5 +1,8 @@
 ---
 
+workertags:
+  # - yourtag
+
 # NOTE use your own bucket when developing
 s3_bucket: kubecf-ci
 s3_bucket_region: eu-central-1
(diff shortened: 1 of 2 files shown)


.concourse/kubecf.yaml
```diff
@@ -1,5 +1,8 @@
 ---
 
+workertags:
```
jimmykarily, line 3: Should we add a comment on how this is supposed to be used? E.g.
`Specify a worker tag if you want your whole pipeline to run on specific (tagged) workers. This can be used to prevent resource sharing between your pipeline and the production one.`

jimmykarily:
This is something easy to forget when adding jobs but there is no simple way to guarantee otherwise, so LGTM.

.concourse/kubecf.yaml
viccuad: done!

PR #1105, opened 2020-07-03
Only test PRs open against master branch

viccuad:
## Description

If we leave testing PRs open against all branches, it will test PRs against
release branches. Those are:
- ensured to fail, since they don't have the bash-is-beautiful build system for
example
- already being tested correctly in their respective release pipelines

## Motivation and Context

Needed by https://github.com/cloudfoundry-incubator/kubecf/issues/1095

## How Has This Been Tested?

When we added the config option.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/kubecf.yaml
+++ b/.concourse/kubecf.yaml
@@ -30,11 +30,9 @@ availableCfSchedulers: # Diego / Eirini
 
 branches: # Repository branches to track
 - master
-- release-2.2
-- release-2.3
 pr_resources:
 - pr
-pr_base_branch: ~ # filter PRs depending on the branch they target. `~` for all branches
+pr_base_branch: master # filter PRs depending on the branch they target. `~` for all branches
 
 # Stable Jobs
 stable_jobs:


viccuad:
flown.

jandubois:
@viccuad 

> If we leave testing PRs open against all branches, it will test PRs against
> release branches. Those are:
> 
> * ensured to fail, since they don't have the bash-is-beautiful build system for
>   example

That should be a temporary problem only.

> * already being tested correctly in their respective release pipelines

I don't understand this argument. Which pipelines are testing PRs against release branches? If you mean the release branch pipelines themselves, then they are running too late. Nothing should be merged into a release branch that hasn't passed CI.

Isn't this argument the same as saying there is no need to tests PRs against master either, as all master PRs will be tested by the master pipeline?

viccuad:
> That should be a temporary problem only.

Sadly, it's not temporary. We need pipelines deployed from the respective branches: `release2-2`, `release-2.3`, etc will never be bazel-less, will have a different set of prod and experimental jobs, and different set of tests.

(See linked catch-all issue https://github.com/cloudfoundry-incubator/kubecf/issues/1095)

> Isn't this argument the same as saying there is no need to tests PRs against master either, as all master PRs will be tested by the master pipeline?

No. In the future we will have a `release-X.Y` pipeline, that both tests PRs against `release-X.Y`, and continuosly the branch, for being able to cut a release. 
This PR we are commenting on is to make the master `kubecf` pipeline not double-test the PRs opened against `release-X.Y`.

jandubois:
> In the future we will have a `release-X.Y` pipeline, that both tests PRs against `release-X.Y`, and continuosly the branch, for being able to cut a release.

Ok, that is even better. I wasn't aware that the release pipelines would also test PRs and not just the merged state (as they are now). Happy to see the direction this is moving!

PR #1110, opened 2020-07-07
docs: replace breaking change commit definition

fargozhu:
## Description

Switch the breaking change definition from a the word "break" to the exclamation mark "!".

## Motivation and Context

To follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) specifications.

## How Has This Been Tested?
N/A

N/A

## Types of changes
N/A

## Checklist:
N/A

### DIFF
--- a/doc/architecture/decisions/0002-git-commit-messages.md
+++ b/doc/architecture/decisions/0002-git-commit-messages.md
@@ -32,16 +32,8 @@ Most wanted rules:
 
 Other rules are also important!
 
-In order to automate the release notes/changelog process, we also need to define basic semantics:
-
-| Commit message                                           | Release type                                            |
-|----------------------------------------------------------|-------------------------------------------------------- |
-| fix: Stop the O2 leak dock comportment (DC)              | Patch release                                           |
-| feat: Implements dog space travel module (BARK)          | Feature release                                         |
-| <type>!: Removes the compase from the control room       | Exclamation after the type mark means a breaking change |
-| docs: Peter and the wolf: Cloud version                  | Documentation only changes                              |
-| style: Replace tabs with spaces                          | Changes that not affect the code                        |
-| test: Add unit test at last                              | Add or correct existing tests                           |
+In order to automate the release notes/changelog process, we will enforce the
+[conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) rules usage. 
 
 ## Consequences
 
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0002-git-commit-messages.md
```diff
@@ -34,18 +34,18 @@ Other rules are also important!
...
-| feat: Implements dog space travel module (BARK)          | Feature release                  |
-| break: Removes the compase from the control room (DOOM)  | Breaking change release          |
-| docs: Peter and the wolf: Cloud version                  | Documentation only changes       |
-| style: Replace tabs with spaces                          | Changes that not affect the code |
-| test: Add unit test at last                              | Add or correct existing tests    |
+| Commit message                                           | Release type                                            |
+|----------------------------------------------------------|-------------------------------------------------------- |
+| fix: Stop the O2 leak dock comportment (DC)              | Patch release                                           |
+| feat: Implements dog space travel module (BARK)          | Feature release                                         |
+| <type>!: Removes the compase from the control room       | Exclamation after the type mark means a breaking change |
```
mook-as, line 41: That gets parsed by the markdown parser as a HTML tag, and is invisible in the preview.  Also, this makes no sense as an example, since `<type>` is unspecified.  We need to rewrite more of this document to be clear.

fargozhu: @mook-as table removed since we are relying on the conventional commits document.

mook-as:
@fargozhu Hi! Just checking up to see if there's anything I should help with here, since this appears to be stuck.

viovanov:
@mook-as I think we've agreed on conventional commits in past calls, this was just updating the ADR.

mook-as:
@viovanov I have no issues with using conventional commits, we just need proper documentation to make sure people do the correct things.  As it is, you've merged it without a (recommended) list of types to use, so at this point it's going to end up even more confusing than before…

fargozhu:
> @viovanov I have no issues with using conventional commits, we just need proper documentation to make sure people do the correct things. As it is, you've merged it without a (recommended) list of types to use, so at this point it's going to end up even more confusing than before…

@mook-as my initial idea was to start with the well defined ones: fix, feat and breaking changes(!) and added more as we move forward.

mook-as:
@fargozhu Unfortunately, `breaking change` is itself not clearly specified (there are two options, and it is unclear if we're taking one over the other).  Having an under-specified spec is kind of… not helpful.

Issue #1010, opened 2020-06-11
Integrate the bazel replacement scripts in pipeline(s) and catapult


viovanov:
## Acceptance Criteria

- [x] all KubeCF pipelines use the new scripts, not `bazel` (https://github.com/cloudfoundry-incubator/kubecf/pull/1111, https://github.com/cloudfoundry-incubator/kubecf/pull/1163)
- [x] catapult no longer uses bazel either (https://github.com/SUSE/catapult/pull/243)
- [x] pipelines are green 🟢
- [x] update docs with instruction on how to build using the new scripts, and remove bazel references (https://github.com/cloudfoundry-incubator/kubecf/pull/1162)
- [x] update also https://kubecf.suse.dev

viovanov:
Blocked by #1009

viccuad:
PR to remove Bazel from Catapult: https://github.com/SUSE/catapult/pull/243. Note that Catapult was not used to build kubecf in CI, so merging this PR has no effect on current CI.

viccuad:
Blocked on https://github.com/cloudfoundry-incubator/kubecf/issues/1095

viccuad:
we should not forget https://kubecf.suse.dev, added to the list

viccuad:
Made minimal changes in kubecf.suse.dev, but opened https://github.com/cloudfoundry-incubator/kubecf-docs/pull/18 for dropping repetition of dev docs and pointing directly to the kubecf repo.

viccuad:
Updated kubecf.suse.dev manually, postponing cloudfoundry-incubator/kubecf-docs#18 to be considered in https://github.com/cloudfoundry-incubator/kubecf/issues/1200.

viccuad:
Card done, closing.

PR #1111, opened 2020-07-09
CI: Remove Bazel usage from the pipeline and port sample-values test to bash build system

viccuad:
## Description

- Substitute Bazel with Bash build system for lint & build jobs in the pipeline.
- Use `splatform/catapult` container image instead of `thulioassis/bazel-docker-image` in those jobs.
- Port functionality of `bazel test //deploy/helm/kubecf:values_doc_test` into `make kubecf-build`.
- Only print errors when testing the sample `values.yaml`, to leave `make kubecf-build` output unchanged.

## Motivation and Context

Part of https://github.com/cloudfoundry-incubator/kubecf/issues/1010.

## How Has This Been Tested?

Testing in https://concourse.suse.dev/teams/main/pipelines/kubecf-rmbazel
lint, build, publish should be green in master (maybe publish will fail because the tar is already there). The pr tab is used for updating this PR's github statuses.

Manual tests for the change in testing sample values doc tests have been performed:
- Bazel:
  * true positive by running bazel test //deploy/helm/kubecf:values_doc_test
  * true negative by running bazel test //deploy/helm/kubecf:values_doc_test and
    removing comment for system_domain in deploy/helm/kubecf/values.yaml
- Bash build scripts:
  * true positive by running `make kubecf-build`
  * true negative by running `make kubecf-build` and removing comment for
    system_domain in deploy/helm/kubecf/values.yaml

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -415,8 +415,7 @@ jobs:
       image_resource:
         type: registry-image
         source:
-          repository: thulioassis/bazel-docker-image
-          tag: 2.0.0
+          repository: splatform/catapult
       inputs:
       - name: kubecf-{{ $branch }}
       outputs:
@@ -427,7 +426,9 @@ jobs:
         - -ce
         - |
           cd kubecf-{{ $branch }}
-          ./dev/build.sh ../output
+          VERSION="$(./scripts/version.sh)"
+          TARGET_FILE=../output/kubecf-${VERSION}.tgz make kubecf-build
+          TARGET_FILE=../output/kubecf-bundle-${VERSION}.tgz make kubecf-bundle
           timestamp=$(date +%s%3N)
           for file in ../output/*.tgz; do
               mv "$file" "${file%.tgz}-${timestamp}.tgz"
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -340,10 +340,11 @@ jobs:
     {{- end }}
     config:
       platform: linux
-      image_resource:
+      image_resource: &catapult-image
         type: registry-image
         source:
           repository: splatform/catapult
+          tag: bedde80dff4cd8c296c5c9774f83a9a191a963af
       inputs:
       - name: kubecf-{{ $branch }}
       run:
@@ -409,9 +410,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: kubecf-{{ $branch }}
       outputs:
@@ -533,9 +532,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: kubecf
       - name: catapult
@@ -594,9 +591,7 @@ jobs:
         config:
           platform: linux
           image_resource:
-            type: registry-image
-            source:
-              repository: splatform/catapult
+            <<: *catapult-image
           inputs:
           - name: semver.gke-cluster
           run:
@@ -768,9 +763,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: catapult
       - name: kubecf
@@ -945,9 +938,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: catapult
       - name: kubecf
@@ -1112,9 +1103,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: catapult
       - name: kubecf
@@ -1282,9 +1271,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: catapult
       - name: kubecf
@@ -1450,9 +1437,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: catapult
       - name: kubecf
@@ -1616,9 +1601,7 @@ jobs:
     config:
       platform: linux
       image_resource:
-        type: registry-image
-        source:
-          repository: splatform/catapult
+        <<: *catapult-image
       inputs:
       - name: catapult
       - name: kubecf
@@ -1739,9 +1722,7 @@ jobs:
 #     config:
 #       platform: linux
 #       image_resource:
-#         type: registry-image
-#         source:
-#           repository: splatform/catapult
+#          <<: *catapult-image
 #       inputs:
 #       - name: catapult
 #       - name: commit-to-test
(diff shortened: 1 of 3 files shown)


viccuad:
Testing blocked by  #1095, hence, opened as draft.

viccuad:
Can be tested by approving, and then pinning this PR in https://concourse.suse.dev/teams/main/pipelines/kubecf-rmbazel?group=pr. This will update the github statuses here.

viccuad:
The bash build system downloads on its own shellcheck, yamllint, bosh, instead of using any of those binaries from the CI catapult image: https://concourse.suse.dev/teams/main/pipelines/kubecf-rmbazel/jobs/lint-viccuad-rm-bazel-ci/builds/5
We should fix that to speed up ci runs, or to not fail on GH blocking us after several downloads. But it doesn't need to block this card.

.concourse/pipeline.yaml.gomplate
```diff
@@ -343,8 +343,7 @@ jobs:
       image_resource:
         type: registry-image
         source:
-          repository: thulioassis/bazel-docker-image
-          tag: 2.0.0
+          repository: splatform/catapult
```
mook-as, line 346: Shouldn't this be pinned to _something_? Otherwise when we ship a new version (e.g. for kubecf 2.8) it will break older pipelines (e.g. kubecf 2.4).

.concourse/pipeline.yaml.gomplate
```diff
@@ -353,11 +352,7 @@ jobs:
         - -ce
         - |
           cd kubecf-{{ $branch }}
-          ./dev/linters/shellcheck.sh
-          ./dev/linters/yamllint.sh
-          ./dev/linters/helmlint.sh
-          bazel test //rules/kubecf:create_sample_values_test
-          bazel test //deploy/helm/kubecf:values_doc_test
+          make lint
```
mook-as, line 355: Does `make -C kubecf-{{ $branch }} lint` work? If yes, we won't need the `/bin/bash` anymore, I think?

.concourse/pipeline.yaml.gomplate
```diff
@@ -418,8 +413,7 @@ jobs:
       image_resource:
         type: registry-image
         source:
-          repository: thulioassis/bazel-docker-image
-          tag: 2.0.0
+          repository: splatform/catapult
```
mook-as, line 416: Same as above: pin to version please.  It might be useful to use YAML anchors here.

.concourse/pipeline.yaml.gomplate
```diff
@@ -430,7 +424,9 @@ jobs:
         - -ce
         - |
           cd kubecf-{{ $branch }}
-          ./dev/build.sh ../output
+          VERSION="$(./scripts/version.sh)"
+          TARGET_FILE=../output/kubecf-${VERSION}.tgz make kubecf-build
```
mook-as, line 428: If we're overriding the target file name anyway, can we not embed the timestamp in the initial file name? Why do we need the `mv` afterwards?

.concourse/pipeline.yaml.gomplate
```diff
@@ -1985,8 +1981,7 @@ jobs:
       image_resource:
         type: registry-image
         source:
-          repository: thulioassis/bazel-docker-image
-          tag: 2.0.0
+          repository: splatform/catapult
```
mook-as, line 1984: (See above)

scripts/kubecf-build.sh
```diff
@@ -61,6 +61,8 @@ done
 echo "operatorChartUrl: \"${CF_OPERATOR_URL//\{version\}/${CF_OPERATOR_VERSION}}\"" > "${HELM_DIR}/Metadata.yaml"
 
 ruby rules/kubecf/create_sample_values.rb "${HELM_DIR}/values.yaml" "${HELM_DIR}/sample-values.yaml"
+MODE=check ruby rules/kubecf/create_sample_values.rb "${HELM_DIR}/values.yaml" "${HELM_DIR}/sample-values.yaml"
```
mook-as, line 64: Should this be part of lint rather than build?

jandubois:
> The bash build system downloads on its own shellcheck, yamllint, bosh, instead of using any of those binaries from the CI catapult image

It should only do this when you have set `PINNED_TOOLS=1` before you invoke the linters. If you don't set this, then the build scripts will check if the installed version satisfies the minimum version requirements, and if satisfied, will use whatever you have in the `PATH`. I assume `catapult` is setting this, as I don't see any code in your PR doit it.

`PINNED_TOOLS=1` exists to match the Bazel behaviour of always downloading a specific version of the tools (which only worked when all dependencies were specified correctly, otherwise Bazel was also pulling things from `PATH`).

.concourse/pipeline.yaml.gomplate
viccuad: I agree. Found the pipeline that releases catapult images, started tagging there:
  https://github.com/SUSE/cloudfoundry/pull/180
  https://concourse.suse.de/teams/main/pipelines/catapult

Added a tag and anchored it throughout the pipeline.

.concourse/pipeline.yaml.gomplate
viccuad: True, changed. Leaving the `/bin/bash` for build step, making a one-liner there muddies things.

.concourse/pipeline.yaml.gomplate
viccuad: Done

.concourse/pipeline.yaml.gomplate
viccuad: True, simplified.

.concourse/pipeline.yaml.gomplate
viccuad: Done

scripts/kubecf-build.sh
viccuad: Could be.
This check needs to be run after the `sample-values` creation, which needs a build folder though. It could be added as a target of `make lint`, but then we are duplicating the creation and handling of the temporal folders. 

Maybe this is better solved when we remove Bazel and refactor `rules/`.

viccuad:
Tackled the review commits, flown the pipeline again, https://concourse.suse.dev/teams/main/pipelines/kubecf-rmbazel/jobs/lint-viccuad-rm-bazel-ci/builds/11 onwards.

viccuad:
sigh, the make lint call fails, I don't see why: https://concourse.suse.dev/teams/main/pipelines/kubecf-rmbazel/jobs/lint-viccuad-rm-bazel-ci/builds/17
But I need to leave, I will leave this undone, sorry.

.concourse/pipeline.yaml.gomplate
```diff
@@ -340,24 +340,18 @@ jobs:
...
         args:
-        - -ce
-        - |
-          cd kubecf-{{ $branch }}
-          ./dev/linters/shellcheck.sh
-          ./dev/linters/yamllint.sh
-          ./dev/linters/helmlint.sh
-          bazel test //rules/kubecf:create_sample_values_test
-          bazel test //deploy/helm/kubecf:values_doc_test
+        - kubecf-{{ $branch }}
```
mook-as, line 353: You're missing the `-C` to change directory, I think.

.concourse/pipeline.yaml.gomplate
mook-as: Hmm, the `repository` part is the same as the merged hash right? Do we need the explicit overrides here?

(As far as I can tell all the other uses of the anchor are fine.)

.concourse/pipeline.yaml.gomplate
viccuad: Correct, thanks!

viccuad:
Thanks for the review!

Fixed make arguments, removed left-over repository key on *catapult-image yaml pointer, added `shellcheck` and `yamllint` to catapult image, and bumped the image's tag being pinned in this pipeline.

PR #1117, opened 2020-07-10
Make xip.io domain configurable for dev setups

jandubois:
`xip.io` is often flaky, so make it easy to switch to an alternative.

### DIFF
--- a/scripts/kubecf-apply.sh
+++ b/scripts/kubecf-apply.sh
@@ -3,6 +3,8 @@ source scripts/include/setup.sh
 
 require_tools helm kubectl
 
+: "${MAGIC_DNS_SERVICE:=xip.io}"
+
 HELM_ARGS=()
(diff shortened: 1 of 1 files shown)


scripts/kubecf-apply.sh
```diff
@@ -3,6 +3,8 @@ source scripts/include/setup.sh
 
 require_tools helm kubectl
 
+: "${XIP_IO:=xip.io}"
```
mook-as, line 6: I believe the term use use [elsewhere](https://github.com/SUSE/cf-ci/blob/567eb68268846b22918eb13321b7bd8424d1fc58/qa-pipelines/tasks/clean_terraform_deploy.yml#L16) for this is `MAGIC_DNS_SERVICE`; would it be okay to reuse that for consistency?

jandubois: Absolutely; force-pushed an update.

PR #1077, opened 2020-06-25
Disable/clear storage class setting for eirini persi broker

andreas-kupries:
## Description

By not specifying an SC we can make use of the DefaultSC admission plugin
to get something sensible in a more complex deployment with many helm charts
and sub charts.

## Motivation and Context

See #819

## How Has This Been Tested?

This pull request is one in a group of 3 related PRs.
They have to be reviewed and tested together.
The instructions for such are found at https://github.com/cloudfoundry-incubator/kubecf/issues/819#issuecomment-649841863
The full set of PRs is listed in the description of #819 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code has security implications.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1027, opened 2020-06-16
Use a newer (fixed) eirinifs/bits image and make ssh-proxy svc public

jimmykarily:
Problems fixed:

- The eirinifs image didn't start the sshd daemon (https://github.com/SUSE/scf-eirini-ci/commit/75fb413094ff0480cd882136f512785baed53381)
- The ssh-proxy service wasn't accessible from the cf-cli (didn't have an external ip or a loadbalancer ip)
- The launchers in the eirinifs image were missing some environment variables and launchers were not executable (https://github.com/SUSE/scf-eirini-ci/commit/68d91221ad72339164a82e34811584f0430390f7)

Deployed this branch on gke and `cf ssh` works.

Fixes: https://github.com/cloudfoundry-incubator/kubecf/issues/1026

## Description

This should fix the rest of the issue in https://github.com/cloudfoundry-incubator/kubecf/issues/1016

(everything other than the extension fix which lives here: https://github.com/SUSE/eirini-ssh/pull/22)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code has security implications.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #819, opened 2020-05-06
eirini-persi-broker: should use default storage class when unspecified


mook-as:
# Status

# Linked Pull requests

  - #1077
  - https://github.com/SUSE/eirinix-helm-release/pull/12
  - https://github.com/SUSE/eirini-persi-broker/pull/6

# Testing

See https://github.com/cloudfoundry-incubator/kubecf/issues/819#issuecomment-649841863

# Post review actions

After positive review and testing of these PRs merging has to be done from the bottom up:

  1. First merge the broker code changes, i.e. https://github.com/SUSE/eirini-persi-broker/pull/6
     This should trigger CI to regenerate all the relevant docker images.
  2. After that the chart changes, i.e. https://github.com/SUSE/eirinix-helm-release/pull/12, should be mergable.
     :warning: I do not know if it will be necessary to bump version numbers and/or tags.
     This may trigger CI to regenerate a downloadable chart. I do not know for sure.
     We certainly will have to get a new version number for the chart, which is needed by the next step.
  3. Now we can update #1077 with the updated chart, and then merge.
     The instructions for updating sub charts are found in `doc/Contribute.md`.

# Description

**Is your feature request related to a problem? Please describe.**
Currently, when using the eirini persi broker, we _must_ specify a storage class.  This is unwieldy for deployment using helm charts, as we cannot rely on the `DefaultStorageClass` admission plugin to give us something sensible.

**Describe the solution you'd like**
- The broker should create a PVC with no storage class (which is distinct from a PVC with an empty string storage class) when no storage class is specified.
- The eirinix helm chart should default to having no storage class specified.

**Describe alternatives you've considered**
Making the customer configure the storage class every time, but that seems terrible.

**Additional context**
This is a feature request for [eirini-persi-broker](https://github.com/SUSE/eirini-persi-broker).

The value is [parsed here](https://github.com/SUSE/eirini-persi-broker/blob/a182eac006237994734e02645c39fe84309fed5b/config/config.go#L45) and [used here](https://github.com/SUSE/eirini-persi-broker/blob/6e9a55465bd9efbc02e9d5831ab55866b342d31e/broker/broker.go#L142); the usage always takes the address of a string, so we cannot pass in null values.  According to the [documentation](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1) we want to _not set_ the storage class name, which is distinct from setting it to the empty string.

andreas-kupries:
[...]

PR #1119, opened 2020-07-10
Updated the eirinix to re-introduce the ssh-proxy fixes.

andreas-kupries:
## Description

This brings back the local #1027 changes which were lost with #1077 updating the sub charts.

## Motivation and Context

See the above referenced PR's for the context.
Note further that this fix should obviate the need for #1118 

## How Has This Been Tested?

It will be tested through the CI (eirini smoke tests)

## Types of changes

## Checklist:

### DIFF
--- a/deploy/helm/kubecf/charts/eirini-extensions/Chart.yaml
+++ b/deploy/helm/kubecf/charts/eirini-extensions/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Eirini Extensions for KubeCF
 name: eirini-extensions
-version: 0.0.0-10+g3df5bb4
+version: 0.0.0-12+ge147751
--- a/deploy/helm/kubecf/charts/eirini-extensions/templates/persi-broker.yaml
+++ b/deploy/helm/kubecf/charts/eirini-extensions/templates/persi-broker.yaml
@@ -43,9 +43,11 @@ spec:
                 - plan_id: {{ .id }}
                   plan_name: {{ .name }}
                   description: {{ .description }}
-                  kube_storage_class: {{ .kube_storage_class }}
+                  {{- with .kube_storage_class }}
+                  kube_storage_class: {{ . }}
+                  {{- end }}
                   free: {{ .free }}
-                  deault_size: {{ .default_size }}
+                  default_size: {{ .default_size }}
                 {{- end }}
               description: {{ .description }}
               long_description: {{ .long_description }}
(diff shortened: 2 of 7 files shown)


andreas-kupries:
@jimmykarily @viovanov  :warning: :warning: :warning: This is not complete yet, and has issues.

  1. The lint failed, see below. Not investigated why yet, it is EOD/EOW.
  1. The https://github.com/SUSE/eirinix-helm-release/pull/13 fixed only the ssh-proxy template, the values.yaml was forgotten.
  1. The https://github.com/SUSE/eirinix-helm-release/pull/14 has the changes to the `values.yaml`, but looks incorrect.
      The original PR #1027 used an explicit version tag for the ssh docker image in the sub chart values.yaml.
      And right now the PR 14 has the pinned image version directly added to it. This likely will not work.

      Because in the `eirinix-helm-release` chart the relevant `values.yaml` has `latest` everywhere for all the tags and
      the https://concourse.suse.dev/teams/main/pipelines/eirinix pipeline replaces that with proper tags.
      The pipeline uses the resource https://concourse.suse.dev/teams/main/pipelines/eirinix/resources/s3.ssh-version
      to determine the version tag to use in the replacement.

     This resource has the tag we want (`v0.0.0-0.g04a3871`), but picks the older tag (`v0.0.0-0.ge2ed1ab`) somehow.
     (I suspect it is a sorting issue on the filenames)

jandubois:
> 1. The lint failed, see below. Not investigated why yet, it is EOD/EOW.

It fails because https://github.com/SUSE/eirinix-helm-release/pull/13/files is missing an `{{- end }}` line at the bottom, to close the `{{- with ... }}` block.

andreas-kupries:
Fixed the missing `{{- end}}` in https://github.com/SUSE/eirinix-helm-release/pull/14.
Seeing conflicts I will also rebase this one.

andreas-kupries:
@jandubois 

Pushed a rebase to undo the #1118 revert while keeping the #1027 work.

This still has some manual fixups for that until https://github.com/SUSE/eirinix-helm-release/pull/14 is properly done, including having resolved with @viovanov and @jimmykarily how to fix the image tagging for that.

That said, with the manual fixups we should be able to run it through CI and see if that
  1. fixes the ssh-proxy, and
  2. the smoke tests.

For that I require a new approval.

jimmykarily:
I think what [the eirinix pipeline](https://concourse.suse.dev/teams/main/pipelines/eirinix) is doing to replace "latest" with proper tags is a bit error prone. When one wants to replace "latest" with whatever tag equals to "latest" they should pull "latest" and then either find the relevant tag in some metadata (if that is available) or simply replace with the digest version. The way we do it we introduced one more moving part which is the s3 file.

In any case, if we fixed the versioning of the various components and started releasing with at least patch versions and not `v0.0.0-some-prerelease-thing-here` sorting would work and all things would be good as they are now. Should I open an issue to change versioning of components to use the patch part instead of pre-release?

andreas-kupries:
@jandubois @jimmykarily That sounds good. Please make sure that the new issue is high-priority, as the versioning stuff here blocks not just this, but #1077 / #819 as well.

Do you know who would have the best knowledge of the pipeline to work on this ?

PR #1133, opened 2020-07-17
feat: Add wait targets for make

mook-as:
## Description
This adds two new targets for the make-based build system, which can be used to wait for cf-operator & kubecf to finish deploying.  This is useful in CI where we may want to let things settle before running the consequent steps (e.g. to run tests).

## Motivation and Context
Wanted to use this in CI (spiking on using GitHub actions, but this is useful elsewhere).

## How Has This Been Tested?
Ran a lot.

## Types of changes

## Checklist:

This change is for development only, this has no effect on the shipped code.  It is also currently not used anywhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,12 +57,18 @@ cf-login:
 cf-operator-apply:
 	@./scripts/cf-operator-apply.sh
 
+cf-operator-wait:
+	@./scripts/cf-operator-wait.sh
+
 kubecf-apply:
 	@./scripts/kubecf-build.sh
 	@./scripts/kubecf-apply.sh
 
 kubecf-delete:
 	@./scripts/kubecf-delete.sh
+
+kubecf-wait:
+	@./scripts/kubecf-wait.sh
 
 ########################################################################
 # Test
--- a/scripts/cf-operator-wait.sh
+++ b/scripts/cf-operator-wait.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# This script waits untils the operator is ready
+
+source scripts/include/setup.sh
+
+require_tools kubectl retry
+
+# Wait for CRDs to show up
+crds=(
+    boshdeployments.quarks.cloudfoundry.org
+    quarksjobs.quarks.cloudfoundry.org
+    quarkssecrets.quarks.cloudfoundry.org
+    quarksstatefulsets.quarks.cloudfoundry.org
+)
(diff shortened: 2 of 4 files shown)


jandubois:
I'll do a code review on Monday, but I've used the scripts several times today, and they are working well.

I have a few ideas I would like to see implemented; but I'm fine merging the scripts as-is (after code review), and opening a feature request instead:

1. I think the output (stdout & stderr) of the command executed by `retry` should not be shown; we expect the command to fail for a while before it succeeds, and the output is just clutter:

        ```
        1/60: kubectl wait --for=condition=Established customresourcedefinition.apiextensions.k8s.io/boshdeployments.quarks.cloudfoundry.org
        Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io "boshdeployments.quarks.cloudfoundry.org" not found
        ```

2. Instead of (or in addition) to the retry count (`29/180`) I would prefer to see elapsed time (`145s`).

3. When printing to a TTY I would prefer that identical lines (except for count/elapsed time) overwrite each other instead of scrolling the display.

Random notes:

I think the `retry` script will eventually move to the `kubecf-tools` repo, but we can leave it here until we start using the additional repo.

Bikeshed: The `DELAY` parameter to the `retry` command should be called `EVERY`. Maybe... 😄

mook-as:
I like those suggestions, but I'd prefer to do those in a follow-up PR, if possible; this would unblock me in the short term.  As they're all about the `retry` function, it looks like what you want can be done with GNU parallel instead… that involves figuring out how to get perl though, so I gave up for now.

jandubois:
> I like those suggestions, but I'd prefer to do those in a follow-up PR

Yes, that is fine, approved PR as-is now.



> As they're all about the `retry` function, it looks like what you want can be done with GNU parallel instead… that involves figuring out how to get perl though, so I gave up for now.

`perl` is installed by default in all Linux distros, but I don't think we need to use `parallel` to do this; it is essentially just something like this:

```
if [ -t 1 ]; then
    CR="\r"
    NL=" "
else
    CR=""
    NL="\n"
fi

COUNT=1
while [ $COUNT -lt 10 ]; do
    printf "${CR}Count is %d${NL}" $COUNT
    COUNT=$(( COUNT + 1 ))
    sleep 1
done
[ -n "$CR" ] && printf "\n"
```

I can create a follow-up PR myself once this one is merged.

PR #1141, opened 2020-07-20
feat: Improve wait scripts for interactive usage

jandubois:
* Don't display errors from the command checking status; it is expected to fail while the status is not yet ready.

* Display time since the start of the script for each wait check, in addition to the retry/max counts.

* Overwrite status output for a single check when writing to a tty to avoid massive scrolling.

Also add an `all` target to deploy cf-operator and kubecf, and then log in when everything is ready, e.g. `FEATURE_EIRINI=1 make all` to run kubecf with Eirini on the cluster specified by the current `KUBECONFIG`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,12 @@ minikube-delete:
 ########################################################################
 # Run
 
-all: cf-operator-apply cf-operator-wait kubecf-apply kubecf-wait cf-login
+all:
+	./scripts/cf-operator-apply
+	./scripts/cf-operator-wait
+	./scripts/kubecf-apply
+	./scripts/kubecf-wait
+	./scripts/cf-login
 
 cf-login:
 	@./scripts/cf-login.sh
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,8 @@ minikube-delete:
 ########################################################################
 # Run
 
+all: cf-operator-apply cf-operator-wait kubecf-apply kubecf-wait cf-login
+
 cf-login:
 	@./scripts/cf-login.sh
 
--- a/scripts/tools/retry.sh
+++ b/scripts/tools/retry.sh
@@ -11,12 +11,12 @@ function retry {
 
     local cr
     local nl
+    local isa_tty="false"
     if [ -t 1 ]; then
+        isa_tty="true"
         # Output is to a TTY; don't scroll display while waiting.
         cr="\r"
         # Display a trailing space between the last character and the cursor.
-        # Also overwrites trailing character when number of seconds shrink from
-        # two to one digit, e.g. "1m 58s" → "2m 3s".
         nl=" "
     else
         # Output is to a file; don't overwrite lines.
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -51,6 +51,8 @@ minikube-delete:
 ########################################################################
 # Run
 
+all: cf-operator-apply cf-operator-wait kubecf-apply kubecf-wait cf-login
```
mook-as, line 54: Nothing guarantees that these will run in the correct order.  Please add [order-only prerequisites](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html#Prerequisite-Types) as appropriate.

scripts/tools/retry.sh
```diff
@@ -9,13 +9,33 @@ function retry {
     local delay="${DELAY:-5}"
     local i=0
 
+    local cr
+    local nl
```
mook-as, line 13: I was going to ask for better names… then realized these are actually the full names, I just normally see them in upper case (`CR`/`LF`). Nothing to do here, unless you feel like adding excessive comments (which is kind of pointless).

scripts/tools/retry.sh
```diff
@@ -9,13 +9,33 @@ function retry {
...
+    else
+        # Output is to a file; don't overwrite lines.
+        cr=""
+        nl="\n"
+    fi
+
     while test "${i}" -lt "${max}" ; do
-        printf "%s/%s: %s\\n" "$(( i + 1 ))" "${max}" "$*"
-        if "$@" ; then
+        printf "${cr}[%dm %ds] %s/%s: %s${nl}" \
```
mook-as, line 28:
```suggestion
        printf "${cr}[%dm %2ds] %s/%s: %s${nl}" \
```

Just to make the seconds line up. If you feel like doing more work, we can also do
```suggestion
        printf "${cr}[%dm %2ds] %${#max}s/%${#max}s: %s${nl}" \
```
This makes the retry counter line up too, hopefully.  We can use `%*s` with an extra format arg before the appropriate value to sub it in later, if you prefer.

scripts/tools/retry.sh
```diff
@@ -9,13 +9,33 @@ function retry {
...
+               "$(( SECONDS / 60 ))" "$(( SECONDS % 60))"\
+               "$(( i + 1 ))" "${max}" \
+               "$*"
+        if "$@" &> /dev/null ; then
```
mook-as, line 32: I would prefer capturing the output and printing it on success (or failure at end of loop), because that's going to be helpful for troubleshooting, especially if we end up using this in CI.

scripts/tools/retry.sh
```diff
@@ -9,13 +9,33 @@ function retry {
...
+            [ -n "${cr}" ] && printf "\n"
```
mook-as, line 33: This feels a bit too magical; we're not really testing for `cf`-is-empty, we're testing for `nl`-contains-a-new-line.

scripts/tools/retry.sh
```diff
@@ -9,13 +9,33 @@ function retry {
...
             return
         fi
         sleep "${delay}"
         i="$(( i + 1 ))"
     done
+    [ -n "${cr}" ] && printf "\n"
```
mook-as, line 39: Same as above.

scripts/tools/retry.sh
jandubois: I made them lowercase because they are local variables. I think this is fine.

Makefile
mook-as: Fixed in cac7e603fdcc330177fa48a36304a9dafc942970 by making it not make-based.

scripts/tools/retry.sh
```diff
@@ -9,13 +9,49 @@ function retry {
...
+        nl="\n"
+    fi
+
+    local output
+    output="$(mktemp)"
     while test "${i}" -lt "${max}" ; do
-        printf "%s/%s: %s\\n" "$(( i + 1 ))" "${max}" "$*"
-        if "$@" ; then
+        printf "${cr}[%2dm %2ds] %s/%s: %s${nl}" \
+               "$(( SECONDS / 60 ))" "$(( SECONDS % 60))"\
```
mook-as, line 31: Super-nit: inconsistent spacing (space before backslash or not?)

… Nope, I'm not blocking a merge on this.

scripts/tools/retry.sh
```diff
@@ -9,13 +9,49 @@ function retry {
...
+        # Display a trailing space between the last character and the cursor.
+        nl=" "
+    else
+        # Output is to a file; don't overwrite lines.
+        cr=""
+        nl="\n"
+    fi
+
+    local output
+    output="$(mktemp)"
```
mook-as, line 28: Hmm, we _could_ be leaving the temporary file behind if we get `^C`ed at the correct time… but I'm not sure I care.

PR #1160, opened 2020-07-23
Eirini: drop cluster.local references

### LINKED ISSUE #872: Don't assume cluster domain is cluster.local
The ops file for the eirini-helm instance group is full of URLs ending in `.svc.cluster.local`:

https://github.com/cloudfoundry-incubator/kubecf/blob/master/deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml

These URLs will only work if the cluster uses the default cluster domain name.

Not sure if there is a better solution, but at a minimum we need to make the cluster domain configurable.

mook-as:
## Description
This drops hard-coded instances where we assumed the Kubernetes cluster domain to be `cluster.local`; doing this is required as this isn't guaranteed on all Kubernetes deployments.

## Motivation and Context
Fixes #872 

## How Has This Been Tested?
Deployed locally (using Eirini), with a custom minikube:
- `MINIKUBE_EXTRA_OPTIONS` includes `--dns-domain=minikube --extra-config=kubelet.cluster-domain=minikube`
- Quarks-operator helm config includes `cluster.domain=minikube`
- Double checked that `/etc/resolv.conf` in the CC container doesn't include `cluster.local`

## Types of changes

## Checklist:

## Other Notes:
⚠️ Need to see how #1119 interferes with this PR.

### DIFF
--- a/deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml
+++ b/deploy/helm/kubecf/assets/operations/instance_groups/eirini-helm.yaml
@@ -19,20 +19,20 @@
 
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/opi?/url?
-  value: "https://eirini-opi.{{ .Release.Namespace }}.svc.cluster.local:8085"
+  value: "https://eirini-opi.{{ .Release.Namespace }}.svc:8085"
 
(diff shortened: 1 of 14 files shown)


andreas-kupries:
Re-tested locally. Eirini smoke tests ok. Going forward with merge.

PR #1161, opened 2020-07-24
docs: ADR for restricting scripting to ruby and bash

viovanov:


jandubois:
> Should we go further, and actually cover non-scripting languages?

I don't expect us to use non-scripting languages for our tooling; in the end we are just assembling a helm charts, deploying and testing it. But I'm fine with explicitly specify golang; if it ever comes up, we don't want to introduce any additional languages.

If we specify data formats, then I think both YAML and JSON should be fine. And I agree with a preference for golang templating, if possible, but feel less strict about it, in case CI requires something different in the future.

One area maybe worth mentioning: We should aim to keep the number of external tools smalls, e.g. use `ruby` to convert to JSON and use `jq` instead of adding a dependency on `yq`, if possible. Also, prefer using a tool written in Ruby to one in Python or Nodejs if the required functionality is available. But if not, then it is still fine to use (but not implement) tools in other scripting languages (e.g. `yamllint`  is implemented in Python). Since we don't use any NPM based tools yet, try to avoid adding a dependency on NodeJS if at all possible.

jandubois:
@viovanov There have been no new comments in 6 weeks, so I've tried to include the feedback so far in the ADR.

Please review and then maybe merge? Any further refinement can then happen via pull request.

doc/architecture/decisions/005-bash-and-ruby.md
```diff
@@ -0,0 +1,58 @@
...
+scripts should be unit-tested if possible.
+
+### Statically Typed and/or Compiled Languages
+
+We do not anticipate the need to write any tooling in a compiled language.
+Should this become necessary, the Go language should be choosen.
+
+### Templating Languages
+
+Using Go `text/template` based templating are preferred whenever possible.
```
andreas-kupries, line 34: `are` -> is`

andreas-kupries (dismissed):
My comment is just about a nit.

andreas-kupries (approved):
+1

Issue #1013, opened 2020-06-11
Versioning common script


viovanov:
## Acceptance Criteria

- should fail if it's not in a git repo
- should fail if there's no semver tag available
- version should be calculated as implemented in the quarks project (development versions should include number of commits since the latest tag and the commit `1.0.2-0.gb3f7a0c1`)
- implemented in ruby with tests
- version should be printed to stdout

Existing scripts for calculating versions:
https://github.com/SUSE/scf/blob/develop/make/include/versioning
https://github.com/SUSE/eirinix-helm-release/blob/master/bin/package#L34
https://github.com/cloudfoundry-incubator/quarks-operator/blob/master/bin/include/versioning
https://github.com/cloudfoundry-incubator/quarks-job/blob/master/bin/include/versioning

viovanov:
Once this is done, we should create stories to use it in all projects.

viovanov:
Blocked by #1012

viovanov:
Keep an eye on #1012 for pushing the script in its final location.

thardeck:
Please review https://github.com/SUSE/kubecf-tools/pull/1

thardeck:
https://github.com/SUSE/kubecf-tools/pull/2

PR #1162, opened 2020-07-24
Update docs for Make & Bash, and remove bazel references

viccuad:
## Description

Update docs for Make & Bash, and remove bazel references.

## Motivation and Context

Needed by https://github.com/cloudfoundry-incubator/kubecf/issues/1010.

## How Has This Been Tested?

No need, contains only doc changes.
Changes to __release-guidelines.md__ are simple enough that with the review should be enough.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code has security implications.
- [ ] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #1165, opened 2020-07-24
Use new versioning script from kubecf-tools

### LINKED ISSUE #766: Helm packaging fails if git revision starts with `0`
**Describe the bug**
If the git commit hash starts with `0`, then `bazel build //deploy/helm/kubecf` fails with:
> `Error: Version segment starts with 0`

**To Reproduce**
1. Make commits until the git commit hash starts with `0`
2. Run `bazel build //deploy/helm/kubecf`

**Expected behavior**
Build completes successfully.

**Environment**
- Kubecf [master](https://github.com/cloudfoundry-incubator/kubecf/commit/1699ccd2728df2a5735596047c990e447cd0513f) + a local commit that happens to start with `0` (`0266605f…`)

**Additional context**
- [`package.tmpl.rb`](https://github.com/cloudfoundry-incubator/kubecf/blob/1699ccd2728df2a5735596047c990e447cd0513f/rules/helm/package.tmpl.rb#L65) makes the version `v0.0.0-0266605f` (or something like that).
- This appears to not work; see https://github.com/helm/helm/issues/7064

We may want to change it to `v0.0.0-g0266605f` or something like that, similar to `git describe` output (where a `g` prefix ensures that identifier does _not_ count as numeric).

jandubois:
Adds kubecf-tools repo as a submodule per discussion with @viovanov 

The new versioning script was developed via #1013

Fixes #766

Example output:

```
$ make version
0.2.0-1455.gfa0a32f0-dirty
```

Requesting review from @viccuad to make sure it is compatible with catapult and CI.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "src/cf-deployment"]
 	path = src/cf-deployment
 	url = https://github.com/cloudfoundry/cf-deployment
+[submodule "src/kubecf-tools"]
+	path = src/kubecf-tools
+	url = https://github.com/cloudfoundry-incubator/kubecf-tools
--- a/scripts/include/versions.sh
+++ b/scripts/include/versions.sh
@@ -1,8 +1,5 @@
 # shellcheck shell=bash
 
-# Version is 0.0.0 on master, and x.y.z on release-x.y branches
-: "${KUBECF_VERSION:=0.0.0}"
-
 # Used by both kind-start and minikube-start
 : "${K8S_VERSION:=1.17.5}"
 
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 source scripts/include/setup.sh
 
-require_tools ruby
+require_tools git ruby
 
 ruby src/kubecf-tools/versioning/versioning.rb
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 source scripts/include/setup.sh
 
-require_tools git
+require_tools ruby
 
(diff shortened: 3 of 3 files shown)


viccuad:
We should update the release docs (https://github.com/cloudfoundry-incubator/kubecf/blob/master/doc/release/release-guidelines.md#files-to-change) with this PR, although it could also go into https://github.com/cloudfoundry-incubator/kubecf/pull/1162

viccuad:
The `make lint` job rightfully fails, because there aren't any tags present in the git checkout of the CI…

The git checkout performed by the concourse PR resource doesn't have any tags, nor remotes. We could add the following clutch or something similar to create a tag on the lint and build PR CI jobs:
```
{{- if ( eq $branch "pr" )}}
git tag 0.0.0-$(cat .git/resource/pr)-$(cat .git/resource/head_sha)
{{- end }}
```
This would make lint and build pass on PRs, and leaves the master ones untouched and working, since those use a different concourse resource.

Looking forward to move to Github actions.

jandubois:
> The `make lint` job rightfully fails, because there aren't any tags present in the git checkout of the CI…

I wish I had seen this before bothering Mark to help me figure this out...

Since we already use a fork of the github-pr-resource, I've fixed this by fetching tags as well: https://github.com/SUSE/github-pr-resource/commit/e92ede5

I've built a new image locally and updated our pipeline to use that: https://github.com/cloudfoundry-incubator/kubecf/pull/1168

Updated pipeline has been deployed, and it fixed this issue, which is already closed.

viccuad:
> We should update the release docs (https://github.com/cloudfoundry-incubator/kubecf/blob/master/doc/release/release-guidelines.md#files-to-change) with this PR, although it could also go into #1162

Updated release-guidelines.md in https://github.com/cloudfoundry-incubator/kubecf/pull/1162

PR #1176, opened 2020-07-29
Read cf-operator URL and SHA256 from dependencies.yaml

jandubois:
That is the only location updated by CI when we bump the cf-operator version, so this commit makes sure we don't get out-of-sync between bash and bazel build rules.

### DIFF
--- a/scripts/include/versions.sh
+++ b/scripts/include/versions.sh
@@ -2,9 +2,3 @@
 
 # Used by both kind-start and minikube-start
 : "${K8S_VERSION:=1.17.5}"
-
-# The cf-operator url and sha256 are currently being read from `dependencies.yaml` via the
-# functions in ../tools/cf_operator_url.sh
-# "${CF_OPERATOR_URL:=https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-{version\}.tgz}"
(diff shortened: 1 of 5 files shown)


scripts/include/versions.sh
```diff
@@ -3,6 +3,8 @@
 # Used by both kind-start and minikube-start
 : "${K8S_VERSION:=1.17.5}"
 
-: "${CF_OPERATOR_URL:=https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-{version\}.tgz}"
-: "${CF_OPERATOR_SHA256:=df225af203596ca6dc11131004704632c01579bbaf80f8439b42aa8a5e24a47a}"
-: "${CF_OPERATOR_VERSION:=5.0.0%2B0.gd7ac12bc}"
+# The cf-operator url and sha256 are currently being read from `dependencies.yaml` via the
+# functions in ../tools/cf_operator_url.sh
+# "${CF_OPERATOR_URL:=https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-{version\}.tgz}"
```
mook-as, line 8: Delete these? There's no point in keeping them as comments; if we ever want to put it back it'll be in git.  As it is, it's just ~guaranteed to be outdated because the checksum / version wouldn't match.

jandubois: Removed in a473657

PR #1022, opened 2020-06-15
Remove local nodePort registry in favor of changed bits service

andreas-kupries:
## Description

Removed `loadbalanced` property.
Moved `nodePort` property from `kubecf` section to `bits` section.

:warning: TODO: Update the bits chart.

## Motivation and Context

See #792
See SUSE/bits-service-release/pull/7 for the bits release changes.

## How Has This Been Tested?

Local minikube, SA eirini. Smoke tests run manually.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code has security implications.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #792, opened 2020-04-29
bits: expose the registry using a nodePort


viovanov:
**Is your feature request related to a problem? Please describe.**
We don't expose the internal eirini registry publicly.
We want to expose it using a nodeport, so the nodes can easily reach it.
Right now we're using our own service for this.

**Describe the solution you'd like**
The upstream bits helm chart should allow exposing its service using a NodePort

andreas-kupries:
@viovanov Questions

>  We don't expose the internal eirini registry publicly.

Is that true ? 
The bits helm chart (BHC) has a `services.loadbalanced` property.
While the default is `false` (causing no type to be set, making it `ClusterIP`),
in the `kubecf`this is set to `true`, giving service `bits` the `type: LoadBalancer`.
Is that not fully exposed ?

>  Right now we're using our own service for this.

Which service is that, and where?

viovanov:
@andreas-kupries the service we use is here:
https://github.com/cloudfoundry-incubator/kubecf/blob/master/deploy/helm/kubecf/templates/eirini.yaml#L54-L68

it should be part of the bits chart

andreas-kupries:
Thanks.

andreas-kupries:
PRs:
  - bits: SUSE/bits-service-release/pull/7
  - kubecf: #1022

andreas-kupries:
Blocked by #1179 as work on that can/will make this here moot.

jandubois:
> Blocked by #1179 as work on that can/will make this here moot.

#1179 has been merged and includes the ingress service inside the `bits` mixin. I believe there is nothing left to do here.

Issue #1179, opened 2020-07-30
Include the Eirini, EiriniX and Bits templates in KubeCF


viovanov:
**Is your feature request related to a problem? Please describe.**
These charts are not being used on their own by anyone.
Using subcharts is harder than just having templates in the main KubeCF chart.

**Describe the solution you'd like**
Include all template in the KubeCF chart.
Make sure to keep the namespaced value keys, so we don't break compatibility.

jandubois:
I want to work on this because I have some thoughts on how to structure the repo layout to support "addon" or "mixin" templates, based on the work I've been doing restructuring our config mechanism for the multi-stack support.

I'm not ready to start work on it right now, but have assigned to myself already. I will update this issue with more info in the next 2 days.

andreas-kupries:
:warning: Ticket #792 and its PRs #1022 (+ drafts #1182, #1185) becomes moot due to this.
:warning: If we make sure that what we want from #792 is done in the work for this ticket as well.

PR #1182, opened 2020-07-31
Next trial.

andreas-kupries:
See #792 for the controlling ticket, and #1022 for the original PR we are trialing a part of here. The important info is in the #1022 comments (bottom).


jandubois:
Superseded by #1179.

Issue #1144, opened 2020-07-21
PR Checking with GitHub Actions


viovanov:
## Acceptance Criteria 

- [x] should create clusters on GKE
- [x] github actions workflow that runs smoke, CATS, SITS and brain tests for PRs
- [x] runs diego
- [x] runs eirini
- [ ] doesn't run for PRs from a fork
- [x] should cleanup any cloud resources
- [x] should not be mandatory for merges

Use the work done here as a starting point: https://github.com/mook-as/kubecf/actions/

> [x] GKE clusters should have only one node

(Edited by @mook-as: Making checkboxes)

mook-as:
@prabalsharma Should this be back to In Progress given that the mentioned PR is merged, but this task isn't complete?

prabalsharma:
I am not sure what is not working in here, but if thats the case we will have to put this back in progress or create a new card to address the shortcomings.

viccuad:
It seems that we are not running SITS, CATS as Brains are not passing (eg: https://github.com/cloudfoundry-incubator/kubecf/runs/1120154551#step:20:1609). FWIW, we are not running Brains in Concourse.

viccuad:
Also, we need to port cats-internetless and cats-eirini configuration from https://github.com/SUSE/catapult/blob/master/modules/tests/kubecf-test.sh to have feature parity with Concourse there.

viccuad:
Brains have been disabled for now.

SITS are falling, because they cannot be run for Eirini, and we are configuring Diego as Eirini (doing `FEATURE_EIRINI=false make kubecf-apply` deploys eirini, `FEATURE_EIRINI` must be empty for it to not take effect).

mook-as:
I believe at this point we have everything for the original acceptance criteria; the only missing things are:
- Don't run from forks: We're still gating runs on labels (triggered by approval).  Also, for PRs from forks, [secrets are not passed in](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow); confirmed this with a [cross-owner PR](https://github.com/mook-as/kubecf/pull/10/checks?check_run_id=1461043602) where Thulio's PR didn't have access to my secrets.  Thus I believe this is doing the correct things.
- Clean up resources: Need to clean up DNS entries; see #1630 / https://github.com/SUSE/catapult/pull/325

mook-as:
[...]

PR #1217, opened 2020-08-13
Actions

prabalsharma:
Implements https://github.com/cloudfoundry-incubator/kubecf/issues/1144

- [x] should create clusters on GKE
- [x] github actions workflow that runs smoke, CATS, SITS and brain tests for PRs
- [x] runs diego
- [x] runs eirini
- [x] doesn't run for PRs from a fork
- [x] should cleanup any cloud resources
- [x] should not be mandatory for merges
- [x] GKE clusters should have only one node

Test: https://github.com/prabalsharma/kubecf/actions

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -166,6 +166,7 @@ jobs:
         AZURE_APP_ID: ${{ secrets.AZURE_APP_ID }}
         AZURE_CLUSTER_NAME: kubecf-ci-${{ github.run_id }}-${{ matrix.backend }}
         AZURE_DNS_RESOURCE_GROUP: ${{ secrets.AZURE_DNS_RESOURCE_GROUP }}
+        AZURE_LOCATION: ${{ secrets.AZURE_LOCATION }}
         AZURE_PASSWORD: ${{ secrets.AZURE_PASSWORD }}
         AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP }}-${{ github.run_id }}-${{ matrix.backend }}
         AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
@@ -305,6 +306,7 @@ jobs:
         AZURE_APP_ID: ${{ secrets.AZURE_APP_ID }}
         AZURE_CLUSTER_NAME: kubecf-ci-${{ github.run_id }}-${{ matrix.backend }}
         AZURE_DNS_RESOURCE_GROUP: ${{ secrets.AZURE_DNS_RESOURCE_GROUP }}
+        AZURE_LOCATION: ${{ secrets.AZURE_LOCATION }}
         AZURE_PASSWORD: ${{ secrets.AZURE_PASSWORD }}
         AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP }}-${{ github.run_id }}-${{ matrix.backend }}
         AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -15,6 +15,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         path: kubecf
+        fetch-depth: 0
         submodules: recursive
     - uses: actions/setup-python@v2
       with:
@@ -46,6 +47,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         path: kubecf
+        fetch-depth: 0
         submodules: recursive
     - uses: actions/setup-python@v2
       with:
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -69,9 +69,10 @@ jobs:
         name: kubecf-bundle.tgz
         path: ${{ github.workspace }}/bundle
 
-  test-diego:
-    name: test-diego
+  tests:
+    name: tests
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     needs: build
     env:
       # For KubeCF
@@ -86,6 +87,15 @@ jobs:
     defaults:
       run:
         working-directory: kubecf
+
+    strategy:
+      matrix:
+        include:
+        - backend: diego
+          experimental: false
+        - backend: eirini
+          experimental: true
+
     steps:
     - name: Checkout KubeCF
       uses: actions/checkout@v2
@@ -154,10 +164,10 @@ jobs:
       working-directory: catapult
       env:
         AZURE_APP_ID: ${{ secrets.AZURE_APP_ID }}
-        AZURE_CLUSTER_NAME: kubecf-ci-${{ github.run_id }}-diego
+        AZURE_CLUSTER_NAME: kubecf-ci-${{ github.run_id }}-${{ matrix.backend }}
         AZURE_DNS_RESOURCE_GROUP: ${{ secrets.AZURE_DNS_RESOURCE_GROUP }}
         AZURE_PASSWORD: ${{ secrets.AZURE_PASSWORD }}
-        AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP }}-${{ github.run_id }}-diego
+        AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP }}-${{ github.run_id }}-${{ matrix.backend }}
         AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
         AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
         OWNER: ${{ github.repository_owner }}
@@ -230,6 +240,7 @@ jobs:
     - run: make kubecf-apply
       env:
         CHART: ${{ github.workspace }}/kubecf-bundle/kubecf_release.tgz
+        FEATURE_EIRINI: ${{ matrix.backend == 'eirini' }}
         VALUES: dev/kubecf/kubecf-values.yaml
     - run: make kubecf-wait
 
@@ -265,7 +276,7 @@ jobs:
       if: always()
       uses: actions/upload-artifact@v2
       with:
-        name: diego-values.yaml
+        name: ${{ matrix.backend }}-values.yaml
         path: ${{ github.workspace }}/kubecf/dev/kubecf/kubecf-values.yaml
 
     - name: Fetch logs
@@ -278,7 +289,7 @@ jobs:
       if: always()
       uses: actions/upload-artifact@v2
       with:
-        name: diego-klog.tgz
+        name: ${{ matrix.backend }}-klog.tgz
         path: ${{ github.workspace }}/kubecf/klog.tar.gz
 
     - name: kubernetes:teardown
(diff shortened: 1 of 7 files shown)


.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+      with:
+        python-version: '3.x'
+    - name: cache.tools
+      uses: actions/cache@v2
+      with:
+        path: ${{ github.workspace }}/tools
+        key: ${{ runner.os }}-tools
+    - run: make tools-install
+
+    # Run the linters; split them out so it's easier to identify which one failed
```
viccuad, line 30: Could we use `make lint` to get new linters added to the pipeline for free?
If they don't get enumerated, I would argue that this functionality should be added to `make lint`.

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+        path: ${{ github.workspace }}/catapult/modules/common/bin
+        key: ${{ runner.os }}-catapult-common-tools
+
+    - run: make k8s
+      working-directory: catapult
+      env:
+        GKE_CLUSTER_NAME: kubecf-ci-${{ github.run_id }}-${{ matrix.backend }}
+        OWNER: ${{ github.repository_owner }}
+        DEBUG_MODE: false
+        GKE_NODE_COUNT: 1
```
viccuad, line 172:
```suggestion
        GKE_NODE_COUNT: 1
        GKE_PREEMPTIBLE: true
```

We should also set the cluster as preemptible, to mirror the concourse setup.

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+      env:
+        GKE_CRED_JSON: ${{ secrets.GKE_CREDS_JSON }}
+
+    - name: cache.catapult-common-tools
+      uses: actions/cache@v2
+      with:
+        path: ${{ github.workspace }}/catapult/modules/common/bin
+        key: ${{ runner.os }}-catapult-common-tools
+
+    - run: make k8s
```
viccuad, line 166: We could also separate the gke env vars that we are setting in `make k8s` onto it's own task as the `set GKE_CRED_JSON`. That way we could swap between providers easier in the future, just changing the `BACKEND` var.

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+      run:
+        working-directory: kubecf
+
+    strategy:
+      matrix:
+        include:
+        - backend: diego
+          experimental: false
+        - backend: eirini
+          experimental: true
```
viovanov, line 99: I think we shouldn't consider eirini experimental

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+
+    steps:
+    - name: Checkout KubeCF
```
viovanov, line 102: Can you please leave one empty row between each step?
If warranted even a comment helps read the yaml.

e.g.

```yaml

    # We need catapult to create Kubernetes clusters
    - name: Checkout catapult
      uses: actions/checkout@v2
      with:
        repository: SUSE/catapult
        path: catapult

    # We'll be caching foo and bar
    - name: cache.tools
      uses: actions/cache@v2
      with:
        path: ${{ github.workspace }}/tools
        key: ${{ runner.os }}-tools
```

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+
+    - id: download-bundle
+      name: Download chart bundle
+      uses: actions/download-artifact@v2
+      with:
+        name: kubecf-bundle.tgz
+        # The artifact upload/download creates an extra directory
+        path: ${{ github.workspace }}/kubecf-bundle
+
+    - name: Download gomplate
```
viovanov, line 130: This should be a dependency we download using `make <something>`.
@jandubois can help us do that.
Ideally, the code to download dependencies should live in `kubecf-tools`.

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+        if test -x "${TOOLS_DIR}/gomplate" ; then
+          echo "gomplate already exists; skipping."
+          exit 0
+        fi
+        mkdir -p "${TOOLS_DIR}"
+        wget -O "${TOOLS_DIR}/gomplate" \
+          https://github.com/hairyhenderson/gomplate/releases/latest/download/gomplate_linux-amd64
+        chmod a+x "${TOOLS_DIR}/gomplate"
+
+    - name: Start SSH agent
```
viovanov, line 141: Why do we need this? If for debugging, we should use the tmate action [1].

[1] https://github.com/marketplace/actions/debugging-with-tmate

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+        set -o errexit -o pipefail -o nounset
+        eval "$(ssh-agent -s)"
+        ssh-keygen -t rsa -b 4096 -N '' -C "KubeCF CI #${{ github.run_id }}" -f ssh-key
+        ssh-add ssh-key
+        echo "::add-mask::$(cat ssh-key.pub)"
+        rm -f ssh-key ssh-key.pub
+        echo "::set-env name=SSH_AUTH_SOCK::${SSH_AUTH_SOCK}"
+        echo "::set-env name=SSH_AGENT_PID::${SSH_AGENT_PID}"
+
+    - name: set GKE_CRED_JSON
```
viovanov, line 152: Please use the vault action [1] for these secrets with our vault deployment [2].
Please read the guide here on how to use a personal access token to login [3].
The `VAULT_TOKEN` is already available to kubecf.

[1] https://github.com/marketplace/actions/vault-secrets
[2] http://volt.cap.explore.suse.dev/
[3] https://www.vaultproject.io/docs/auth/github#github-auth-method

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
...
+        # at the end of the run.
+        cat "${KUBECONFIG}" | gzip | base64 --wrap=0
+      working-directory: catapult
+
+    - name: Add tools to path
+      run: |
+        echo "::add-path::${TOOLS_DIR}"
+
+    - name: Generate KubeCF Configuration
+      run: |
```
viovanov, line 191: We shouldn't generate `values.yaml` anymore.
Do you want to make this change here (seems easier to me), or in a different PR?

.github/workflows/pull-request-ci.yaml
mook-as: This was for debugging and we can drop it.

The tmate action isn't great because
- It's by some random guy; I've tried to limit the actions here to be first-party Microsoft ones (including GitHub).
- it hangs the CI until you connect and create a file.  That's not useful if we just want to be able to get in only if the run fails or looks suspicious.

.github/workflows/pull-request-ci.yaml
prabalsharma: Let us handle it in different PR. I think it is more important to get initial commits on GHA which will enable us for quick iterations to improve it.

.github/workflows/pull-request-ci.yaml
prabalsharma: commented. It is a requirement for catapult.

viovanov: @prabalsharma what does catapult use it for?

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,319 @@
...
+    name: lint
+    runs-on: ubuntu-latest
+    env:
+      PINNED_TOOLS: true
+      TOOLS_DIR: ${{ github.workspace }}/tools
+    defaults:
+      run:
+        working-directory: kubecf
+    steps:
+    - name: Checkout KubeCF
```
viovanov, line 14: Can you please add the comment / space for all steps in all jobs?

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,307 @@
+name: CI
+on: push
```
viovanov, line 2: I don't think this does what the description asks for.
Please see [1]. We need this to run for checking PRs, but not those that are opened from forks. Just from local branches.

[1] https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on

.github/workflows/pull-request-ci.yaml
viovanov: It's ok not to put it in `kubecf-tools`, but please add a new tool here: https://github.com/cloudfoundry-incubator/kubecf/tree/master/scripts/tools

.github/workflows/pull-request-ci.yaml
viovanov: Ok, can you please add an issue for it?

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,319 @@
...
+
+        {{- $ | toYAML }}
+        EOF
+      env:
+        BACKEND: gke
+        AUTOSCALER: "true"
+
+    - run: make cf-operator-apply
+      env:
+        CF_OPERATOR_URL: ${{ github.workspace }}/kubecf-bundle/cf-operator.tgz
```
viovanov, line 249: Why is this required?

.github/workflows/pull-request-ci.yaml
prabalsharma: Sorry, I cannot remember now, I will ask around.

.github/workflows/pull-request-ci.yaml
prabalsharma: Done https://github.com/cloudfoundry-incubator/kubecf/pull/1217/commits/9e0ba90cdfe9b72db1ac02f86d916c00002b384e

.github/workflows/pull-request-ci.yaml
prabalsharma: https://github.com/cloudfoundry-incubator/kubecf/issues/1219

.github/workflows/pull-request-ci.yaml
prabalsharma: https://github.com/cloudfoundry-incubator/kubecf/issues/1220

.github/workflows/pull-request-ci.yaml
mook-as: Yep, I did it on push in my branch for spiking; this will need to change before we actually merge.

.github/workflows/pull-request-ci.yaml
mook-as: https://github.com/SUSE/catapult/wiki/AKS#preparation documents the requirement (remember that I started with AKS).

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,361 @@
...
+        path: ${{ github.workspace }}/kubecf-bundle
+
+    # Start SSH agent for catapult
+    - name: Start SSH agent
+      run: |
+        set -o errexit -o pipefail -o nounset
+        eval "$(ssh-agent -s)"
+        ssh-keygen -t rsa -b 4096 -N '' -C "KubeCF CI #${{ github.run_id }}" -f ssh-key
+        ssh-add ssh-key
+        echo "::add-mask::$(cat ssh-key.pub)"
```
mook-as, line 180: I'm not sure I ever got the mask here to work correctly.

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,361 @@
...
+      env:
+        GKE_CLUSTER_NAME: kubecf-ci-${{ github.run_id }}-${{ matrix.backend }}
+        OWNER: ${{ github.repository_owner }}
+        DEBUG_MODE: false
+        GKE_NODE_COUNT: 1
+        GKE_PREEMPTIBLE: true
+        GKE_INSTANCE_TYPE: n1-highcpu-16
+
+    # Providing kubeconfig for debugging
+    - name: export KUBECONFIG
```
mook-as, line 213: _This_ was the thing I was referring to earlier about debugging.

.github/workflows/pull-request-ci.yaml
mook-as: Given that we already have it (because we have a bundle), downloading it again seems dumb.  This also ensures we don't accidentally use `latest`.

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,361 @@
...
+    # Wait for kubecf pods to be ready
+    - run: make kubecf-wait
+
+    # Run smoke tests
+    - run: make smoke-tests
+
+    # Run brain tests
+    - run: make brain-tests
+
+    # Run sits tests
```
mook-as, line 294:
```suggestion
    # Run sync integration tests
```
"Run sync integration test suite tests" is… technically valid, but rather verbose.  I'd also be fine with _removing_ the comments for the one-line commands, as they don't actually add any value.

.github/workflows/pull-request-ci.yaml
viovanov: I would expect `make cf-operator-apply` to install kubecf's dependency.
If it doesn't, it's a bug.

jandubois: > I would expect `make cf-operator-apply` to install kubecf's dependency.

It does `helm install cf-operator "$(cf_operator_url)" ...` and `cf_operator_url` will return the resolved URL from `dependencies.yaml`, i.e. `https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-{version}.tgz` with `{version}` replaced by the version also specified in `dependencies.yaml`.

.github/workflows/pull-request-ci.yaml
```diff
@@ -0,0 +1,361 @@
...
+      run: |
+        pwd
+        ls -a
+        echo ${{ github.workspace }}
+        ls -a ${{ github.workspace }}
+        echo $PATH
+        ls -a ${{ github.workspace }}/kubecf/output/bin/
+        echo "::add-path::${{ github.workspace }}/kubecf/output/bin"
+        echo $PATH
+        gomplate -v
```
mook-as, line 162: `add-path` won't be in effect until _after_ the current step.

.github/workflows/pull-request-ci.yaml
prabalsharma: done

.github/workflows/pull-request-ci.yaml
prabalsharma: done

Issue #1224, opened 2020-08-17
grootfs cleanup does not work correctly


univ0298:
**Describe the bug**
We observed that the contents of /var/vcap/data/store were not cleaned up when a cell pod restarts. This led to garden errors like `guardian.create.volume-creator.image-plugin-create.grootfs.create.groot-creating.making-image.overlayxfs-creating-image.applying-quotas.run-tardis.tardis.set-quota.set-quota.setting-quota-to-project-id-failed` , caused by stale information about projectids being carried over from a previous 'life' of the data directory from a previous pod. 

Looking at the bpm-pre-start-garden container logs we can see an error where the clean did not work correctly in the pre-start

**To Reproduce**
Restart a cell pod and then look at the contents under /var/vcap/data/grootfs of the replacement pod that starts. You will see contents that are older than the lifetime of the new pod, demonstrating that the cleanup did not succeed. You can also look at the bpm-pre-start-garden logs and you will see an error from the find command, saying "paths must precede expression". For example:

```
kubectl -n kubecf logs diego-cell-3 bpm-pre-start-garden
+ /var/vcap/jobs/garden/bin/bpm-pre-start
find: paths must precede expression: bin
Usage: find [-H] [-L] [-P] [-Olevel] [-D help|tree|search|stat|rates|opt|exec] [path...] [expression]
2020-08-17T16:58:16.630871991Z - running thresholder
2020-08-17T16:58:16.637805608Z - done
real	0m0.059s
user	0m0.026s
sys	0m0.010s
```

**Expected behavior**
After a pod restarts, there should be no error in the bpm-pre-start-garden log and the /var/vcap/data/grootfs contents should all be created since the pod restarted.

**Environment**
 - CF Version: 13.9
 - Component/Plugin Name
 - Others

**Additional context**
Here is the code in question: https://github.com/cloudfoundry-incubator/kubecf/blob/master/bosh/releases/pre_render_scripts/diego-cell/garden/jobs/patch_pre-start.sh#L18

and here is the change that added this grootfs cleanup: https://github.com/cloudfoundry-incubator/kubecf/commit/3d0562919480c6b40bf09a21221f625f72d46e0c

slack discussion: https://cloudfoundry.slack.com/archives/CQ2U3L6DC/p1597411497472300

viovanov:
should be fixed by #1264

PR #1225, opened 2020-08-17
Correct find syntax to fix cleanup

univ0298:
The find syntax is incorrect for the wildcard, causing cleanup to fail - https://github.com/cloudfoundry-incubator/kubecf/issues/1224

## Types of changes

## Checklist:

### DIFF
--- a/bosh/releases/pre_render_scripts/diego-cell/garden/jobs/patch_pre-start.sh
+++ b/bosh/releases/pre_render_scripts/diego-cell/garden/jobs/patch_pre-start.sh
@@ -15,7 +15,7 @@ fi
 # Patch the pre-start script to setup /var/vcap/data
 patch --verbose "${target}" <<'EOT'
 2a3,4
-> find /var/vcap/data/grootfs/ -iname * -delete
+> find /var/vcap/data/grootfs/ -iname '*' -delete
 > 
 20a23,25
 > 


jandubois:
I'm curious why the script is using `find` to delete the directory to begin with:

```
find /var/vcap/data/grootfs/ -iname '*' -delete
```

Why is it not simply

```
rm -rf /var/vcap/data/grootfs/
```

@mook-as Do you have any idea?
@viovanov Is this somehow related to the PVC mount problem?

univ0298:
@jandubois yes and in fact we located code in SCF where it is an `rm -rf` : https://github.com/SUSE/scf/blob/master/container-host-files/etc/scf/config/scripts/cleanup-garden-graph.sh#L4

jandubois:
Yeah, the `find` method was added by @viovanov in 3d05629 as part of #489 (which I suspected when I asked above about the PVC mount issues).

In https://github.com/cloudfoundry-incubator/kubecf/pull/489#issuecomment-600383822 I even quote the old SCF code that does the `rm -rf`, so I'm not sure whay Vlad changed it.

One difference is that the `find` code also removes the `grootfs` directory itself, whereas the `rm -rf` only removes all files (but not hidden files) from within. But if the goal was to removed everything include the dir itself, then you only have to strip the trailing `*` from the `rm -rf` command, so I still don't understand.

We need to wait until @viovanov can chime in.

viovanov:
Unfortunately I don't remember. It's possible I started with a pattern, and it evolved to `*`.
If the new syntax works, I'd leave it as is.

viovanov (approved):

jandubois:
> If the new syntax works, I'd leave it as is.

I agree that it is not terribly important, but if the more obvious `rm -rf ...` works as well, then I would prefer to merge the simpler form. Otherwise people will wonder if they ever look at this why the complex expression is necessary.

I guess it depends on @univ0298's willingness to try and test the `rm -rf ...` form as well. Otherwise, I'm also fine with merging as-is.

univ0298:
We could try the rm approach fairly easily I think. But it seems quite surprising that @viovanov would have switched from rm to find, without any reason

jandubois:
> We could try the rm approach fairly easily I think.

I would appreciate if you could do that.

> But it seems quite surprising that @viovanov would have switched from rm to find, without any reason

I agree. Which makes it even more important to add a comment to explain why this command is necessary, and how it is different from a plain `rm -rf`.

I don't think it is though: The term `-iname *` doesn't make much sense, it should be `-name *`, but that still is a condition that is *always* true, so you could just write `find /var/vcap/data/grootfs/ -delete`. And now it is even less clear how this could possibly be different from `rm -rf /var/vcap/data/grootfs/`.

I suspect that this was just a random/desperate change (that didn't actually make any difference) while trying to get the nested PVC mounting to work, which took several days. And when it was finally working, the inconsequential changes just remained in place to avoid having to go through even more testing cycles.

univ0298:
@jandubois I tried with the `rm -rf` alternative and it also worked just fine, so you could amend to that. I simply remain a bit suspicious that @viovanov had some mysterious good reason to deviate away from that in the first place! But it seems fairly clear that rm would be equally safe and cleaner

PR #1228, opened 2020-08-18
PR triggering on code review

viovanov:
This should work for forks as well.

### DIFF
--- a/.github/workflows/pull-request-queue.yaml
+++ b/.github/workflows/pull-request-queue.yaml
@@ -1,0 +1,18 @@
+name: enqueue-pr-test
+on:
+  # When a PR is approved, add the queue label
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  enqueue:
+    if: github.event.pull_request.base.ref == 'master' && github.event.review.state == 'approved'
+    name: enqueue
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions-ecosystem/action-add-labels@v1
+        with:
+          # We need a bot token because the regular GH Actions token is not allowed to trigger
+          # secondary GH Actions events.
+          github_token: ${{ secrets.GH_BOT_TOKEN }}
+          labels: pr-test-queue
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request-queue.yaml
```diff
@@ -0,0 +1,18 @@
...
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  enqueue:
+    if: github.event.pull_request.base.ref == 'master' && github.event.review.state == 'approved'
+    name: enqueue
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions-ecosystem/action-add-labels@v1
```
mook-as, line 13: Would it make sense for us to fork this somewhere where we know it won't change under us?  Depending on [some random guy](https://github.com/orgs/actions-ecosystem/people) doesn't seem awesome.

Alternatively, I think [github-script](https://github.com/actions/github-script#apply-a-label-to-an-issue) would work too — and that one is first-party.

viovanov: I don't know if it makes sense to fork.
Is this any different from depending on something like [1] or [2]?

[1] https://github.com/bmatcuk/doublestar
[2] https://github.com/FooBarWidget/multipart-parser

mook-as: Both of those are libraries (go and C++, respectively), which means they can be sensibly vendored.  In GitHub actions that really isn't a thing (though we _could_ pin to a commit; that seems fine too).

viovanov: Seems like a slippery slope to assume maliciousness at this stage.
A version can be replaced for go libraries as well, and I would consider that more of a problem than this.

mook-as: Go libraries only update when we do some action (bump things in `go.mod`); GitHub Actions versions are pulled when they run.  Note that Actions uses _branch names_, not tags, for versioning.

viovanov: Fair enough 👍, I will try pinning to a commit ID.

PR #1242, opened 2020-08-20
feat: add validation label so it can be easily identified before moved to a project

fargozhu:
## Description

To support the early identification through filters when a new issue is opened and to avoid orphans issues for too long.

## Motivation and Context

I need to be constantly searching for "no:project" issues and this can't be applied to Github filters.

## How Has This Been Tested?

Will check by opening a new issue .


fargozhu:
> We may want to start having documentation around the multitude of labels we're getting these days.

Github labels must contain a self explanatory description, if not then we need to enforce that.

PR #1243, opened 2020-08-21
Remove "version: every" when "passed:" contraints are defined

jimmykarily:
because that blocks automatic triggering of jobs

https://github.com/concourse/concourse/issues/736

This combination of settting was introduced when we tried to force
Concourse to run later jobs on every version that passed the previous
one. E.g.
ver1 passed job 1, ver2 passed job 1
job2 should trigger both for ver1 and ver2 otherwise ver1 would never
be tested on job2. The new scheduling algorighm may have solved this:

https://concourse-ci.org/scheduler.html#scheduling-behavior

Given our pipeline needs us to manually trigger jobs all the time, this
worths a shot.

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -391,7 +391,6 @@ jobs:
     - {{ . -}}{{ end }}
     {{- end }}
     trigger: true
-    version: "every"
     passed:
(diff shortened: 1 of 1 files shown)


jimmykarily:
See "Breaking change" here: https://github.com/concourse/concourse/releases/tag/v6.0.0

PR #1258, opened 2020-08-26
feat: Implement default memory limits for kubecf

### LINKED ISSUE #955: Expose <role>.cpu|memory.request|limit in values.yaml
In scf it was possible to set memory and CPU requests and limits for each CF role via helm. It's not clear if this is possible in KubeCF. This needs to be exposed.

If this is possible currently, there should be some indication in the Helm chart `values.yaml` how these can be set.

### LINKED ISSUE #1142: Implement default memory limits
Right now it is possible to override memory limits on a per-container basis, like this:

```
operations:
  inline:
  - type: replace
    path: /instance_groups/name=log-cache/jobs/name=log-cache/properties/quarks?/bpm/processes/name=log-cache/limits/memory
    value: 2G
```

But it is impractical to expect the user to know the list of instance groups and jobs, that needs to be configured, and what a sensible memory limit would be for each.

This request if for:

[ ] collect typical memory requirements for each job, verify them against possible defaults in the `bpm.yml` files, and then include them in kubecf.

[ ] Provide a mechanism in kubecf to globally enable memory limits with the default settings

This request does not cover disk size limits. They are not supported by the bpm model, and not implemented by cf-operator.

## Acceptance Criteria

KubeCF should start and run CATS normally within these limits.
Collect peak values from SUSE's sandbox and add 50% to that.

andreas-kupries:
## Description

  - Extended schema for memory information and group/job conditions.
  - Extended schema for new feature flag controlling use of memory info.
  - New feature flag (s.a.) for user configuration.
  - Placeholder config files for memory and condition information.
  - First set of igroups and jobs for conditions.
  - First set of memory limits from the sandbox.

## Motivation and Context

Fixes #1142 
Fixes #955

## How Has This Been Tested?

Local minikube, in various configurations.

## Types of changes

## Checklist:


chart/config/conditions.yaml
```diff
@@ -0,0 +1,125 @@
+# Specify the feature flag conditions applying to instance groups and
```
viovanov, line 1: I don't think this paragraph explains the purpose of the file properly.
Why is it required? Why are we doing conversions?

chart/config/memory.yaml
```diff
@@ -0,0 +1,123 @@
...
+##
+# scheduler:
+#       loggr-syslog-binding-cache: 18Mi
+
+# # ## ### ###### ######## #############
+
+limits:
+  memory:
+    api:
+      cc_uploader: 16Mi
```
viovanov, line 38: Seems small - did we collect values from the sanbox and increase by 50% as mentioned in #1258 ?

chart/values.schema.yaml
```diff
@@ -322,6 +322,31 @@ properties:
...
+  instance_group_condition:
+    type: object
+    additionalProperties:
+      type: object
+      additionalProperties:
+        oneOf:
+        - { type: boolean }
+        - { type: string }
+
+  limits:
```
viovanov, line 334: Please use the `sizing.` key.
e.g. `sizing.diego-cell.limits.memory`.

dev/kube/ig-tools/snap-resources.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
viovanov, line 1: Do you think this + max memory could live in kubecf tools?
Other projects could benefit.

dev/kube/ig-tools/yaml-memory.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
```
viovanov, line 1: Does this need to be a separate file? Why not include it in max-memory? Is max-memory output useful without this yaml converter?

viovanov (changes requested):

chart/config/conditions.yaml
andreas-kupries: Will work on better text.

chart/config/memory.yaml
andreas-kupries: In one of the earlier commits I mentioned that these are from my minikube, and without padding.
Getting proper data from the sandbox and pad is in the TODO (See PR description, between the construction warnings).

chart/values.schema.yaml
andreas-kupries: This changes the structure in a way which makes it impossible (I believe) to have default values (per igroup, and global). Also note that your structure limits sizing to the igroup/pod level ... The current structure (ignoring limits vs sizing) easily allows for defaults, and allows limits at the job/container level.

```
limits:
  memory:
    '$default': 10Mi
    api:
      '$default': 100Mi
      cloud_controller_ng: 2Gi
```

whereas with

```
sizing:
  diego-cell:
    limits:
      memory: 4Gi
```

slotting in defaults looks to much harder, if not impossible.

dev/kube/ig-tools/snap-resources.sh
andreas-kupries: Yes, it could ... And agree, it should.

dev/kube/ig-tools/yaml-memory.sh
andreas-kupries: The separation allows me to
  - feed something manually written between max detection and conversion steps.
  - And another transformation too, must importantly a transform taking the max data and padding it.

chart/values.schema.yaml
jandubois: > slotting in defaults looks to much harder, if not impossible.

It could work the same way, but I agree that it looks convoluted because we have the additional job level to consider too:

```
sizing:
  '$defaults':
    limits:
      memory: 10Mi
  api:
    limits:
      memory:
        '$default': 128Mi
        cloud_controller_ng: 2Gi
```

I'm especially unhappy about having `$default` and `$defaults` properties, and how they have different internal structure. But the nesting is also getting too deep.

Of course the defaults are mostly in the internal config file as a user override would not bother with defaults normally and just specify e.g.

```
sizing:
  api:
    limits:
      memory:
        cloud_controller_ng: 4Gi
```

I still can't really make up my mind: What is more important, seeing all the sizing/limits information for a specific instance group in one place, or seeing all the memory limits grouped together in one place, and all the instance counts in another?

Conceptually seeing all the information about the pods in one place makes sense. But when I'm looking at the memory requirements of the cluster, having all the limits close together is more helpful than having them scattered between all the other settings.

But I think we need to settle on one ordering mechanism: if we have a separate `limits` structure, then maybe we should split `sizing` into multiple trees too? Why combine instance counts and affinity settings? But then, `limits` should not combine `memory` and `cpu` limits either; otherwise why not fold it into `sizing`?

One other difference is that `sizing` stops at the pod (instance group) level, while `limits` are applied to the container (job).

After writing all this I'm even less able to decide which one feels best...

jandubois: Oh, one more thought: maybe having per-instance group defaults doesn't really make much sense. You just have one global default, and then have to override any job that needs extra memory.

I know I suggested this originally, but now I can't think of an actual use case where setting different per-pod limits is required.

andreas-kupries: I am ok with just having a global default.

viovanov: Using the bosh manifest deployment as an example, more often than not, it's useful to see all the information for one instance group in one place. Things that make you have to change context in the yaml file (like addons or variables) are painful. As an analogy, even though the variables are defined as an array in the deployment manifest, I never care about looking at them at-a-glance. It's always their context that matters.

If you need info grouped by limits, or disk size, or privileges, you should be able to script it easily.

---

For the defaults - can you please explain why per-ig defaults are useful?
The containers in a pod have very different memory needs - I think each individual container should have a value.

dev/kube/ig-tools/yaml-memory.sh
viovanov: Ok - maybe call it something else then? like `max-memory-as-yaml.sh` ?
I think `yaml-memory` is not very descriptive, plus there's also `max-memory`, which makes it more confusing.

chart/config/conditions.yaml
viovanov: Thanks!

chart/values.schema.yaml
andreas-kupries: So, based on that 

```
sizing:
  '$defaults':
    limits:
      memory: 10Mi
  api:
    limits:
      memory:
        cloud_controller_ng: 2Gi
```

or 

```
sizing:
  '$defaults':
    limits:
      memory: 10Mi
  api:
    cloud_controller_ng:
      limits:
        memory: 2Gi
```

I prefer the second form (`group/job/'limit'/'memory'`) over (`group/'limits'/'memory'/job`).

andreas-kupries: > please explain why per-ig defaults are useful

They came from Jan, and he now has no idea either regarding their usefulness.

dev/kube/ig-tools/snap-resources.sh
andreas-kupries: Note: The `yaml-memory.sh` is specific to `kubecf` and its pods/containers. Should stay here. Also a reason to keep it separate from the `max memory`, which is generic.

dev/kube/ig-tools/yaml-memory.sh
andreas-kupries: Will work on a better name. Some of the confusion will be gone with the other scripts moving to `kubecf-tools`.

chart/values.schema.yaml
jandubois: > I prefer the second form (`group/job/'limits'/'memory'`) over (`group/'limits'/'memory'/job`).

That doesn't really work as there are other properties below the group already:

```
sizing:
  '$defaults':
    limits:
      memory: 10Mi
  api:
    instances: 3
    affinity: { ... }
    limits:
      memory:
        cloud_controller_ng: 2Gi
```

So you would still need something like this:

```
sizing:
  '$defaults':
    limits:
      memory: 10Mi
  api:
    instances: 3
    affinity: { ... }
    jobs:
      cloud_controller_ng:
        limits:
          memory: 4Gi
      cloud_controller_clock:
        limits:
          memory: 512Mi
```

(yes, I know `cloud_controller_clock` is in the `scheduler` pod, just an example of setting 2 limits for one group).

So as soon as you set multiple limits I think grouping limits together works better:

```
sizing:
  '$defaults':
    limits:
      memory: 10Mi
  api:
    instances: 3
    affinity: { ... }
    limits:
      memory:
        cloud_controller_ng: 4Gi
        cloud_controller_clock: 512Mi
```

Of course once you add in CPU limits and maybe CPU and memory requests it all becomes a toss-up again:

```
sizing:
  '$defaults':
    limits:
      cpu: 2
      memory: 256Mi
    requests:
      memory: 64Mi
  api:
    instances: 3
    affinity: { ... }
    limits:
      cpu:
        cloud_controller_ng: 4
      memory:
        cloud_controller_ng: 4Gi
        cloud_controller_clock: 512Mi
    requests:
      memory:
        cloud_controller_ng: 2Gi
        cloud_controller_clock: 256Mi
```

I think I still slightly prefer this grouping, as you may need/want not set non-default cpu limits for the same jobs that you want to specify non-default memory limits for.

dev/kube/ig-tools/snap-resources.sh
andreas-kupries: See https://github.com/cloudfoundry-incubator/kubecf-tools/pull/8

andreas-kupries: See f96ab53a9b3968104baa30518261ab915377866b for shift of the generics.

dev/kube/ig-tools/yaml-memory.sh
andreas-kupries: See f96ab53a9b3968104baa30518261ab915377866b for new name (`make-memory-config.sh`).

chart/config/conditions.yaml
andreas-kupries: See fc3d09598586e1dfc5fae74dda9bbc488fddd2bc
Note, the referenced file  `../assets/operations/memory.yaml` does exist yet.
Well, while I have something locally, it is not in a state to be committed.
In part due to the memory structure still in flux.

chart/config/conditions.yaml
```diff
@@ -0,0 +1,131 @@
...
+    loggr-udp-forwarder: true
+  routing-api:
+    routing-api: features.routing_api.enabled
+  scheduler:
+    cc_deployment_updater:        true
+    cloud_controller_clock:       true
+    loggr-syslog-binding-cache:   true
+    loggr-udp-forwarder:          true
+    service-discovery-controller: true
+    ssh_proxy:                    !features.eirini.enabled
```
jandubois, line 119: The value must be quoted here because `!foo` is YAMl syntax for a local tag called `foo`. The actual value of `ssh_proxy` in the current form is `null`.

chart/config/conditions.yaml
```diff
@@ -0,0 +1,131 @@
...
+    statsd_injector:              true
+    tps:                          !features.eirini.enabled
+  singleton-blobstore:
+    blobstore:       features.blobstore.provider == singleton
```
jandubois, line 123: While technically correct, seeing complex strings with punctuation and whitespace without surrounding quotes always makes me nervous. I would prefer if we always quoted condition expressions.

jandubois (dismissed):

chart/values.schema.yaml
andreas-kupries: Ah. I forgot/missed that we already have `sizing`, from `values.yaml`. :facepalm:
Ok. Now reading the remainder of this comment.

If @viovanov is ok with this last proposed structure I am ok with it as well.

chart/config/conditions.yaml
andreas-kupries: Note that I am not even sure if this kind of condition works with your `_config.condition` macro. There is a question at the top of the file about that.

chart/config/conditions.yaml
andreas-kupries: Will fix.

chart/config/conditions.yaml
jandubois: BTW, forgot to add: the condition as written is not currently supported by the `_config.condition` evaluator:

1. There is no comparison operator `==`; the only implemented operations are grouping `()`, negation `!`, and `&&` and or `||`, in the precedence order listed here. So every operand is always treated as a boolean.

2. There is no support for literal strings (because they would all be considered `true`, except for the empty string being `false`. The term `singleton` would currently be evaluated as the value of `$.Values.singleton`, which doesn't exist, so would be considered `true`.

Now, I'm all for extending the `_config.condition` mechanism as needed, but right now we should look how these features are being defined. Normally there should be a `feature.external_database.enabled` property or something like that. This is an issue I want to open, for us to create a more consistent structure for enabling/disabling features, and how to configure them. Once we do that, the current `_config.condition` mechanism may still be sufficient for the current use-case. Revising the feature settings is outside the scope of this PR though.

andreas-kupries: Done with commit 7eb9b9ff3ea2e0cf8f54eab43f2d08fc5d0ee875

chart/config/conditions.yaml
andreas-kupries: Done with commit 7eb9b9ff3ea2e0cf8f54eab43f2d08fc5d0ee875

chart/config/conditions.yaml
andreas-kupries: I agree. The `features.blobstore.provider` currently supports only `s3` and `singleton`, and the difference is external vs local.
The provider tests use either string.

I am all in for switching to a proper boolean `features.external_blobstore.enabled` and then updating the tests using`features.blobstore.provider` to that. I would retain the features.blobstore` hierarchy tough, because that contains the S3 configuration when `provider == s3` i.e. external. And I do not wish to move that around. That said, if you consider movign better, I can do that too.

@viovanov Do you agree to make this switch ? I would do it as part of this PR then.

chart/values.schema.yaml
viovanov: This last one looks ok - but I still don't think the defaults are useful. e.g. 256Mi for the CC is not enough, but too much for log cache, or nats or other such things. If we already have real-world numbers, why bother with the defaults?

```
sizing:
  '$defaults':
    limits:
      cpu: 2
      memory: 256Mi
    requests:
      memory: 64Mi
  api:
    instances: 3
    affinity: { ... }
    limits:
      cpu:
        cloud_controller_ng: 4
      memory:
        cloud_controller_ng: 4Gi
        cloud_controller_clock: 512Mi
    requests:
      memory:
        cloud_controller_ng: 2Gi
        cloud_controller_clock: 256Mi
```

andreas-kupries: Ok, how about leaving out the defaults for now, and if it comes to it us wanting them later on it should be no trouble to add them based on the above.

chart/config/conditions.yaml
jandubois: I would not want to modify how `blobstore` is configured as part of a PR that adds memory limits to containers. If we can't make it work without changing the config settings, then maybe this PR needs to be blocked until the other change has been done.

That said, I really don't want to change our settings piecemeal over time; there should be a single release that moves everything around to the proper place, with safeguards in `unsupported` and tight schema restrictions, but afterwards things should be rather stable. This is our external API, after all.

So I think this needs to be thought through comprehensively, and not one feature at a time. I will create a separate issue for this soonish.

About re-arranging blobstore config settings, there is an example of how I thought we should be doing this in https://github.com/cloudfoundry-incubator/quarks-operator/issues/895#issuecomment-616817788 under the `Example` heading.

I'm not sure if I still agree with my previous suggestion, but that is also something that I would want to at least consider before making any changes.

viovanov: @andreas-kupries yes - let's do it
who doesn't like consistency? 😄

andreas-kupries: @jandubois  @viovanov 

> can't make it work without changing the config settings

Is each config file run through regular helm templating as part of its integration ?
Because if yes we could maybe use something like

```
{{- if eq $.Values.features.blobstore.provider "s3" }}
features:
  external_blobstore:
    enabled: true
{{- else }}
features:
  external_blobstore:
    enabled: true
{{- end}}

to generate this new bool key, from a config file named to be very early in the operation.

Alternatively, put this into the config support code, no config file required.

If that works, we can then handle the restructuring of the blobstore setup wholly separate, with only the config code proposed above the overlapping point (the restructure would remove it again).

File: chart/values.schema.yaml (hunk 3 above)
Comment (jandubois) [reply chart/values.schema.yaml:334]: > I still don't think the defaults are useful. e.g. 256Mi for the CC is not enough, but too much for log cache, or nats or other such things.

The numbers here are just placeholders, but remember they are `limits`, not `requests`.

They exist to for the benefit of *other* pods on the same node, in case any of our containers has a memory leak. Just like an insurance policy, we ideally want non of our limits to ever get triggered, we just want to prevent one container from consuming all memory on the node, essentially performing a DoS attack on the other pods.

So to me it totally makes sense to specify a pretty generous limit as the default, maybe even 512Mi, and only specify larger limits for the few containers that absolutely need it (e.g the CC and the Java components like credhub and UAA).

If NATS only uses 128Mi, it doesn't matter at all that we specify a limit of `512Mi`, it will never trigger, so it behaves exactly the same as a limit of `256Mi`. or `160Mi`. It only means we won't kill NATS if it happens to unexpectedly use `320Mi`, but I don't consider that an issue either.

Again these are limits, not requests, so no memory is "reserved" for it; it just is the number at which the OOM killer will enter a container.

So I feel having just a shorter list of the exceptionally large limits is more maintainable than a list that includes every single container. Remember that we will then have to update the list whenever a job is added anywhere in the system.

Finally, I wonder how all this applies to init containers. I think we need a mechanism to tell Quarks to set some kind of limit on them as well.

File: chart/config/conditions.yaml (hunk 7 above)
Comment (jandubois) [reply chart/config/conditions.yaml]: > Is each config file run through regular helm templating as part of its integration ?

No, it is not; it is loaded verbatim, just like `values.yaml`. I think it may be possible to add templating to config files, but I would have to look into it first before I can confirm for sure.

Comment (andreas-kupries) [reply chart/config/conditions.yaml]: Don't bother with templating for config then. ... Should be possible to do that i nthe memory templating to come, with dynamic modification of the `.Values` tree, similar to what you do when loading the config files.

File: chart/values.schema.yaml (hunk 3 above)
Comment (andreas-kupries) [reply chart/values.schema.yaml:334]: No idea if it applies to init containers at all. I suspect not.

Comment (jandubois) [reply chart/values.schema.yaml:334]: > No idea if it applies to init containers at all. I suspect not.

I didn't think so either. This was a comment for @viovanov that we may need additional support for this. Because init containers run pre-start scripts, which could leak memory and never terminate... So as an admin, I may still want to prevent unlimited growth.

**Comment by jandubois** [comment]
# About `config/conditions.yaml`

I think the file should be called `config/jobs.yaml` because it contains a list of all jobs that are being deployed by kubecf. This is similar to `config/releases.yaml`. The fact that jobs might have a condition to enable/disable them is secondary (and shared with other config data, like releases). The primary function is to provide the canonical list of jobs (per instance group).

I might also change the top property name from `instance_group_condition` to just `jobs`, to match the filename and primary purpose.

Further, a lot of the data should derived from `cf-deployment.yml` and not maintained manually. The manifest may add or drop jobs, or move them to a different instance group during a bump. So maintaining the list manually is burdensome.

To mirror the mechanism `config/releases.yaml` is augmented with `cf-deployment.yml` information, you should create a file `templates/_jobs.tpl` with a `_jobs.update` defintion that will be called from `_config.load` in `templates/_config.tpl`. Check out `templates/_releases.tpl` for things to copy.

So, for reference I would assume `config/jobs.yaml` to look like this:

```
jobs:
  diego-cell:
    '$default': '!features.eirini.enabled && !features.multiple_cluster_mode.control_plane.enabled'
  asmetrics:
    '$default': 'features.autoscaler.enabled'
     eventgenerator: ~
     metricsserver:  ~
# ...
```

Then in `templates/_jobs.tpl`:

```
{{- define "_jobs.update" }}
  {{- $_ := include "_config.lookupManifest" (list $ "instance_groups") }}
  {{- range $mf_ig := $.kubecf.retval }}
    {{- if not (hasKey $.Values.jobs $mf_ig.name }}
      {{- $_ := set $.Values.jobs $mf_ig.name dict }}
    {{- end }}
    {{- $ig  := index $.Values.jobs $mf_ig.name }}
    {{- range $mf_job := $mf_ig.jobs }}
      {{- if not (hasKey $ig $mf_job.name }}
        {{- $_ := set $ig $mf_job.name nil }}
      {{- end }}
    {{- end }}
  {{- end }}
  {{- range $ig := $.Values.jobs }}
    {{- $default := index $ig "$default" }}
    {{- $_ := unset $ig "$default" }}
    {{- range $job, $_ := $ig }}
      {{- if kindIs "invalid" }}
        {{- $_ set $ig $job $default }}
      {{- end }}
      {{- /* evaluate condition to a boolean here */}}
      {{- $_ := set $ig $job (eq "true" (include "_config.condition" (list $ (index $ig $job)))) }}
    {{- end }}
  {{- end }}
{{- end }}
```

(I've just typed this into the Github web UI, so it probably doesn't work without some fixes).

I've included code above to also evaluate the condition right away, so the `assets/operations/jobs.yaml` templates don't have to deal with it. I'm not totally sure if this is the right thing (and haven't done that for "releases" previously), so maybe leave that part out.

I think the same mechanism should also be used to propagate defaults through the `sizing` structure should be used, so that templating code doesn't have to deal with it. Once templates are expanded, the data structures should all be completely filled in and normalized. The `$default` meta keys should be removed, so that looping over the maps only generates valid entities.

**Comment by andreas-kupries** [comment]
Currently I only have a comment about this:

> Further, a lot of the data should derived from cf-deployment.yml and not maintained manually. The manifest may add or drop jobs, or move them to a different instance group during a bump. So maintaining the list manually is burdensome.

The difficulty there is that we (`kubecf`) are doing our own shuffling of jobs as well.
See `chart/assets/operations/job_moving/` for the ops files doing that.
This touches the groups `api`, `scheduler`, `doppler`, and creates `auctioneer`, `routing-api`, `log-api`.

It was this which caused me to restart redoing the conditions file using the memory file as base, because that was created from the resource capture, and thus contained the exact group/job associations in the active cluster, after all shuffling applied.

So, overall I would say that `data should be derived from cf-deployment.yml` is out of scope for this PR and ticket, and have a separate ticket/spike for work on the automation needed by that.

That said, on the other hand, reading closer, is the `_jobs.update` what you propose to derive things from `cf-deployment.yaml` ?
```
{{- define "_jobs.update" }}
  {{- $_ := include "_config.lookupManifest" (list $ "instance_groups") }}
```
These lines seem to indicate that, i.e. looking in the `instance_groups` key of the cf-deployment.yaml ?
Third hand, the information does not seem to be used, or I do not have a proper mental model of this yet.
Could this actually be meant to load the `conditions.yaml`/`jobs.yaml`, and then iterate over that tree, resolving the conditions ?

Ah! The `retval` is the clue. The first loop indeed iterates over the instance groups from cf-deployment, and creates entries in the `jobs` tree for all groups not found in it, and further adds entries for jobs missing in the tree, with a `nil` condition`.

That way I have to declare only the groups which have non-true conditions ? (There does not seem to be a global default to use for groups missing wholesale.)

I am not sure if that works with our shuffling of the jobs I mentioned at the top.

It sort of depends on the order of things, i.e.

  - when are the static ops files applied (moving jobs, dyn created them),
  - when does the templating to resolve the in-memory structure (generating more ops (memory config)
  - when do we apply the ops for the memory configs

Right now my mental model is:

  - The manifest is an asset file.
  - The templating reads this into memory (as it is yaml) and uses the information inside as part of building the `.Values` in-memory structure
  - This further generates all the memory ops instruction
  - After templating we have the manifest, plus all the ops files, neatly wrapped in k8s object (config maps, secrets, etc).
  - Then the operator takes over, and applies all the ops files
  - At this point the kubecf-specific shuffling of groups and jobs happens
  - I have no idea what order is imposed ops instructions.
  - That means I have no idea if the memory limit ops are applied before or after jobs are moved around.

  - If before, then your proposal can work, with the memory limits applied to group+job before the move, and then getting moved.
   __Or not__. The move ops reference asset files to get the job definitions, and these asset files are (I believe) extracted from cf-deployment.yaml as part of creating the helm chart. IOW these definitions will not contain any memory limit information.

   - If after, then the code making the ops instruction has to know which of the jobs are moved so that it can put in the name of the group using it post-move, instead of where it was defined.

----

Here are more complications:

The `database` pod and container is managed by a sub chart. The relevant group is deleted wholesale.
I have still not thought deep enough to see how we set these memory limits.
This also applies to the eirini pods.

The autoscaler groups are again like with the moved jobs. Dynamically created.

----

Next comment:

Having started to run this through `make kubecf-build` as a means of applying everything and finding all the bugs in the existing templating I have to ask.

Do we have scripts helping with debugging (outside of helmlint), for example something which ingests just config files, then dumps the modified internal structures back to a file for inspection, etc. pp. ?

**Comment by jandubois** [comment]
> The `database` pod and container is managed by a sub chart. The relevant group is deleted wholesale.
> I have still not thought deep enough to see how we set these memory limits.
> This also applies to the eirini pods.

Yes, I'm aware, but forgot to mention it. The `jobs` property only applies to bosh releases, and `assets/operations/limits.yaml` can only setup ops to add the limits for these bosh jobs. Native containers should still be configured via the `sizing` property, but their templates need to "manually" include these settings in their Pod/Stateful Set/Deployment specs.

> The autoscaler groups are again like with the moved jobs. Dynamically created.

These are bosh releases, so they should be handled like all the others. Except there is no deployment manifest for them, so you will have to enumerate the jobs and instance groups manually. Keep their settings grouped together and separate in `config/jobs.yaml`, as this data will eventually be moved to a separate config file when autoscaler is moved to its own directory (similar to the subcharts for eirini etc; I'm working on the scaffolding for this right now.

**Comment by jandubois** [comment]
> Do we have scripts helping with debugging (outside of helmlint), for example something which ingests just config files, then dumps the modified internal structures back to a file for inspection, etc. pp. ?

There is no such tooling so far. I've been running something like this when working on the multi-stack work:

```
make kubecf-build
helm template output/tmp/helm --set system_domain=x | less
```

The templates normally emit some kind of unique comment and I then defined a local keyboard macro so that I can type `///` in `less` and it will expand to e.g. `/default_stack↩` to move me to the section I'm looking for...

**Comment by jandubois** [comment]
> It sort of depends on the order of things, i.e.
> 
> * when are the static ops files applied (moving jobs, dyn created them),

The ops are grouped together in configmaps, and the order in which they are applied is defined in
https://github.com/cloudfoundry-incubator/kubecf/blob/master/chart/templates/bosh_deployment.yaml#L32-L79

> * when does the templating to resolve the in-memory structure (generating more ops (memory config)

Helm will load all `templates/_*.tpl` first. None of those files should generate any output, but just define templating functions.

All templates `templates/*.yaml` are supposed to start with this line:

```
{{- $_ := include "_config.load" $ }}
```

They don't right now, but will once my current PR is done, reviewed, and merged. That function will merge all `config/*` files into `$.Values` and then invoke the various `_foo/update` templates to fill in data from the manifest and apply defaults, so that the actual templates can be as simple as possible.

> * when do we apply the ops for the memory configs

We probably want to do this before jobs are being moved around, so you should put the ops into a subdirectory, create a separate configmap for them in https://github.com/cloudfoundry-incubator/kubecf/blob/master/chart/templates/ops.yaml and then list them in the `bosh_deployment.yaml` before the moving ops.

**Comment by andreas-kupries** [comment]
Rebased and rebooted. In debugging now.

**Comment by andreas-kupries** [comment]
What is now pushed
- Fixes issues in the `_memory.update` function leaving settings behind.
- It suffers from cloudfoundry-incubator/quarks-operator/issues/1114, where the applied limits cause requests as well. (That is bad for small deployments like minikube. It uses normally less memory than the limits, and fits, but with the request it will not fit anymore.
- For a number of pods/containers the limits are still not applied (nats/nats, the moved jobs, ...)

For the moved jobs I understand why. The move-code creates the job yaml completely new, using the cf-deployment info. This wipes out the previous tree with the limit assignment.

I suspect that something similar is going on with the other jobs, some other ops wiping out the part with the limit and replacing it with something of their own.

The first experiment to solve that however do not seem to get beyond the ig job so far.

**Comment by andreas-kupries** [comment]
First try at fitting the deployment into minikube despite mem requests failed.
Second try, without auto-scaler, works.
Used a modified memory config without defaults, using the minikube memory info with only light padding.

Now, results wrt application of limits to pods:

|Pod                    |Limit Ok       |Notes          |
|---                    |---            |---            |
|api                    |               |               |
|auctioneer             |:boom:         |Created in move|
|credhub                |:boom:         |?              |
|diego-api              |               |               |
|diego-cell             |               |               |
|doppler                |:boom:         |?              |
|log-api                |               |               |
|log-cache              |:boom:         |Created in move|
|nats                   |               |               |
|router                 |               |               |
|routing-api            |:boom:         |Created in move|
|singleton-blobstore    |               |               |
|tcp-router             |               |               |
|uaa                    |               |               |

The `created in move` are those I believe to understand. Using auctioneer as example the ops for the limit is
```
    - type: replace
      path: /instance_groups/name=scheduler/jobs/name=auctioneer/properties?/quarks/bpm/processes/name=auctioneer/limits/memory
      value: 22Mi
```
The ops performing the move of auctionerr out of scheduler into its on igroup then is
```
    - type: remove
      path: /instance_groups/name=scheduler/jobs/name=auctioneer
    - type: replace
      path: /instance_groups/name=scheduler:after
      value:
        name: auctioneer
        instances: 2
        stemcell: default
(...)
```

It wipes out the existing job in the scheduler, and creates a completely new igroup and job.
The spec for that job is the unmodified data from the cf-deployment.

The fix has to be the same as for auto-scaler.
Set the limits when all the igroups have moved into the correct place.

Note that the above does not explain `credhub`, nor `doppler`.

**Comment by andreas-kupries** [comment]
Fixed in the last two pushes. Handled the moved jobs as planned. "Solved" the other two (`credhub`, `doppler`) then in the same manner.

Note: With the operator adding requests identical to the limits deployment on a minikube can be very iffy. Had to temporarily disable setting limits on the `log-cache` (LC) to test that all the other limits are properly set. (LC is one of the pod with one of the larger containers around). And this was with using a `config/memory.yaml` heeding very close to the actual numbers pulled out of minikube from runs without limits. See the attached [memory.yaml.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5171800/memory.yaml.txt).

**Comment by andreas-kupries** [comment]
Note to self: Rebase and redo the `include` calling on and within config.load. Per Jan on slack: 
> Feel free to change that everywhere, once you have rebased on top of latest master. Remember to change mixins/*/templates/.yaml as well :slightly_smiling_face:

**Comment by andreas-kupries** [comment]
Rebased to latest head, bringing in the CF bump. Collapsed history of small fixes.
Remember the PR with tooling (helpers): cloudfoundry-incubator/kubecf-tools#8

**Comment by andreas-kupries** [comment]
The tweaks done in the last commit are based on a snapshot from the sandbox.

|Data|Description|
|---|---|
|[mem-sandbox-1-snap.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5176947/mem-sandbox-1-snap.txt)|Raw snap|
|[mem-sandbox-2-cleaned.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5176946/mem-sandbox-2-cleaned.txt)|Cleanup, removed POD|
|[mem-sandbox-3-max.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5176945/mem-sandbox-3-max.txt)|Reduced containers of replicated pods to max across|
|[mem-sandbox-4-padded.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5176944/mem-sandbox-4-padded.txt)|Data padded by 50%|
|[mem-sandbox-5-mi-unit.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5176943/mem-sandbox-5-mi-unit.txt)|Data rounded to full Mi|

In the config most of the Mi are further rounded up to next full ten, or 5.
In the case of the very large Mi switched to Gi and rounded further.

**Comment by viovanov** [comment]
@andreas-kupries please rebase 

This is currently blocked by [1], correct?

[1] https://github.com/cloudfoundry-incubator/quarks-operator/issues/1114

File: chart/templates/service_discovery_controller.yaml [hunk 8]
```diff
@@ -1,3 +1,4 @@
+{{- include "_config.load" $ }}
```
viovanov, line 1: I think I mentioned this before - can we lint for this?

chart/values.yaml
```diff
@@ -161,6 +161,25 @@ services:
...
+#
+# All values are integers; cpu values are in millicpus (m) and memory is in megabytes (Mi).
+#
+# More information about the `resources` structure can be found in the config/resources.yaml file
+# inside this helm chart.
+
+resources:
+  diego-cell:
+    garden:
+      garden: {memory: {limit: 2048, request: 2048}}
```
viovanov, line 181: This default seems small, I think it should be left to auto by default.
If we keep this non-prod item as a default, there need to be clear instructions on how to remove the limit and request.

chart/values.yaml
```diff
@@ -178,6 +197,13 @@ settings:
 
 
 features:
+  # Set default memory limits and requests for all containers
+  memory_limits:
+    enabled: true
+    # The memory request size default is calculated as a percentage of the limit.
+    # The default is always at least a minimum value, but never larger than the limit itself.
```
viovanov, line 204: 👍

scripts/kubecf-apply.sh
```diff
@@ -53,9 +56,19 @@ fi
 if [ -z "${CHART:-}" ]; then
     CHART="output/kubecf-$(./scripts/version.sh).tgz"
     export TARGET_FILE="${CHART}"
+    if [ -n "${RENDER_LOCAL:-}" ]; then
```
viovanov, line 59: What does `RENDER_LOCAL` mean? Can you render remotely?
Adding a new concept of `NO_IMAGELIST` - could just check for `RENDER_LOCAL`.
This adds 2 new concepts, I think they are both hard to understand intuitively. `RENDER_LOCAL` should be renamed and `NO_IMAGELIST` removed.

viovanov (approved):

chart/values.yaml
andreas-kupries: Possible workaround to get `auto`: Set garden/rep process lists in `config/job.yaml` to empty.
Cannot set to nil, this picks up defaults.

TODO: Testing the workaround.

andreas-kupries:
Have to require changes.
Current setup does *not* deploy with limit active.
Does deploy when disabling the limits.

Visible effect of the failure: ig pod completes init containers, and begins to start the kubecf pods. At some point later the ig pod and nearly all the kubecf pods. I.e. they decide to terminate, on their own, or ordered by ig/operator. Only the database pod is left. (And the seeder, but that is normal).

__Note__: Cannot do it officially as the original creator of the branch and PR.

chart/templates/_stacks.tpl
```diff
@@ -76,6 +76,12 @@
   {{- range $stack_name, $stack := $.kubecf.retval }}
     {{- $_ := set $stack "enabled" (has $stack_name $.Values.install_stacks) }}
 
+    {{- /* *** Mark all releases in the stack as data-only *** */}}
```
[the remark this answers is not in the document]
jandubois: It is only used for resources right now. It is also used to create the empty BPM process lists for the data-only releases, but that part is redundant; we need to add the data-only marker to non cf-deployment stack releases too before we can remove the BPM code from stacks.yaml. Anyways, this is all a trivial amount of code.

However, I find the name `data-only` more descriptive than `no-resources`, which would be just a side effect. I can see `no-processes`, but still prefer `data-only`, as that is really what these releases are: they have no active components.

chart/templates/service_discovery_controller.yaml
jandubois: Yes, there is a plan for this somewhere; might still be in the multi-stack issue. Either way, I'm aware and will file an issue for it if we don't have one yet.

scripts/kubecf-apply.sh
jandubois: I used `RENDER_LOCAL` because Andreas had created a copy of `kubecf-build.sh` called `kubecf-render-local, with the only different being that it called `helm template` instead of `helm upgrade`. I changed it to have an env setting to control this instead, to avoid the code duplication.

It is a feature for development/debugging of the templates.

Same thing for `NO_IMAGELIST`. During development I often insert a variable dump at a strategic place:

```
{{- include "_config.fail" (toYaml $.Values) }}
```

(or just some subtree that I'm interested in). This obviously breaks image_list.rb, which runs the templates internally, so I don't get a working chart that I can test with my own `--set` and `--values` overrides in `kubecf-render-local`. For this reason I object to removing the `NO_IMAGELIST` config variable.

Re-reading this I realize you want to just re-use one variable for both purposes. Will think about this on Wednesday when I'm back at work...

chart/values.yaml
jandubois: It feels weird to not specify a limit/request for this one pod when you explicitly ask for limits to be enabled.

I intentionally copied these 2 container settings into `values.yaml` so that they are kind-of in the face of the user.

I would rather set `features.memory_limits.enabled=false` by default and put a comment directly above that explains the need to configure the diego-cell size.

jandubois:
> Current setup does _not_ deploy with limit active.

I have successfully deployed this branch on minikube many times, including after the latest rebase/changes.

To be sure I deployed yet once more right now, and it is working too. The cluster is ready in 17 minutes (or 9 minutes after the `ig` job was finished).

The only thing special about this setup is that I run with a larger VM than normal:

```
VM_MEMORY=40960 make minikube-start
```

I haven't checked how much memory is actually needed by summing up all the default requests.

viovanov (dismissed):

chart/config/resources.yaml
```diff
@@ -0,0 +1,156 @@
...
+features:
+  cpu_limits:
+    enabled: true
+
+# resource definitions for addons. It has the same schema as the `resources` tree.
+
+addon_resources:
+  '$defaults': 32
+
+# `resources` contains a maping of instance groups to jobs to processes.
```
andreas-kupries, line 19: Comment nit, spelling, not a blocker.
```
- maping
+ mapping
```

chart/config/resources.yaml
```diff
@@ -0,0 +1,156 @@
...
+#       some_job:
+#         some_process:
+#           cpu:
+#             limit: ~
+#             request: ~
+#           memory:
+#             limit: ~
+#             request: ~
+#
+# Each level can define a "$defaults" resource definitions that will be applied
```
andreas-kupries, line 35: Spelling, nit, not a blocker.

```
- definitions
+ definition
```

viovanov (dismissed):

viovanov (dismissed):

jandubois (approved):

jandubois (approved):

jandubois (approved):

jandubois (dismissed):

jandubois (dismissed):

jandubois (approved):

viovanov (approved):

viovanov (dismissed):

jandubois (dismissed):

viovanov (approved):

jandubois (approved):

jandubois (approved):

PR #1264, opened 2020-08-28
fix: diego cells should use EmptyDir storage by default

### LINKED ISSUE #1241: Always use a hostpath storage class for diego cells
The diego cell volume is host to application data that changes often and seems to cause problems with storage classes based on NFS.
Other storage classes, such as the ones on EKS can also present problems.

Given that the recommended way to deploy diego cells on Kubernetes clusters is to tie them 1-to-1 to Kube Nodes, we should, by default, always use a hostpath storage class for the cell volumes. The option to override the storage class used for the cell should still be available.

Example storage class:

```
---
kind: StorageClass
apiVersion: storage.k8s.io/v1beta1
metadata:
  name: diego-cell-sc
provisioner: example.com/hostpath
parameters:
  path: /tmp
```

viovanov:
## Description
Diego cells no longer use a PVC. They will use EmptyDir storage from the host node.
The host node will need to have sufficient disk space to host the cell.
The amount of disk used by grootfs is calculated using the amount of ephemeral storage configured for the cells.

## Motivation and Context
Fixes #489 
Fixes #1241 

## How Has This Been Tested?
Ran locally.

## Types of changes

## Checklist:

### DIFF
--- a/bosh/releases/pre_render_scripts/diego-cell/garden/jobs/patch_grootfs_utils.sh
+++ b/bosh/releases/pre_render_scripts/diego-cell/garden/jobs/patch_grootfs_utils.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+target="/var/vcap/all-releases/jobs-src/garden-runc/garden/templates/bin/grootfs-utils.erb"
+sentinel="${target}.patch_sentinel"
+if [[ -f "${sentinel}" ]]; then
+  if sha256sum --check "${sentinel}" ; then
+    echo "Patch already applied. Skipping"
+    exit 0
+  fi
+  echo "Sentinel mismatch, re-patching"
+fi
+
+# Patch grootfs-utils to restrict the size of the sparse stores
+patch --verbose "${target}" <<'EOT'
+7,8c7,17
+<   /var/vcap/packages/thresholder/bin/thresholder "<%= p("grootfs.reserved_space_for_other_jobs_in_mb") %>" "$DATA_DIR" "$GARDEN_CONFIG_DIR/grootfs_config.yml" "<%= p("garden.graph_cleanup_threshold_in_mb") %>" "<%= p("grootfs.graph_cleanup_threshold_in_mb") %>"
+<   /var/vcap/packages/thresholder/bin/thresholder "<%= p("grootfs.reserved_space_for_other_jobs_in_mb") %>" "$DATA_DIR" "$GARDEN_CONFIG_DIR/privileged_grootfs_config.yml" "<%= p("garden.graph_cleanup_threshold_in_mb") %>" "<%= p("grootfs.graph_cleanup_threshold_in_mb") %>"
+---
+>   let grootfs_size="<%= p("grootfs.reserved_space_for_other_jobs_in_mb") %>"
+>   let disk_size=`df -BM /var/vcap/data/grootfs/store/ | tail -n 1 | awk '{gsub("M", "", $2); print $2}'`
+>   let reserved_disk="$disk_size - $grootfs_size"
+> 
+>   if [ "$grootfs_size" -gt "$disk_size" ]; then
+>     echo "The node running this cell doesn't have enough disk space. You requested ${grootfs_size}M but the disk is ${disk_size}M in size."
+>     exit 1
+>   fi
+> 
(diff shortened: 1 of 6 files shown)


JimmyMa:
@viovanov so the PVC will not be supported for diego-cell? if that's true, I think the diego-cell will not work well in some cloud providers, for example: the worker node has very limit local disk, like 100G, and the local disk size is not customizable in some cloud providers, and this blocks diego-cell to host more apps.

viovanov:
> @viovanov so the PVC will not be supported for diego-cell? if that's true, I think the diego-cell will not work well in some cloud providers, for example: the worker node has very limit local disk, like 100G, and the local disk size is not customizable in some cloud providers, and this blocks diego-cell to host more apps.

I think that's a fair concern, but I think this is a limitation of Diego. We've noticed instability when trying to use storage classes like NFS.
@JimmyMa you still have the ability to override this setting if you know that the storage class can handle diego.
Please see below:

```yaml
# not tested, but it's what the old setting used to do
operations:
  inline:
  - type: replace
    path: /instance_groups/name=diego-cell/env?/bosh?/agent?/settings?/ephemeralAsPVC?
    value: true
  - type: replace
    path: /instance_groups/name=diego-cell/vm_resources?/ephemeral_disk_size?
    value: 102400
```

cfcontainerizationbot (approved):

viovanov:
> I would have to actually deploy and test this before I feel comfortable approving it.
> 
> Maybe I need to look more closely, but your patch seems to only change the thresholder configuration. Maybe I misremember this, but I thought that is just a garbage collection limit, and the initial size of the sparse file system is set by `grootfs init-store`, and I don't see providing an argument for it.

I've looked at the sparse files before opening the PR, and their size is correct.

jandubois:
> I've looked at the sparse files before opening the PR, and their size is correct.

I've done this too now, and it is indeed correct. I guess my understanding of `grootfs init-store` is incomplete then, but I'm going to take this as-is and will not look deeper into why it actually works. 😄 

> @JimmyMa you still have the ability to override this setting if you know that the storage class can handle diego.

I wonder if we shouldn't provide a more direct way to configure this than inline ops files.

I also wonder what the correct default value is; I suspect that the storage class issues might be limited to NFS. In that case I would think the default config should stay as PVC, with an override to switch to `emptyDir` for users who deploy with NFS.

jandubois (dismissed):
I'm approving as-is, because you requested a review. Ideally I would like to see this behaviour configurable via `values.yaml` and leave the current behaviour as the default for backwards compatibility.

I'm not going to merge; you can decide if you want to make further changes, or move the configurability into a separate PR. I would just be worried that we make a 2.5 release with a changed default, after which is becomes harder to revert to the previous one.

viovanov:
> I'm approving as-is, because you requested a review. Ideally I would like to see this behaviour configurable via `values.yaml` and leave the current behaviour as the default for backwards compatibility.
> 
> I'm not going to merge; you can decide if you want to make further changes, or move the configurability into a separate PR. I would just be worried that we make a 2.5 release with a changed default, after which is becomes harder to revert to the previous one.

I also thought about this, and I don't think making this configurable makes sense (or even documenting it as an alternative). Actual deployments of KubeCF should deploy cells on nodes with taints, and local disk size must be made available. I think this needs to be made more clear.

There are some StorageClasses that could make sense for cell storage (even though you'd likely have to pay for them too much: Azure/AWS/GKE elastic storage, vSphere disks and maybe OpenStack Cinder)?

jandubois:
One thing I'm worried about, but haven't been able to check on minikube:

Is the volume size for the `emptyDir` volume guaranteed? Or is it possible that other containers on the same node could use up the disk space eventually, leaving the sparse filesystems without sufficient backing store expand to full capacity?

This could be an ugly error condition; I don't know how XFS reacts, but btrfs would turn the while filesystem read-only, breaking all apps running on that cell, but not actually terminating the cell itself.

viovanov:
> One thing I'm worried about, but haven't been able to check on minikube:
> 
> Is the volume size for the `emptyDir` volume guaranteed? Or is it possible that other containers on the same node could use up the disk space eventually, leaving the sparse filesystems without sufficient backing store expand to full capacity?
> 
> This could be an ugly error condition; I don't know how XFS reacts, but btrfs would turn the while filesystem read-only, breaking all apps running on that cell, but not actually terminating the cell itself.

It's definitely not guaranteed afaik. The node can use up space and starve the cell.
This is why taints are important.

jandubois (dismissed):
Approving so that CI can run.

I haven't actually reviewed the latest changes, so this is not an approval for merging! 😃

jandubois (dismissed):

jandubois (approved):

jandubois (approved):

PR #1266, opened 2020-08-28
Return empty string when looking up non-existing properties

jandubois:
The documentation says:

> Lookup $path under $context and return either the value found, or the empty string.

However, if along the path there is an empty map or slice, lookup would return this map or slice, even though should have been more path elements to traverse.

This commit fixes the logic to set `$.kubecf.retval` to `nil` and return an empty string instead. The only time when _config.lookup should return `{}` or `[]` is when the value **at the end of the path** exists and is an empty map or slice.

### DIFF
--- a/chart/templates/_config.tpl
+++ b/chart/templates/_config.tpl
@@ -169,6 +169,9 @@
       {{- end }}
     {{- end }}
   {{- else }}
+    {{- if gt (len $path) 1 }}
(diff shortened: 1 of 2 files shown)


jandubois:
@viovanov This fix should solve the issue we had earlier today in #1264

viovanov (approved):

Issue #776, opened 2020-04-28
Make c2c networking work


viovanov:
**Is your feature request related to a problem? Please describe.**
c2c networking is not enabled in KubeCF.

**Describe the solution you'd like**

- [x] implement a CoreDNS plugin that replaces the bosh dns adapter
- [ ] enable the CATS group for c2c networking
- [x] support in Quarks

**Describe alternatives you've considered**
n/a

**Additional context**
This will likely require more containers with extra privileges and support in Quarks for more BOSH DNS features.

fargozhu:
@viovanov do we have any use cases to support this spike?

viovanov:
@fargozhu this is a feature of CF that has many usecases, and is also requested by our IBM friends
ping @qu1queee

smoser-ibm:
@viovanov this would be the last missing  piece for us regarding a CF Bosh ->  K8s migration. Can we bump the importance of this maybe ?

qu1queee:
This is on my radar, trying to find someone from our side to help, I dont have too much bandwidth for this atm.

viovanov:
@JimmyMa can you please add a comment here with your setup and your findings?

mook-as:
This is expected to run (and fail) on minikube.

viovanov:
We need to implement an "App DNS" as described below:

![app dns diagram](https://docs.google.com/drawings/d/e/2PACX-1vTWqgVvy2_tGJ4I3jisuw6IvWxM8Jt3B6_eRV_FabhC2tY_CyC0mUbOMJ7FhXYglBygJVx2yMEfEiGz/pub?w=1465&h=653)

viovanov:
Ping @JimmyMa ⬆️
@f0rmiga is starting work on that app dns server

f0rmiga:
A few notes based on our call:
- A patch to the `silk-cni` job is most likely needed to be able to get the k8s svc cluster IP from the injected env var to be used by the Garden Container `/etc/resolv.conf`. We have done similar patches in the past, so it should be simple.
- Configuring the Apps DNS to use the Quarks DNS as the upstream server might be tricky unless we make the Apps DNS a BOSH-release and let the quarks-operator do the magic to point its `/etc/resolv.conf` to the Quarks DNS. That's not going to be my first attempt but I'll certainly fall back to this approach to avoid doing any clever tricks that might compromise future maintainability.
[...]

PR #1275, opened 2020-09-01
feat: Implement eirini-dns-aliases

### LINKED ISSUE #284: Make Eirini apps work with our DNS
**Describe the bug**

Since Eirini apps cannot resolve the DNS records from the kubecf coreDNS instance, credhub.servie.cf.internal becomes unreachable.

**To Reproduce**

Deploy kubecf with Eirini and check that credhub CATS suite doesn't pass.

**Expected behavior**

Eirini apps should be able to resolve the kubecf DNS entries. It may be possible with an EiriniX mutator.

**Environment**

All versions.

mook-as:
## Description
This adds the [eirini-dns-aliases](https://github.com/SUSE/eirini-dns-aliases) eirini extension; it causes the application containers (as in, things deployed with `cf push`) to use the BOSH DNS resolver instead of the default Kubernetes one.

## Motivation and Context
Fixes #284

## How Has This Been Tested?
- Deployed locally, with eirini
- From an application container, `curl -kv https://credhub:8844`
- Resolving `credhub` (to `credhub.service.cf.internal`, then to an IP address) was only possible because of the extension.

## Types of changes

## Checklist:

## Additional Information:

- We may want to wait for #1259 to land first, and then rebase this on top of that work.
- This needs to wait for https://github.com/SUSE/eirini-dns-aliases/pull/1 to be merged first.
- This will affect https://github.com/cloudfoundry-incubator/kubecf/issues/776 — but that should just mean `DNS_SERVICE_HOST` from `bosh-dns.kubecf.svc` to whatever the "App DNS" service mentioned in https://github.com/cloudfoundry-incubator/kubecf/issues/776#issuecomment-684916460 is instead.
- I intentionally omitted resource constraints, as #1142 is still in flux.  If that lands first, we'll need to rebase this to include it.

I'm marking this blocked for now to wait on #1259. /cc @jandubois as a heads up for that.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -438,3 +438,7 @@ eirinix:
     icon_image: Eirini broker
     secrets:
       auth-password: ~ # Password is randomly generated if not given
+  dns-aliases:
+    image: ghcr.io/cfcontainerizationbot/eirini-dns-aliases
+    image_tag: 0.0.0-18.gdf4c9f7b
+    pullPolicy: IfNotPresent
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -438,7 +438,3 @@ eirinix:
     icon_image: Eirini broker
     secrets:
       auth-password: ~ # Password is randomly generated if not given
-  dns-aliases:
-    image: ghcr.io/cfcontainerizationbot/eirini-dns-aliases
-    image_tag: 0.0.0-18.gdf4c9f7b
-    pullPolicy: IfNotPresent
--- a/mixins/eirinix/templates/dns-aliases/deployment.yaml
+++ b/mixins/eirinix/templates/dns-aliases/deployment.yaml
@@ -6,15 +6,15 @@ kind: Deployment
 metadata:
   name: eirini-dns-aliases
   labels:
-    {{- list . "eirini-dns-aliases" | include "kubecf.component.labels" | nindent 4 }}
+    {{- list . "eirini-dns-aliases" | include "component.labels" | nindent 4 }}
 spec:
   selector:
     matchLabels:
-      {{- list . "eirini-dns-aliases" | include "kubecf.component.selector" | nindent 6 }}
+      {{- list . "eirini-dns-aliases" | include "component.selector" | nindent 6 }}
   template:
     metadata:
       labels:
-        {{- list . "eirini-dns-aliases" | include "kubecf.component.labels" | nindent 8 }}
+        {{- list . "eirini-dns-aliases" | include "component.labels" | nindent 8 }}
     spec:
       serviceAccountName: eirinix
       containers:
--- a/mixins/eirinix/templates/dns-aliases/deployment.yaml
+++ b/mixins/eirinix/templates/dns-aliases/deployment.yaml
@@ -32,11 +32,9 @@ spec:
             value: "8443"
           - name: SERVICE_NAME
             value: eirini-dns-aliases
-          - name: OPERATOR_FINGERPRINT
-            value: eirini-dns-aliases
           - name: DNS_SERVICE_HOST
             value: {{ printf "apps-dns.%s.svc" .Release.Namespace }}
           ports:
-          - name: webhook-tls
+          - name: webhook
             containerPort: 8443
 {{- end }}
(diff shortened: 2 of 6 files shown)


mook-as:
This should be ready for review now:

- #1259 has landed
- SUSE/eirini-dns-aliases#1 has landed.
- #776 has landed (and this has been adjusted to match)
- #1142 is still pending, ignoring for now.

chart/values.yaml
```diff
@@ -436,3 +436,7 @@ eirinix:
     icon_image: Eirini broker
     secrets:
       auth-password: ~ # Password is randomly generated if not given
+  dns-aliases:
```
viovanov, line 439: Can you include these details in here?  https://github.com/cloudfoundry-incubator/kubecf/blob/master/mixins/eirinix/config/eirinix.yaml

Like all the other eirinix components.

mixins/eirinix/templates/dns-aliases/deployment.yaml
```diff
@@ -0,0 +1,42 @@
+{{- $_ := include "_config.load" $ }}
+{{- if .Values.features.eirini.enabled }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: eirini-dns-aliases
+  labels:
+    {{- list . "eirini-dns-aliases" | include "kubecf.component.labels" | nindent 4 }}
```
viovanov, line 9: this is less readable, i'd rather we see the labels, they don't change often

viovanov (changes requested):

viovanov (approved):

viovanov (approved):

mook-as:
I can't reproduce the CI issue locally (that is: eirini smoke tests succeed locally).

Hopefully https://github.com/cloudfoundry-incubator/eirini-loggregator-bridge/pull/26 will help.

viovanov (dismissed):

mixins/eirinix/templates/dns-aliases/service.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    {{- list . "eirini-dns-aliases" | include "component.labels" | nindent 4 }}
+spec:
+  type: ClusterIP
+  selector:
+    {{- list . "eirini-dns-aliases" | include "component.selector" | nindent 4 }}
+  ports:
+  - protocol: TCP
+    name: webhook
+    port: 443
```
mudler, line 18: If you are using latest EiriniX, you can just

```suggestion
    port: 8443
```

mixins/eirinix/templates/dns-aliases/service.yaml
```diff
@@ -0,0 +1,20 @@
...
+    targetPort: webhook-tls
```
mudler, line 19:
```suggestion

```

mixins/eirinix/templates/dns-aliases/deployment.yaml
```diff
@@ -0,0 +1,42 @@
...
+          env:
+          - name: EIRINI_NAMESPACE
+            value: {{ .Values.eirini.opi.namespace }}
+          - name: WEBHOOK_NAMESPACE
+            value: {{ .Release.Namespace }}
+          - name: PORT
+            value: "8443"
+          - name: SERVICE_NAME
+            value: eirini-dns-aliases
+          - name: OPERATOR_FINGERPRINT
```
mudler, line 35: This is something that should be hardcoded in the binary and not something we should edit later on - it causes mutatingwebhooks and secrets to change names, which probably would break during upgrades

mixins/eirinix/templates/dns-aliases/service.yaml
mudler: for reference: https://github.com/cloudfoundry-incubator/kubecf/blob/master/mixins/eirinix/templates/loggregator-bridge.yaml#L106

mudler:
The EiriniX version https://github.com/SUSE/eirini-dns-aliases/blob/master/go.mod#L9 used by eirini-dns-aliases is quite behind, I'd bump that also for coherence with the other extensions

mudler (changes requested):
eirini-dns-aliases comes with any test (unit-tests at least on the extensions would be prefered)

viovanov (approved):

viovanov (dismissed):

viovanov (approved):

viovanov (dismissed):

viovanov (approved):

mudler (approved):

PR #1283, opened 2020-09-03
fix: cells can be scaled without restarting the entire set

### LINKED ISSUE #981: After helm upgrade to scale diego cell count from 3 to 4, 2 out of 3 diego-cell pods gets restarted
**Describe the bug**
When I use `helm upgrade` to scale up the diego cell count for 3 to 4. diego-cell-2 and diego-cell-1 get restarted consecutively, once new diego-cell-3 is up. 
diego-cell-0 never gets restarted

It takes 19 min approx for all pods to get in `running` state

**To Reproduce**

- HA deployment (3 diego cells)
- Deploy an app with 3 instances 
- run helm upgrade to bring the count of deigo cell to 4

**Expected behavior**
A new diego cell should appear, no pods should restart.

**What happens**
some of the old diego-cell pods (2 out of 3) get restarted

**Environment**

 kubecf-2.0-RC1                  2.0-RC1 
cf-operator-4.5.6+0.gffc6f942   4.5.6+0.gffc6f942

**Additional context**
```
diego-cell-0                             7/7     Running     2          3h6m
diego-cell-1                             7/7     Running     2          6m57s
diego-cell-2                             7/7     Running     2          10m
diego-cell-3                             7/7     Running     2          24m
```

viovanov:
## Description
Configures the diego cell IG so it doesn't use unnecessary link instances or the REPLICAS env var.

## Motivation and Context
#fixes #981 

## How Has This Been Tested?
Locally scaled from 2 to 4 cells and back.

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ password
 *.vmdk
 personal-setup
 .concourse/config.yaml
-
+.envrc
+/*values.yaml
 /output
 /tmp
--- a/scripts/cf-operator-delete.sh
+++ b/scripts/cf-operator-delete.sh
@@ -3,15 +3,16 @@ source scripts/include/setup.sh
 
 require_tools kubectl helm
 
-set +e
+if helm_ls 2>/dev/null | grep -qi cf-operator ; then
+    helm_delete cf-operator --namespace cf-operator
+fi
 
-helm delete cf-operator
-kubectl delete ns "${CF_OPERATOR_NS}"
-kubectl delete clusterrole cf-operator-quarks-job-cluster
-kubectl delete clusterrole cf-operator-cluster
-kubectl delete clusterrolebinding cf-operator-quarks-job-cluster
-kubectl delete clusterrolebinding cf-operator-cluster
-kubectl delete crd boshdeployments.quarks.cloudfoundry.org
-kubectl delete crd quarksjobs.quarks.cloudfoundry.org
-kubectl delete crd quarkssecrets.quarks.cloudfoundry.org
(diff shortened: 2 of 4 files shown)


scripts/cf-operator-delete.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+source scripts/include/setup.sh
+
+require_tools kubectl helm
+
+set +e
```
viccuad, line 6: Nit: I would prefer to `set -e`, and use

```
if helm_ls 2>/dev/null | grep -qi cf-operator ; then
    helm_delete cf-operator --namespace cf-operator
fi
kubectl delete --ignore-not-found ns "${CF_OPERATOR_NS}"
```

and so on

viccuad (approved):

.gitignore
```diff
@@ -8,6 +8,7 @@ password
 *.vmdk
 personal-setup
 .concourse/config.yaml
-
+.envrc
```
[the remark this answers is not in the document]
viovanov: @viccuad my envrc looks like this:

```
export VALUES=./dev-values.yaml
```

I do all my settings from `dev-values.yaml`. 
```

scripts/cf-operator-delete.sh
viccuad: With the new changes, we can drop the `set +e`. I would like to do `set -e` in scripts/include/setup.sh to catch errors in the scripts in the future, but that's out of this PR.

viccuad (approved):

cfcontainerizationbot (approved):

PR #1290, opened 2020-09-03
feat: scripts/image_list: support VALUES override

mook-as:
## Description
This adds support for the `VALUES` override when building the image list.  This is helpful when testing local changes to be able to customize the list of images that will be needed (so they can be traferred to the kubernetes cluster).  This should not be used in CI builds.

Note that most of the changes are to make rubocop happy.

## Motivation and Context
While working on #284 I needed to set an override for the image for a new thing that we are not publishing yet (to use my WIP image).  That means I need to provide overrides, but that wasn't possible with the way we were doing the image list.

## How Has This Been Tested?
Ran locally, and loaded all the resulting images into minikube.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/image_list.rb
+++ b/scripts/image_list.rb
@@ -18,6 +18,16 @@ class Hash
     end
     super
   end
+
+  def deep_merge!(other)
+    merge! other do |_key, old, new|
+      if old.respond_to? :deep_merge!
+        old.deep_merge! new
+      else
+        new
+      end
+    end
+  end
 end
 
 # deep_populate_nil_values is useful for setting the nil values found under the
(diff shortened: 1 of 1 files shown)


scripts/image_list.rb
```diff
@@ -53,23 +67,20 @@ def deep_populate_nil_values(hash)
 }
 
 # Process the non-BOSH releases.
-values['releases'].keys.each do |release_name|
+values.releases.keys.each do |release_name|
   # Filter out the 'defaults' key as it's not a release.
   next if release_name == 'defaults'
```
jandubois, line 72: This should be `$defaults` now, for correctness. In practice it doesn't matter because the default entry will not have an `image` key, so will be skipped in the next condition.

jandubois (approved):
LGTM; the `defaults` → `default` rename is technically outside the scope of this PR.

Issue #1289, opened 2020-09-03
Logging agents in the diego cells are not identified properly in a multi-az deployment


viovanov:
We currently use the pod ordinal as a replacement for spec index in the env vars for the logging jobs.
We should use the actual spec index once it's available.

Needs https://www.pivotaltracker.com/story/show/174661471

PR #1291, opened 2020-09-03
fix: logging agents configuration

### LINKED ISSUE #407: Stratos Console only showing 1 of 2 diego cells (because of missing metrics)
**Describe the bug**

Stratos Console only showing 1 of 2 diego cells.

Paraphrased from @nwmac 

If I look in Prometheus, I only get one series for the metric 

`firehose_value_metric_rep_capacity_remaining_containers` - e.g.

```
firehose_value_metric_rep_capacity_remaining_containers{bosh_deployment="kubecf",bosh_job_id="diego-cell-0",bosh_job_name="diego-cell-loggr-forwarder-agent",cloud_foundry="https://api.10.86.2.34.xip.io",environment="wss://doppler.10.86.2.34.xip.io:443",instance="metrics-f-exp-service-0.metrics.svc.cluster.local:9186",job="cf-firehose",origin="rep",unit="Metric"}
```

There should be a series for each Diego Cell... so I'll assume this is a KubeCF bug/issue

Possibly unrelated, but important: The metadata for KubeCF is empty - if you go to /v2/info - name, build, support, description, version are all empty 

**To Reproduce**
Install kubecf-v0.0.0-e7534b6 using cf-operator-2.3.0+0.g27a91cdf with at least 2 diego cells, then deploy console-helm-chart-3.0.0-rc.1-cf818986f (or probably any version of Stratos). Second diego cell will not appear in any Stratos views. and 

**Expected behavior**
We should see all diego cells in Stratos

### LINKED ISSUE #1198: Diego cell info does not appear in Stratos
**Describe the bug**
"Cell information for this Cloud Foundry cannot be found." message in stratos instead of Diego cell info (admin view).

**To Reproduce**
Install KubeCF 2.2.2 with Diego enabled and Stratos 3.2.1 or 4.0 RC.

**Expected behavior**
Stratos should show which Diego cell an app instance is running on (app view) and show which apps are running on a given Diego cell (CF admin view).

**Environment**
 - KubeCF 2.2.2
 - on CaaSP 4.2 (on Open Stack) or on EKS
 - nginx-ingress

**Additional context**
Not sure if I've ever seen this working properly with KubeCF. Worked with SCF.

viovanov:
## Description
Fixes logging agents configuration for spec index and IPs, which are not available in BPM yaml rendering.

## Motivation and Context

fixes #407 
fixes #1198 

## Types of changes

## Checklist:

### DIFF
--- a/chart/assets/operations/addons.yaml
+++ b/chart/assets/operations/addons.yaml
@@ -2,6 +2,8 @@
   path: /addons/name=bpm
 
 # The loggr forwarder agent needs a special env var with tags that identify the job
+# The AGENT_INDEX env var cannot be rendered properly as part of BPM rendering, it can only be set
+# correctly through pod reflection. Without this, the index would be 0 for any replica
 - type: replace
   path: /addons/name=forwarder_agent/jobs/name=loggr-forwarder-agent/properties?/quarks/envs
   value:
@@ -25,6 +27,8 @@
     value: "deployment:$(AGENT_DEPLOYMENT),job:$(AGENT_IG_NAME)-loggr-forwarder-agent,index:$(AGENT_INDEX),ip:$(AGENT_IP)"
 
 # The loggregator agent needs env vars to identify the job
+# The AGENT_INDEX env var cannot be rendered properly as part of BPM rendering, it can only be set
+# correctly through pod reflection. Without this, the index would be 0 for any replica
 - type: replace
   path: /addons/name=loggregator_agent/jobs/name=loggregator_agent/properties?/quarks/envs
   value:
--- a/chart/assets/operations/addons.yaml
+++ b/chart/assets/operations/addons.yaml
@@ -1,13 +1,40 @@
 - type: remove
   path: /addons/name=bpm
 
+# The loggr forwarder agent needs a special env var with tags that identify the job
 - type: replace
-  path: /addons/name=loggregator_agent/jobs/name=loggregator_agent/properties/quarks?/envs?
+  path: /addons/name=forwarder_agent/jobs/name=loggr-forwarder-agent/properties?/quarks/envs?
   value:
   - name: AGENT_IP
     valueFrom:
       fieldRef:
-        apiVersion: v1
         fieldPath: status.podIP
   - name: AGENT_INDEX
-    value: "0"
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.labels['statefulset.kubernetes.io/pod-name']
+  - name: AGENT_DEPLOYMENT
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.labels['quarks.cloudfoundry.org/deployment-name']
+  - name: AGENT_IG_NAME
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.labels['quarks.cloudfoundry.org/instance-group-name']
+  - name: AGENT_TAGS
+    value: "deployment:$(AGENT_DEPLOYMENT),job:$(AGENT_IG_NAME)-loggr-forwarder-agent,index:$(AGENT_INDEX),ip:$(AGENT_IP)"
+
+# The loggregator agent needs env vars to identify the job
+- type: replace
+  path: /addons/name=loggregator_agent/jobs/name=loggregator_agent/properties?/quarks/envs?
+  value:
+  - name: AGENT_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: status.podIP
+  - name: AGENT_INDEX
+    valueFrom:
+      fieldRef:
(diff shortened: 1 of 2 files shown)


chart/assets/operations/addons.yaml
```diff
@@ -1,13 +1,40 @@
...
   - name: AGENT_IP
     valueFrom:
       fieldRef:
-        apiVersion: v1
         fieldPath: status.podIP
   - name: AGENT_INDEX
-    value: "0"
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.labels['statefulset.kubernetes.io/pod-name']
```
mook-as, line 15: Why isn't this just
```suggestion
        fieldPath: metadata.name
```

mook-as (changes requested):
This needs
- A *much* better commit message, with at least a paragraph of detail (after the subject line).
- Alternatively, decent levels of documentation (as comments) inside the ops file.

Not of what things it's trying to do, but of the details of how the environment variables get overridden.

chart/assets/operations/addons.yaml
viovanov: Consistent with the other envs

chart/assets/operations/addons.yaml
```diff
@@ -1,13 +1,40 @@
...
+  - name: AGENT_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: status.podIP
+  - name: AGENT_INDEX
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.labels['statefulset.kubernetes.io/pod-name']
+  - name: AGENT_TAGS
+    value: "deployment:$(AGENT_DEPLOYMENT),job:$(AGENT_JOB),index:$(AGENT_INDEX),ip:$(AGENT_IP)"
```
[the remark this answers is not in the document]
viovanov: `AGENT_INDEX ` is correct because you're likely only looking at pod ordinal 0

chart/assets/operations/addons.yaml
mook-as, line 15:
```suggestion
        fieldPath: metadata.name
```
Sorry, I accidentally deleted this comment when I meant to delete something else.

chart/assets/operations/addons.yaml
```diff
@@ -1,13 +1,40 @@
...
+  path: /addons/name=loggregator_agent/jobs/name=loggregator_agent/properties?/quarks/envs?
+  value:
+  - name: AGENT_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: status.podIP
+  - name: AGENT_INDEX
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.labels['statefulset.kubernetes.io/pod-name']
```
mook-as, line 38:
```suggestion
        fieldPath: metadata.name
```
Would be good to explain in the comment / commit message that we always need to override `AGENT_INDEX` because `spec.id` in the `ig` container always ends up being `<ig>-0`, so it will be incorrect on instances with `index > 0`.  This is based on us rendering `bpm.yaml` _once_ for all replicas.

chart/assets/operations/instance_groups/diego-cell.yaml
```diff
@@ -54,6 +54,19 @@
...
+  path: /instance_groups/name=diego-cell/jobs/name=loggr-udp-forwarder/properties?/quarks/envs?
+  value:
+  - name: IP
+    valueFrom:
+      fieldRef:
+        fieldPath: status.podIP
+  - name: INDEX
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.labels['statefulset.kubernetes.io/pod-name']
```
mook-as, line 68:
```suggestion
        fieldPath: metadata.name
```

mook-as (changes requested):
Poking at it more, the problem (that was totally unexplained) was that things go wrong with `spec.index > 0`, because we render BPM once for all replicas.  Please say that either in comments, or at least in the (amended) commit message, so when we go look at this change in a year we can figure out why.

chart/assets/operations/addons.yaml
mook-as: But that isn't consistent with `AGENT_IP` at all? Why are we looking at a StatefulSet-specific annotation, when we just need the pod-specific unique identifier?

viovanov: Because of this: https://github.com/cloudfoundry-incubator/kubecf/issues/1289.
This will become a quarks label when it's available.

mook-as: Okay, that's a good reason to use an annotation, but not _this_ annotation. So given that we don't have a proper one to use, we should be… not using this _particular_ annotation?

mook-as (approved):

Issue #1307, opened 2020-09-09
incorrect NODE_INDEX and AZ_INDEX in log-cache


ShuangMen:
**Environment**
cf-operator: 5.2.0
kubecf: v2.2.3 

deploy kubecf with multi_az="false" and log-cache instance =3, 

**Issue**
After deployment, The issue is that in all 3 log-cache pod, container log-cache environment show the "AZ_INDEX: 1" and "NODE_INDEX: 0".

below is an example when describes one of the log-cache pod `log-cache-2`.

```
Name:         log-cache-2
Namespace:    kubecf
Priority:     0
Node:         10.240.1.13/10.240.1.13
Start Time:   Tue, 08 Sep 2020 22:21:22 -0400
Labels:       controller-revision-hash=log-cache-5696ddc959
              quarks.cloudfoundry.org/az-index=0
              quarks.cloudfoundry.org/deployment-name=kubecf
              quarks.cloudfoundry.org/instance-group-name=log-cache
              quarks.cloudfoundry.org/pod-ordinal=2
              quarks.cloudfoundry.org/quarks-statefulset-name=log-cache
              statefulset.kubernetes.io/pod-name=log-cache-2
...
    Environment:
      ADDR:                    :8080
      AZ_INDEX:                1
      CA_PATH:                 /var/vcap/jobs/log-cache/config/certs/ca.crt
      CERT_PATH:               /var/vcap/jobs/log-cache/config/certs/log_cache.crt
      KEY_PATH:                /var/vcap/jobs/log-cache/config/certs/log_cache.key
      MAX_PER_SOURCE:          100000
      MEMORY_LIMIT_PERCENT:    50
      METRICS_CA_FILE_PATH:    /var/vcap/jobs/log-cache/config/certs/metrics_ca.crt
      METRICS_CERT_FILE_PATH:  /var/vcap/jobs/log-cache/config/certs/metrics.crt
      METRICS_KEY_FILE_PATH:   /var/vcap/jobs/log-cache/config/certs/metrics.key
      METRICS_PORT:            6060
      NODE_ADDRS:              log-cache-0:8080,log-cache-1:8080,log-cache-2:8080
      NODE_INDEX:              0
...
```
The issue is that in all 3 log-cache pod, container log-cache environment show the "AZ_INDEX: 1" and "NODE_INDEX: 0".
[...]

PR #1310, opened 2020-09-10
fix: the log-cache instance number can not be adjusted

JimmyMa:
## Description
This change makes the log-cache instance number adjustable in user's values.yml file.

## Motivation and Context
the log-cache instance number can not be adjusted in user's values.yml file, and it's hard to have multiple log-cache instances.

## How Has This Been Tested?
1. tested with values.yml without log-cahce instances defined
2. tested with values.yml with 3 log-cache instances

## Types of changes

## Checklist:


jandubois:
Note that log-cache was intentionally removed from the `doppler` instance group so that it can be made a singleton: https://github.com/cloudfoundry-incubator/kubecf/pull/619

There is some theory that log-cache has a memory leak when running with more than one instance due to the way the caches communicate with each other: https://github.com/cloudfoundry/log-cache-release/issues/29

Later experiments seem to indicate that log-cache doesn't have a leak, but uses the total memory size of the node it is running on to determine the cache size, making it look like it has unbounded memory usage: https://github.com/cloudfoundry/gosigar/issues/48

This issue still seems unresolved, so I don't think it is pre-mature to allow the instance count to be increased right now.

Once the memory configuration question has been resolved, we should first decide if log-cache should be returned to the `doppler` instance group before deciding if log-cache needs to be independently scalable.

JimmyMa:
@jandubois thank you for your comment, I don't know the history, and your comment helped me a lot.

I think only one log-cache POD is not good for a production environment, and seems like no quick solution for https://github.com/cloudfoundry/gosigar/issues/48 . A workaround for the memory issue could be to set a lower memory percent for log-cache: https://github.com/cloudfoundry/log-cache-release/blob/develop/jobs/log-cache/spec#L35-L37

Another blocking issue for multiple log-cache PODs is https://github.com/cloudfoundry-incubator/kubecf/issues/1307 , the log-cache will not be functional when there are multiple log-cache PODs, because their NODE_INDEX is same (0),  please see: https://github.com/cloudfoundry/log-cache-release/blob/fa637cd7e2899d58930f893e52b9911cb8672755/src/internal/cache/log_cache.go#L181-L221

Adding below statements to log-cache.yml could make each log-cache POD has different NODE_INDEX, but this is only working for non-multi-az,  when multi-az is enabled, the log-cache in different az will have deplicated NODE_INDEX.

```
- type: replace
  path: /instance_groups/name=log-cache/jobs/name=log-cache/properties/quarks?
  value:
    envs:
    - name: NODE_INDEX
      valueFrom:
        fieldRef:
          fieldPath: metadata.labels['quarks.cloudfoundry.org/pod-ordinal']
```

jandubois:
@JimmyMa We talked about this PR in our planning meeting today and created #1312 to create a PR to fix the issue upstream.

If we do decide to merge this PR, it should be changed so that even when `high_availability` is set, the count remains at `1`, and the `values.yaml` entry should have a warning about increasing the instance count.

Here is a workaround we used before to limit the `log-cache` memory growth:

```
bosh:
  instance_groups:
  - name: log-cache
    jobs:
    - name: log-cache
      properties:
        memory_limit_percent: 3
```

This of course requires that all your node have the same total memory size; otherwise you need to use taints and tolerations to limit your log-cache pods to nodes with a uniform size.

I think I would prefer to not make any changes until the upstream issue has been fixed; it all feels like a hack otherwise.

Once upstream can determine the memory available inside the pod, we can set a memory limit, and everything should start working. Is there a reason we shouldn't then move `log-cache` back into the `doppler` role (like cf-deployment) instead of having it in its own instance group?

jandubois (changes requested):
Just blocking the PR from getting accidentally merged. See comment above about what we would like to see if we are to merge before the upstream fix is in place.

jandubois:
Since I was just looking at a BPM rendering issue, I wanted to leave this here, for whenever we get back to it:

The reason `NODE_INDEX` is wrong is because it is set in `bpm.yml.erb` from `spec.id`: https://github.com/cloudfoundry/log-cache-release/blob/3229f06/jobs/log-cache/templates/bpm.yml.erb#L9-L11

Quarks-operator is rendering `bpm.yml.erb` just once in the `ig` job and uses the rendered file to generate the pod spec for the stateful sets. That means `spec.address`, `spec.index`, `spec.id`, etc. in BPM template rendering always contain the value for instance 0.

Issue #1109, opened 2020-07-07
Proposal: Create *much* smaller buildpack container images that never need to be updated


jandubois:
# Creating buildpack images

## Building the buildpack and their assets

I was temporarily under the impression that the buildpacks themselves (the artifacts referenced from the ZIP files on the individual buildpack repos) were built on top of the stemcells because the buildpack images are updated whenever the stemcell is updated.

This would be wrong, but hard to detect, as our rootfs and stemcells are very close.

I have since learned that the buildpacks are built on OBS "on top of a similar-to-rootfs OS" using [Meta Configuration of Cloud:Platform:buildpacks](https://build.opensuse.org/projects/Cloud:Platform:buildpacks/meta). I'm not really sure what this means, but it sounds correct-ish.

There should never be a reason to rebuild a buildpack on top of a newer rootfs because all buildpack releases have to be backwards compatible to the earliest still supported rootfs. All fixes for CVEs etc. should be in the shared libraries, that are part of the rootfs. Fixes for vulnerabilities in bundled assets are handled by updates to the buildpacks themselves.

## Building buildpack images for kubecf

Buildpack releases are converted to container images like any other bosh release, which means they are built on top of the current stemcell.

This is quite wasteful, as the buildpacks themselves will run on top of the stack rootfs and don't need the stemcell at all.

There are only 2 scripts that are executed on these images (inside init containers), the generic `spec-copier` generated by `quarks-operator`, and a `pre-start` script we generate from the helm chart via ops-file.

The `spec-copier` runs (under `time sh -c`) both in the `ig` job and the `api` pod:

```
mkdir -p /var/vcap/all-releases/jobs-src/binary-buildpack
cp -ar /var/vcap/jobs-src/* /var/vcap/all-releases/jobs-src/binary-buildpack
chown vcap:vcap /var/vcap/all-releases/jobs-src/binary-buildpack -R
```

This is completely unneccessary because buildpacks have neither properties nor templates; the cf-operator does not need access to the spec files. For a deployment with both `cflinuxfs3` and `sle15` stacks these `spec-copier` scripts add an extra 20 init containers that don't do anything.

The generated `pre-start` scripts runs in the `api` pod and looks like this:

```
mkdir -p "/var/vcap/data/shared-packages/binary-buildpack/"
cp -r /var/vcap/packages "/var/vcap/data/shared-packages/binary-buildpack/"
```

[...]

PR #1315, opened 2020-09-11
fix: Adapt pipelines to new path

thardeck:
## Description
The SUSE buildpacks and stemcell information was moved from `values.yaml` to `chart/config/sle15.yaml`. The stemcell and buildpack update pipelines were adapted accordingly.

The replacement is a simple regex to reduce complexity of the pipelines and save time since we might implement #1109 in the nearby future so it is not worth covering every corner case.

Since the stemcell pipeline triggers every time there is a new stemcell I do not think it is necessary to handle different stemcell versions for each buildpack.

## How Has This Been Tested?
The pipelines were updated manually and ran against current master. Since all buildpacks and the stemcell are already up to date there were no changes to commit yet.

## Types of changes

## Checklist:


thardeck:
https://github.com/cloudfoundry-incubator/kubecf/pull/1315

viovanov:
lgtm

PR #1320, opened 2020-09-11
fix: apps-dns labels for configmap

f0rmiga:


mook-as (dismissed):
LGTM.

I'll have some helpers in #1275 and then do some follow-up PRs later.

viovanov (approved):

Issue #1191, opened 2020-08-04
Configurable database seeder timeout


mook-as:
## Acceptance Criteria

- the timeout for the seeder should be configurable

---

**Describe the bug**
The database seeder has a script that waits for the database to be up, with a hard-coded 3 second timeout:
https://github.com/cloudfoundry-incubator/kubecf/blob/a06e4ebbc2f187a321f991381a14e69df31a158c/deploy/helm/kubecf/assets/scripts/jobs/pxc/seeder.sh#L13

This doesn't work well when my cluster is slow:
```
database-seeder-4d0a31667279bc1e-wtck5:~ # time echo 'SELECT "Ready";' | mysql
Ready
Ready

real    0m4.025s
user    0m0.016s
sys     0m0.004s
```

**To Reproduce**
Found this while trying to do air-gapped installs, so maybe try that on minikube? I'm not sure.  I'm currently on 4547685b57ee49f53ae724f7235f4147275c846a.

**Expected behavior**
Database seeder job succeeds.

**Environment**
 - KubeCF Version 4547685b57ee49f53ae724f7235f4147275c846a
 - Minikube (that works fine normally)
 - No internet access (I turned off the NAT network for minikube).

PR #1324, opened 2020-09-14
feat: Configurable database seeder connection timeout

andreas-kupries:
## Description

  - Added timeout to `values.yaml`. Already handled by the schema, generically.
  - Added templating to database, propagating the data in the `values.yaml` into the pod environment.
  - Extended the seeder script to use the value from the environment.

## Motivation and Context

#1191 

:warning: On looking at the modified code I am not sure why a configurable timeout is needed. The command where the timeout applies is wrapped into a loop which repeats forever until the database is ready. Regardless of timeout used AFAICT.

As such I am not sure of the original diagnosis in #1191 :warning:

## How Has This Been Tested?

Used a minikube to deploy kubecf, with the new key set to

```
features:
  embedded_database:
    connect_timeout: 10
```

and inspected the generated kube object to verify that

  - It contains the value in the proper/new environment variable.
  - The inlined script properly used the new environment variable.

## Types of changes

## Checklist:

### DIFF
--- a/chart/assets/scripts/jobs/pxc/seeder.sh
+++ b/chart/assets/scripts/jobs/pxc/seeder.sh
@@ -10,7 +10,7 @@ password=${DATABASE_ROOT_PASSWORD}
 EOF
 
 echo "Waiting for database to be ready..."
-until echo "SELECT 'Ready!'" | mysql --connect-timeout=3 1> /dev/null 2> /dev/null; do
+until echo "SELECT 'Ready!'" | mysql --connect-timeout="${DATABASE_CONNECT_TIMEOUT}" 1> /dev/null 2> /dev/null; do
   sleep 1
 done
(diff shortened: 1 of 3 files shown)


jandubois:
> ⚠️ On looking at the modified code I am not sure why a configurable timeout is needed. The command where the timeout applies is wrapped into a loop which repeats forever until the database is ready. Regardless of timeout used AFAICT.

This issue is about the database being so slow that it takes more than 3 or 10, or even 12 seconds to *connect*.

So even though it retries, if the connection attempt is always aborted after 3 seconds, then it will never become ready.

mook-as (approved):

andreas-kupries:
@jandubois :bulb: thank you.

viovanov (approved):

Issue #1257, opened 2020-08-26
HA upgrades with multiple app instances have noticeable app downtime


gaktive:
As spotted during kubecf 2.2.2 to 2.2.3 upgrade testing, we're getting reports that app downtime is obvious. This is despite having an HA setup with multiple app instances and load distributed across Diego cells.

@HartS and @svollath can provide details on their environments but it's across different Kubernetes implementations.

mook-as:
As far as I can tell, the HA configuration in catapult in fact doesn't HA the database (because it doesn't respect the HA flag correctly).

Still investigating.

mook-as:
The actual reason seems to be because the `router-public` service has:
```yaml
selector:
  app.kubernetes.io/component: router
  app.kubernetes.io/instance: susecf-scf
  app.kubernetes.io/version: v2.2.3
```
The service has the version as part of the selector, so it will never match the old router (only the new one once it rolls over).  That's defined [here](https://github.com/cloudfoundry-incubator/kubecf/blob/c4af91cf8a779405b41a33ad44dacc3d800003ea/chart/templates/ingress.yaml#L121).

Judging by the history, we've never had it work before? That code's been in there since v0.1.0.

jandubois:
> Judging by the history, we've never had it work before? That code's been in there since v0.1.0.

Yeah, but what about SCF? 😉

mook-as:
Fixed in #1325 — the upgrades are still likely to _fail_ (because of other issues), but the apps stay up at least? :D

PR #1325, opened 2020-09-14
fix: chart: public svcs: don't select on version

mook-as:
## Description
In the public services (used when we're not using ingress), don't have a version label in the selector when finding backends.  This is necessary to ensure that, during upgrades, we can select for whichever pod happens to be running at the time (including the old version).  Otherwise we'll unnecessarily introduce downtime as we refuse to connect to the perfectly fine pods (from the previous version).

The same logic also applies to downgrades (where the working pods are from a newer version).

## Motivation and Context
For https://github.com/cloudfoundry-incubator/kubecf/issues/1257 - but this might not fix the whole thing (as the database is still not properly HA).

## How Has This Been Tested?
Deployed locally on a HA cluster.

## Types of changes

## Checklist:

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -118,7 +118,6 @@ spec:
   selector:
     app.kubernetes.io/component: router
     app.kubernetes.io/instance: {{ $root.Release.Name | quote }}
-    app.kubernetes.io/version: {{ default $root.Chart.Version $root.Chart.AppVersion | quote }}
   ports:
   - name: http
     protocol: TCP
@@ -162,7 +161,6 @@ spec:
   selector:
     app.kubernetes.io/component: ssh-proxy
     app.kubernetes.io/instance: {{ $root.Release.Name | quote }}
-    app.kubernetes.io/version: {{ default $root.Chart.Version $root.Chart.AppVersion | quote }}
   ports:
   - name: ssh
     protocol: TCP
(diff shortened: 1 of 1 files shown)


viovanov:
@mook-as do you have some docs links - the change seems to constrain the selection of the backend even more

jandubois:
> the change seems to constrain the selection of the backend even more

@viovanov I don't understand this. If you remove a condition from the selector, how does it constrain the selection? This PR removes the constraint on the version, so *any* version will match now.

This seems to be what we want/need. The only concern I have that this means that there should never be a change in the services that is not backwards compatible. Not sure if that is realistic, but we can always add a version constraint back when we need one.

mook-as:
Rebased; attempted to rebase onto the merge-base of 2.2.3 for so we can merge easier for #1257 but led to merge conflicts that GitHub couldn't resolve by itself, so I gave up.

> This seems to be what we want/need. The only concern I have that this means that there should never be a change in the services that is not backwards compatible. Not sure if that is realistic, but we can always add a version constraint back when we need one.

Hmm, so that'd be something like changing the port number for http/https to a non-privileged port internally, so even if the old pods match they don't have the right listening port? Yeah, in that case we'll need to have additional selectors again (but being careful to do it in a way that _further_ upgrades will not be an issue).  But in that case we'll have downtime anyway (since at the start of the upgrade nothing will match), so that's… maybe a moot point?

jandubois:
> in that case we'll have downtime anyway

yes

viovanov:
> > the change seems to constrain the selection of the backend even more
> 
> @viovanov I don't understand this. If you remove a condition from the selector, how does it constrain the selection? This PR removes the constraint on the version, so _any_ version will match now.
> 
> This seems to be what we want/need. The only concern I have that this means that there should never be a change in the services that is not backwards compatible. Not sure if that is realistic, but we can always add a version constraint back when we need one.

:) my dark reader made the line change look greenish so I thought this was adding it 

+1

viovanov (approved):

PR #1339, opened 2020-09-17
fix: move all the rfds into an specific rfd folder.

fargozhu:
## Description

Move all the RFDs folders into an specific `rfd` so the links from the main page can work and for better organization.

## Motivation and Context

Fix and Organize.

## How Has This Been Tested?
N/A

N/A

## Types of changes
N/A

## Checklist:
N/A


mook-as:
Could you describe what kind of links from the main page we're talking about? I don't see anything mentioning `RFD` at the [top level readme](https://github.com/cloudfoundry-incubator/kubecf/blob/master/README.md) at the moment, and I'm confused to as why adding an additional level makes a difference.

viovanov:
@fargozhu the new paths have double `rfd`, why do you need that?

viovanov:
ping @fargozhu

fargozhu:
yes so we can "tool" around like: updating the RFD landing page when a new RFD is pushed and some other automation processes.

mook-as:
Wouldn't that already be possible, since all RFDs were in a subdirectory already? It's unclear to me why we needed the extra directory (and in fact, we currently appear to have extra directories _inside_ extra directories).

Issue #1007, opened 2020-06-11
Use proper options when running CATs on Eirini


jimmykarily:
**Is your feature request related to a problem? Please describe.**
CATs fail on Eirini. We need to run only CATs that make sense and should be green when running on Eirini deployments.

**Describe the solution you'd like**
Look here on how this can be implemented:
(https://github.com/SUSE/catapult/blob/master/modules/tests/kubecf-test.sh#L139).

jimmykarily:
Upstream eirini team was using these: https://github.com/cloudfoundry-incubator/eirini-ci/blob/09dcce6d9e900f693dfc1a6da70b5a526cf7de18/pipelines/dhall-modules/jobs/run-core-cats.dhall#L52-L86

jimmykarily:
Started adding support for the Eirini cats suite here: https://github.com/SUSE/catapult/commits/cats_eirini

jimmykarily:
capi_experimental has been removed: https://github.com/cloudfoundry/cf-acceptance-tests/pull/383/commits/00fe329ab742a9ce31039d627e3aa62c95e607a1

Running `v3` should be enough.

jimmykarily:
`credhub` has been replaced by: `credhub_mode`, `credhub_location`, `credhub_client`, `credhub_secret`


https://github.com/cloudfoundry/cf-acceptance-tests/commit/0823b205764bdedb720f5a94a627b6e6c1f0beba
https://github.com/cloudfoundry/cf-acceptance-tests/commit/e28bfaed9e5fd2154580ae53b223ecd865eed131

jimmykarily:
I'm debugging and fixing various CAT failures here: https://github.com/jimmykarily/cf-acceptance-tests/compare/master..fix_eirini_cats

For now I'm finding workarounds trying to understand what the problem is. Before we try to upstream we need proper fixes for all problems (e.g. the loggregator bridge missing logs).

jbuns:
Test results summary so far:

`Summarizing 11 Failures:

[Fail] [apps] Environment Variables Groups Staging environment variable groups [It] Applies environment variables while staging apps
/../cf-acceptance-tests/apps/environment_variables_group.go:135

[Fail] [apps] Healthcheck when the healthcheck is set to none [It] starts up successfully
/../cf-acceptance-tests/apps/healthcheck.go:50

[...]

PR #1343, opened 2020-09-17
fix: bump cats release

jimmykarily:
to use the SUSE fork with our changes:

https://github.com/SUSE/cf-acceptance-tests/commit/536591ba792785e0090d3a21240a0e818dcf1486

and bump cf cli:
https://github.com/SUSE/cf-acceptance-tests-release/commit/3c01b157d7d912d5b5d5c88457ca5d54d42a7d80

This is part of this story:
https://github.com/cloudfoundry-incubator/kubecf/issues/1007

## How Has This Been Tested?
Run CATs locally on kind using the same setting and cf cli version on kubecf master and Eirini

## Types of changes

## Checklist:

### DIFF
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -29,7 +29,7 @@ releases:
     condition: false
   cf-acceptance-tests:
     condition: testing.cf_acceptance_tests.enabled
-    version: 0.0.16
+    version: 0.0.18
   credhub:
     condition: features.credhub.enabled
     version: ~


chart/config/releases.yaml
```diff
@@ -29,7 +29,7 @@ releases:
     condition: false
   cf-acceptance-tests:
     condition: testing.cf_acceptance_tests.enabled
-    version: 0.0.16
```
jimmykarily, line 32: Is this all it takes to consume this release? https://ci.flintstone.cf.cloud.ibm.com/teams/quarks/pipelines/release-images/jobs/build-cf-acceptance-tests/builds/53

viccuad: from https://github.com/cloudfoundry-incubator/kubecf/blob/master/doc/cf_deployment/bump.md#bump-cats-release

it seems to be it. The outdated docs point to when the versions were pinned in values.yaml, which I think informs us that there is no other place to change besides what you already changed.

viovanov (approved):

chart/config/releases.yaml
jandubois: This looks correct to me (as long as there is an image of that version that has been built with the required stemcell).

Issue #1333, opened 2020-09-15
`high_availability` doesn't HA database pods


mook-as:
**Describe the bug**
When we set `high_availability: true` in the deployment values, I'd expect the database to also be HA.  However, I only get one database pod instead.

**To Reproduce**
1. `helm install … --set high_availability=true`
2. `make kubecf-wait`
3. `kubectl get pods --all-namespaces | grep database`

**Expected behavior**
More than one database pod.

**Environment**
- KubeCF `master` c4af91cf8a779405b41a33ad44dacc3d800003ea

**Additional context**
https://github.com/cloudfoundry-incubator/kubecf/blob/98e278dc7bb949fe7cab75888e5dda8af51d3d86/chart/templates/database.yaml#L199-L204

PR #1345, opened 2020-09-17
fix: Modified database template to handle `features.high_availability`.

andreas-kupries:
## Description

Reworked the templating logic to check `features.high_availability` after `sizing.database.instances` before falling back to the minimum default.

## Motivation and Context

See #1333 

## How Has This Been Tested?

Deployed to local minikube for all possibilities, i.e. SA, HA, explicit sizing (with HA, to verify that sizing has precedence).

## Types of changes

## Checklist:

### DIFF
--- a/chart/templates/database.yaml
+++ b/chart/templates/database.yaml
@@ -196,11 +196,15 @@ spec:
         app.kubernetes.io/version: {{ default .Chart.Version .Chart.AppVersion | quote }}
         helm.sh/chart: {{ include "kubecf.chart" . }}
     spec:
-      {{- with $replicas := default 1 .Values.sizing.database.instances }}
-      {{- if eq (mod $replicas 2) 0 }}
-      {{- include "_config.fail" "The number of database instances must be odd to avoid split-brain" }}
-      {{- end }}
-      replicas: {{ $replicas }}
+      {{- if $.Values.sizing.database.instances }}
+        {{- if eq (mod $.Values.sizing.database.instances 2) 0 }}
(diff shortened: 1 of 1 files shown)


chart/templates/database.yaml
```diff
@@ -196,11 +196,15 @@ spec:
...
+      {{- if $.Values.sizing.database.instances }}
+        {{- if eq (mod $.Values.sizing.database.instances 2) 0 }}
+        {{- include "_config.fail" "The number of database instances must be odd to avoid split-brain" }}
+        {{- end }}
+      replicas: {{ $.Values.sizing.database.instances }}
+      {{- else if $.Values.high_availability }}
+      replicas: 3
+      {{- else }}
+      replicas: 1
       {{- end }}
```
mook-as, line 208: Hmm, intuitively it feels like assigning to a variable ahead of time would be clearer, but I'm not sure that's _actually_ the case:
```suggestion
      {{- $replicas := 1 }}
      {{- if $.Values.sizing.database.instances }}
        {{- $replicas = $.Values.sizing.database.instances }}
      {{- else if $.Values.high_availability }}
        {{- $replicas = 3 }}
      {{- end }}
      {{- if eq (mod $replicas 2) 0 }}
        {{- include "_config.fail" "The number of database instances must be odd to avoid split-brain" }}
      {{- end }}
      replicas: {{ $replicas }}
```

(Unfortunately, we can't do `{{ with x }} … {{ else with y }}`.)

mook-as (commented):
I _think_ we'll also need to check that HA upgrades work correctly in this case, but I'm not sure it should be part of this PR.

mook-as (approved):

Issue #703, opened 2020-04-16
Document how to rotate generated secrets


jandubois:
There is a documentation page for it already: https://github.com/cloudfoundry-incubator/kubecf/blob/master/doc/secret_rotation.md

However, it only documents how to rotate the database encryption key, which is just one subtopic of rotating all the secrets in the cluster.

The rotation of generated secrets will have to be managed by cf-operator, but we need documentation about how to trigger it.

viovanov:
There should be a script (like a one line) to perform rotation for all secrets.

viovanov:
Also see the following resources:

https://quarks.suse.dev/docs/quarks-secret/development/#_secretrotation-controller_
https://quarks.suse.dev/docs/quarks-secret/#rotate-credentials

mook-as:
@jandubois Is this already done by #854? That documented how to rotate a QuarksSecret.  It doesn't contain a script to rotate all secrets, though.

viovanov:
@mook-as I don't think that's enough, and a sample script should be part of this

mudler:
Can we please avoid to document this inside the kubecf repo, but to https://kubecf.suse.dev/ instead? It sounds more user-facing docs

mook-as:
Yep, this will be an update to https://github.com/cloudfoundry-incubator/kubecf-docs/blob/master/content/en/docs/Tasks/secrets.md

But to make it _reasonable_ for the user, I'll attempt to make a QuarksJob to automate the rotate-all-secrets case, and tell the user to trigger that instead.

PR #1346, opened 2020-09-17
feat: QuarksJob to rotate all secrets

mook-as:
## Description
This adds a job to trigger secret rotation for all generated secrets.

This is useful for administrators to rotate all secrets (generally as a precaution); attempting to document manual instructions is likely to be error-prone.

## Motivation and Context
This is in support of #703 — trying to get the administrator/operator to type in a script correctly is going to be painful; much better to automate it instead.

## How Has This Been Tested?
Triggered the QuarksJob locally to see that secrets are rotated.

Generally speaking though, my experiences has tended towards this actually breaking the cluster (because the database password gets out of sync).  That will be filed as a bug afterwards, because the user can _already_ manually rotate the secrets; this just automates the process.

## Types of changes

## Checklist:

  (new service account + role + role binding; all minimally scoped)
  (That will be a follow-up)

### DIFF
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -48,6 +48,11 @@ releases:
   postgres:
     condition: features.autoscaler.enabled
     version: "39"
+  secret-rotation:
+    # secret-rotation is not a BOSH release.  It requires ruby & kubectl.
+    image:
+      repository: splatform/fissile-stemcell-sle
+      tag: SLE_15_SP1-27.4
   sync-integration-tests:
     # XXX SITS only makes sense when using Diego; add error check somewhere?
     condition: testing.sync_integration_tests.enabled
--- a/chart/templates/secret-rotation.yaml
+++ b/chart/templates/secret-rotation.yaml
@@ -1,0 +1,79 @@
+# This creates a (manually triggered) Quarks Job to rotate all generated
+# secrets.
+---
+apiVersion: quarks.cloudfoundry.org/v1alpha1
+kind: QuarksJob
+metadata:
+  name: secret-rotation
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- list $ "secret-rotation" | include "component.labels" | nindent 4 }}
+spec:
+  trigger:
+    strategy: manual
+  template:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: generate-rotation-config-map
+            {{- with $image := index $.Values.releases "secret-rotation" "image" }}
+            image: {{ printf "%s:%s" $image.repository $image.tag }}
+            imagePullPolicy: {{ $image.pullPolicy | quote }}
+            {{- end }}
+            command:
+            - /usr/bin/env
+            - ruby
+            - -e
+            - |
+              {{- .Files.Get "assets/scripts/helm/secret-rotation/rotate-secrets.rb" | nindent 14 }}
+            env:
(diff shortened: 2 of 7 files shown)


chart/config/releases.yaml
```diff
@@ -48,6 +48,11 @@ releases:
   postgres:
     condition: features.autoscaler.enabled
     version: "39"
+  secret-rotation:
+    # secret-rotation is not a BOSH release.  It requires ruby & kubectl.
+    image:
+      repository: splatform/fissile-stemcell-sle
+      tag: SLE_15_SP1-27.4
```
mook-as, line 55: Not really happy with this; suggestions welcome.

chart/templates/secret-rotation.yaml
```diff
@@ -0,0 +1,79 @@
+# This creates a (manually triggered) Quarks Job to rotate all generated
+# secrets.
+---
+apiVersion: quarks.cloudfoundry.org/v1alpha1
+kind: QuarksJob
+metadata:
+  name: secret-rotation
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- list $ "secret-rotation" | include "component.labels" | nindent 4 }}
```
mook-as, line 10: https://github.com/cloudfoundry-incubator/kubecf/blob/d192fa2f97033f8c9cd83dbb7dac1d2f8caed012/chart/templates/_helpers.tpl#L68

chart/config/releases.yaml
viovanov: @mook-as create an issue for creating such a base (we'll likely have to do it on OBS)

viovanov (changes requested):
If running this breaks the cluster because of the db password and/or encryption keys, please exclude them from rotation.

mook-as:
Hmm, doing more testing, I seem to have issues with rotating secrets:

While I'm rotating everything, I'll focus on the `locket` database password:
- The secret `var-locket-database-password` is correctly rotated (as in, it gets a new randomly generated value).
- `ig-resolved.diego-api-v1` is **not** updated; it has the old value.
- I've confirmed that the `diego-api` QuarksStatefulSet does have `updateOnConfigChange: true`.

This leads to database connection errors (because the `database-seeder` is re-run, changing all the passwords).  Not sure yet why the `ig` job isn't being re-started.

mook-as:
This is now blocked by https://github.com/cloudfoundry-incubator/quarks-secret/issues/55 — it turns out that rotating the CA certificate does _not_ rotate the certificate that was made with that CA.  This means that if we rotate everything, there's a chance the CA certificate is rotated _after_ the certificates it signs, causing those to no longer verify with the (new) CA certificate.

We also have an issue with https://github.com/cloudfoundry-incubator/quarks-operator/issues/1136 where the various QuarksStatefulSets do not get updated when the secrets they depend on change; this means that we end up not being able to restart things on connection loss.  Fortunately that could be worked around by adding annotations to the various things (to cause them to get updated correctly), even though there's a feature that should make the annotations unnecessary.

viovanov:
@mook-as this originated as a documentation story;
can this still be done manually? we can work on automation later.

mook-as:
@viovanov I do not believe attempting to get users to do this manually is reasonable:
- Finding all the secrets will be error-prone
- Even if they _did_, it will still break because quarks-operator doesn't update dependent secrets correctly; this means rotating the wrong secret will still break the cluster _some of the time_.

PR #1353, opened 2020-09-21
fix: remove the validation label

fargozhu:
## Description

The latest changes to the GitHub templates for some reason made the templates useless 'cause they are not being triggered since then.

## Motivation and Context

To have the Bug and Feature templates back to the game.

## How Has This Been Tested?

Nope.

## Types of changes
N/A

## Checklist:
N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: 'bug'
+labels: 'Type: Bug'
 assignees: 'fargozhu'
 
 ---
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,8 +2,8 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: 'bug', 'Status: Validation'
-assignees: ''
+labels: 'bug'
```
mook-as, line 5: I believe the label is currently `Type: Bug` instead.  I wonder if _that_ is the reason things broke?

fargozhu: it's not possible to debug on GH just by try-error but it was a good catch @mook-as ...

PR #1364, opened 2020-09-22
fix: build script when KUBECONFIG is exported

f0rmiga:
When having `KUBECONFIG` environment variable set on my shell session, the `make kubecf-build` command would fail with:
```
Traceback (most recent call last):
	4: from scripts/image_list.rb:48:in `<main>'
	3: from /usr/lib64/ruby/2.5.0/psych.rb:314:in `safe_load'
	2: from /usr/lib64/ruby/2.5.0/psych.rb:350:in `parse'
	1: from /usr/lib64/ruby/2.5.0/psych.rb:402:in `parse_stream'
/usr/lib64/ruby/2.5.0/psych.rb:402:in `parse': (<unknown>): mapping values are not allowed in this context at line 1 column 85 (Psych::SyntaxError)
make: *** [Makefile:38: kubecf-build] Error 1
```

### DIFF
--- a/scripts/kubecf-build.sh
+++ b/scripts/kubecf-build.sh
@@ -3,6 +3,8 @@ source scripts/include/setup.sh
 
 require_tools bosh cf_operator_url git helm jq j2y ruby y2j
 
+unset KUBECONFIG
+
 if [[ ! "$(git submodule status -- src/cf-deployment)" =~ ^[[:space:]] ]]; then
     die "git submodule for cf-deployment is uninitialized or not up-to-date"
 fi


jandubois:
I would like to understand what exactly in your kubeconfig is interfering with the script. It must be something particular to your setup, as it is not a problem in general, e.g.

```
$ export KUBECONFIG=~/.kube/config
$ make kubecf-build
Successfully packaged chart and saved it to: output/kubecf-0.2.0-1947.gade8d6c3.tgz
```

So I wonder if all scripts running `helm` internally are broken in your configuration, in which case the fix in this PR would be insufficient.

f0rmiga:
> I would like to understand what exactly in your kubeconfig is interfering with the script. It must be something particular to your setup, as it is not a problem in general, e.g.
> 
> ```
> $ export KUBECONFIG=~/.kube/config
> $ make kubecf-build
> Successfully packaged chart and saved it to: output/kubecf-0.2.0-1947.gade8d6c3.tgz
> ```
> 
> So I wonder if all scripts running `helm` internally are broken in your configuration, in which case the fix in this PR would be insufficient.

Thanks for verifying it in your setup. That per se is evidence that the build system is not deterministic since we can get different results in different environments. Also, to make things worse, today I do **not** have the same problem anymore. No, I did not change my environment.

jandubois:
I have an idea what is happening, but I need to find some time to verify it.

f0rmiga:
I converted the PR to a draft so we can keep the conversation but not merge it.

f0rmiga:
Closing in favour of https://github.com/cloudfoundry-incubator/kubecf/issues/1515.

Issue #1367, opened 2020-09-23
Missing CF info API values


viccuad:
We are not filling the API info values.

Example:
Hitting http://api.vic-cap.ci.kubecf.charmedquarks.me/info  I only get
```
{
  "name": "",
  "build": "",
  "support": "",
  "version": 0,
  "description": "",
  "authorization_endpoint": "https://login.vic-cap.ci.kubecf.charmedquarks.me",
  "token_endpoint": "https://uaa.vic-cap.ci.kubecf.charmedquarks.me",
  "allow_debug": true
}

```

It should look like: (from https://apidocs.cloudfoundry.org/268/info/get_info.html)
```
{
  "name": "vcap",
  "build": "2222",
  "support": "http://support.cloudfoundry.com",
  "version": 2,
  "description": "Cloud Foundry sponsored by Pivotal",
  "authorization_endpoint": "http://localhost:8080/uaa",
  "token_endpoint": "http://localhost:8080/uaa",
  "min_cli_version": null,
  "min_recommended_cli_version": null,
  "api_version": "2.89.0",
  "app_ssh_endpoint": "ssh.system.domain.example.com:2222",
  "app_ssh_host_key_fingerprint": "47:0d:d1:c8:c3:3d:0a:36:d1:49:2f:f2:90:27:31:d0",
  "app_ssh_oauth_client": null,
  "routing_endpoint": "http://localhost:3000",
  "logging_endpoint": "ws://loggregator.vcap.me:80"
}
```
This is consumed and shown in Stratos.

This would also be changed and consumed downstream, too, so if possible it needs to be able to change without recreating a bosh release.

ShuangMen:
I try to curl different api versions and get different outputs.

```
$ curl https://api.mytest.kubecf.cloud/info |jq
{
  "name": "",
  "build": "",
  "support": "",
  "version": 0,
  "description": "",
  "authorization_endpoint": "https://login.mytest.kubecf.cloud",
  "token_endpoint": "https://uaa.mytest.kubecf.cloud",
  "allow_debug": true
}

$ curl https://api.mytest.kubecf.cloud/v2/info |jq
{
  "name": "",
  "build": "",
  "support": "",
  "version": 0,
  "description": "",
  "authorization_endpoint": "https://login.mytest.kubecf.cloud",
  "token_endpoint": "https://uaa.mytest.kubecf.cloud",
  "min_cli_version": null,
  "min_recommended_cli_version": null,
[...]

PR #1395, opened 2020-09-28
fix: Add missing data for the CF info endpoints.

andreas-kupries:
## Description

Added ops instructions setting the CAPI cc_ng properties controlling the values to be returned by the endpoints `info`, `v2/info`, and `v3/info`.

## Motivation and Context

See #1367 

## How Has This Been Tested?

Deployed on minikube, logged inot CF, then used `cf curl ENDPOINT` for the endpoints above and checked that the values are not empty anymore, and the expected.

## Types of changes

## Checklist:

### DIFF
--- a/chart/templates/_config.tpl
+++ b/chart/templates/_config.tpl
@@ -17,6 +17,10 @@
 | uppercase config filename, to be merged first. Obviously no config file can overwrite
 | an explict user choice from $.Values.
 |
+| __Note__ that while templates are allowed in config files, they only have access to
+| the global helm variables, and the user-provided values from `values.yaml`. They cannot
+| use the internal config data, as that has not been merged yet.
+|
 | The $.Values.release.$defaults are saved in $.kubecf.defaults because they are used
 | in templates to set the default and addon stemcells.
 |
@@ -46,7 +50,7 @@
 
     {{- $configs := dict }}
     {{- range $name, $bytes := $.Files.Glob "config/*" }}
-      {{- $config := $bytes | toString | fromYaml }}
+      {{- $config := tpl ($bytes | toString) $ | fromYaml }}
       {{- if hasKey $config "Error" }}
         {{- include "_config.fail" (printf "Config file %q is invalid:\n%s" $name $config.Error) }}
       {{- end }}
--- a/scripts/helmlint.sh
+++ b/scripts/helmlint.sh
@@ -17,9 +17,12 @@ helm lint "${TARGET_DIR}" --set system_domain=example.com
 # Running json schema validator a second time to also check the embedded config files.
 # Done in a separate run because the config files are merged here on top of values.yaml,
 # which will cause replacement of any arrays.
+#
+# NOTE: We exclude all config files which contain templating.
 
 # shellcheck disable=SC2046
 # We want word splitting with find.
 helm template "${TARGET_DIR}" --set system_domain=example.com \
-     --values "$(perl -e 'print join ",", @ARGV' -- $(find "${TARGET_DIR}/config" -type f))" \
+     --values "$(perl -e 'print join ",", @ARGV' -- \
+              $(grep -L '{{' $(find "${TARGET_DIR}/config" -type f)))" \
      > /dev/null
--- a/chart/config/api.yaml
+++ b/chart/config/api.yaml
@@ -1,0 +1,11 @@
+properties:
+  api:
+    cloud_controller_ng:
+      name: KubeCF
+      version: 2
+      description: 'CloudFoundry KubeCF'
+      support_address: 'https://kubecf.io/'
+      build: {{ default .Chart.Version .Chart.AppVersion | quote }}
+
+# Note that the version information has to be a plain integer, as per the
+# Membrane::SchemaValidation occuring in `bosh-pre-start-cloud-controller-ng`.


andreas-kupries:
@troytop I asked for your review because I do not know if the support address I entered (https://scc.suse.com) is (still) correct.

chart/assets/operations/instance_groups/api.yaml
```diff
@@ -32,6 +32,27 @@
...
   value: 3600
 
+# Default settings for the CF info endpoints.
+##
+# Note that the version information has to be a plain integer, as per the
+# Membrane::SchemaValidation occuring in `bosh-pre-start-cloud-controller-ng`.
+
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/name?
+  value: 'SUSE CAP Kubecf'
```
mook-as, line 42: This would be incorrect for KubeCF (but would be correct for CAP).

chart/assets/operations/instance_groups/api.yaml
```diff
@@ -32,6 +32,27 @@
...
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/build?
+  value: {{ default .Chart.Version .Chart.AppVersion | quote }}
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/version?
+  value: 2
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/support_address?
+  value: 'https://scc.suse.com'
```
mook-as, line 51: This would also be incorrect for KubeCF (SUSE shouldn't be getting support requests for things we don't sell).

mook-as (dismissed):

chart/assets/operations/instance_groups/api.yaml
andreas-kupries: Tweaked in commit 6997dff1ed9dd3f7eef0472ad4ea5ffe04afb737

chart/assets/operations/instance_groups/api.yaml
andreas-kupries: Tweaked in commit 6997dff1ed9dd3f7eef0472ad4ea5ffe04afb737 as FIXME.

chart/assets/operations/instance_groups/api.yaml
```diff
@@ -32,6 +32,27 @@
...
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/version?
+  value: 2
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/support_address?
+  value: 'https:// ??? FIXME ???'
+- type: replace
+  path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/description?
+  value: 'CloudFoundry KubeCF'
+
```
jandubois, line 55: Instead of a series of ops, I would write this in a config file, e.g. `config/api.yaml`:

```
properties:
  api:
    cloud_controller_ng:
      name: KubeCF
      version: 2
```

Of course this doesn't work for `build` because we currently don't support template expressions in the config files (but that might be fixable). Just leave a comment in the config file that explains that the `build` value is set via ops.

andreas-kupries: +1. Should make it easier to modify when KubeCF is deployed as part of CAP.

jandubois: > we currently don't support template expressions in the config files (but that might be fixable).

I would try this (in https://github.com/cloudfoundry-incubator/kubecf/blob/master/chart/templates/_config.tpl#L49)

change

```
{{- $_ := set $configs $name ($bytes | toString | fromYaml) }}
```

to

```
{{- $_ := set $configs $name ($bytes | toString | tpl | fromYaml) }}
```

I think it should work will become generally useful.

andreas-kupries: Ok. Will try that in a second iteration of my current changes. Which implement the basic move to config/api.yml.

jandubois: > Should make it easier to modify when KubeCF is deployed as part of CAP.

It doesn't matter; CAP could still use `properties` to overwrite whatever you are generating with your ops. Internally `properties` still generates the same kind of ops, but they are applied last, so they can overwrite *everything*: https://github.com/cloudfoundry-incubator/kubecf/blob/master/chart/templates/bosh_deployment.yaml#L90-L91

I still prefer to move config stuff as much as possible out of ops files into simple config data because it is easier to maintain.

jandubois: If you add the `tpl` functionality, please also add a comment that explains that templates in config files only have access to global helm variables, plus the user-provided values from `values.yaml`. They are evaluated *before* all the config files are merged into `$.Values`, so none of the internal config data can be used.

jandubois: > none of the internal config data can be used

Otherwise config files would become self-referential. We won't allow that to make sure our helm charts don't become self-aware. Given the complexities of the templates this could be a real danger... 😛

andreas-kupries: @jandubois note that the insertion of `tpl` into the pipeline does not work:
```
executing "_config.load" at <tpl>: wrong number of args for tpl: want 2 got 1
```

andreas-kupries: It requires a path as first argument ? The name of the file we are processing ?

jandubois: > It requires a path as first argument ? The name of the file we are processing ?

No, it requires the context as the second argument. Try this:

```
{{- $_ := set $configs $name (tpl ($bytes | toString) $) | fromYaml) }}
```

andreas-kupries: After reading docs the proper incantation looks to be
```
{{- $_ := set $configs $name (tpl ($bytes | toString) $ | fromYaml)}}
```

andreas-kupries: Done with commit ccd904248b11b735a4df5c43391c5984a486119f

chart/templates/_config.tpl
```diff
@@ -46,7 +46,14 @@
 
     {{- $configs := dict }}
     {{- range $name, $bytes := $.Files.Glob "config/*" }}
-      {{- $_ := set $configs $name ($bytes | toString | fromYaml) }}
+      {{/*
+      **** !Attention! Templates in config files have only access to the global helm variables,
```
jandubois, line 50: I was thinking that you would add this to the documentation of this function at the top, and not in the middle of the function): explain that config files are evaluated as templates, but that the available data is limited.

chart/templates/_config.tpl
```diff
@@ -46,7 +46,14 @@
...
+      **** and the user-provided values from `values.yaml`. As they are evaluated *before* all
+      **** the config files are merged into `$.Values`, none of the internal config data can be
+      **** used.
+      ****
+      */}}
+      {{- $_ := set $configs $name (tpl ($bytes | toString) $ | fromYaml)}}
```
jandubois, line 56: Nit-picking:

```suggestion
      {{- $_ := set $configs $name (tpl ($bytes | toString) $ | fromYaml) }}
```

chart/templates/_config.tpl
andreas-kupries: Will change

andreas-kupries: Done: 4b1b0cf0e199ac06bc5bc9dddef10e3bf29f30a0

chart/assets/operations/instance_groups/api.yaml
andreas-kupries: Fixed with 313c1b19f22c980f80ff3cd094f1ffd811409410

jandubois (dismissed):

scripts/helmlint.sh
```diff
@@ -17,9 +17,12 @@ helm lint "${TARGET_DIR}" --set system_domain=example.com
 # Running json schema validator a second time to also check the embedded config files.
 # Done in a separate run because the config files are merged here on top of values.yaml,
 # which will cause replacement of any arrays.
+#
+# NOTE: Excluded api.yaml, as a config file with templating.
```
jandubois, line 21: Should this comment be removed now that config files with templating are excluded generically

andreas-kupries: Yes.

andreas-kupries: Not removed, updated, commit 668858cb443f059c793cf7a319961f9397d30c2f

jandubois (approved):

PR #1411, opened 2020-09-30
fix: limit reuse of config values

jandubois:
## Description

This PR blocks the use of --reuse-values when upgrading from 2.5.0 or earlier because those release may include release-specific information (image names and versions etc), or may have persisted changes made by the templating code.

Upgrades from 2.4.0 and earlier are blocked via an `unsupported` rule that checks for the `eirini.env.DOMAIN` property that has been removed in 2.5.0.

Upgrades from 2.5.0 using `--reuse-values` are blocked via schema validation that does not allow limits or requests to be `nil`. Upgrades from 2.5.0 using the existing `values.yaml` file are not blocked and should be benign as the version specific settings have already been moved to `config/*.yaml` files.

## Motivation and Context

We have run into problems with `--reuse-values` in our own CI and we want prevent users from running into the same traps.

## How Has This Been Tested?

It has not yet been tested.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/chart/config/unsupported.yaml
+++ b/chart/config/unsupported.yaml
@@ -20,3 +20,9 @@ unsupported:
   : >
     The database must be enabled via either features.embedded_database or
     features.external_database, or this must be the cell segment in a multi-cluster deployment.
+
+  # This setting was in values.yaml in kubecf 2.4.0 and earlier; it no longer exists:
+  eirini.env.DOMAIN: |
+    You are reusing an old and incompatible configuration, either via a values.yaml file from an
+    older release, or by using the --reuse-values option. Please start with a fresh configuration,
+    only changing settings relative to the defaults in the current release.
--- a/chart/templates/_config.tpl
+++ b/chart/templates/_config.tpl
@@ -47,6 +47,9 @@
     {{- $_ := set $ "kubecf" dict }}
   {{- end }}
   {{- if not $.kubecf.manifest }}
+    {{- /* copy all structures under $.Values so our changes do not affect --reuse-values */}}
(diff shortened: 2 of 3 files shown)


chart/templates/_config.tpl
```diff
@@ -47,6 +47,9 @@
     {{- $_ := set $ "kubecf" dict }}
   {{- end }}
   {{- if not $.kubecf.manifest }}
+    {{- /* copy all structures under $.Values so our changes do not affect --reuse-values */}}
+    {{- $_ := set $ "Values" (deepCopy $.Values) }}
```
mook-as, line 51:
```suggestion
    {{- $_ := deepCopy $.Values | set $ "Values" }}
```
No need to take this one — it just felt odd approving something with no comments 😁

mook-as (approved):
👍

chart/config/unsupported.yaml
```diff
@@ -20,3 +20,9 @@ unsupported:
   : >
     The database must be enabled via either features.embedded_database or
     features.external_database, or this must be the cell segment in a multi-cluster deployment.
+
+  # This setting was in values.yaml in kubecf 2.4.0 and earlier; it no longer exists:
+  eirini.env.DOMAIN: |
+    You are reusing an old and incompatible configuration, either via a values.yaml file from an
+    older release, or by using the --reuse-values file. Please start with a fresh configuration,
```
andreas-kupries, line 27: Nit: file -> option.

andreas-kupries (commented):
Just a comment nit found.

andreas-kupries (approved):
LGTM

PR #1413, opened 2020-09-30
fix: Increase default memory limit for addon processes from 32Mi to 64Mi

jandubois:
This is the same default we already use for all regular processes that don't have an explicit limit defined.

We've seen at least the syslog forwarder getting occasionally OOMKilled with the 32Mi limit.


andreas-kupries (approved):
:+1:

mook-as (approved):
/rubber-stamp

PR #1419, opened 2020-10-01
Revert "feat: support seeding external database"

viovanov:
Reverts cloudfoundry-incubator/kubecf#1388


mook-as (approved):
Approving assuming we're just trying to get a CI run — please don't actually merge

mook-as:
Rejecting and closing: reproduced the issue† locally, and the problem was that the container was getting OOMKilled.

† Triggered the rotate-cc-database-key QuarksJob, and watched it fail.

PR #1430, opened 2020-10-02
fix: Remove allowance for old catapult cruft

jandubois:
Catapult (used to) add additional properties under the `kube` key that are completely ignored by kubecf.

It is my understanding that these settings have been removed from catapult, so we should tighten the schema so it will catch misspelled property names provided by actual users.

### DIFF
--- a/chart/values.schema.yaml
+++ b/chart/values.schema.yaml
@@ -332,8 +332,7 @@ properties:
           default: {"$ref": "#/definitions/string_or_null"}
         additionalProperties: false
       storage_class: {"$ref": "#/definitions/string_or_null"}
-    # XXX Catapult sets additional (unused) properties; allow them for now to pass CI
-    additionalProperties: true
+    additionalProperties: false
 
   move_jobs:
     type: object


mook-as (approved):
Approving to run on CI

jandubois:
> Approving to run on CI

Well, if it passes on CI it proves that catapult no longer uses the obsolete properties, so it can be merged.

Did you have any concerns about merging this?

mook-as (dismissed):
Well I guess that didn't work.

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
kubecf:
- kube: Additional property service_cluster_ip_range is not allowed
- kube: Additional property organization is not allowed
- kube: Additional property pod_cluster_ip_range is not allowed
- kube: Additional property registry is not allowed
```

jandubois:
That's good to know, so the cruft is still there. We'll leave this PR open until catapult is fixed, and then trigger it again...

Will create a card for fixing catapult.

viccuad:
Catapult fixed, this should be able to go through now.

viccuad:
While the title is factually correct, perhaps it would be nice to rename it to `feat: Enable schema values checking for unused properties` (or something similar)?

mook-as (approved):

PR #1223, opened 2020-08-17
RFD-0000: Proposal to Include RFDs in Future System Changes

fargozhu:
## Description

The goal is to have an constructive discussion within the team about an idea and capture it in a
document known by RFD.

## Motivation and Context

It's difficult to have a proper discussion around an idea or intention by email or slack without loosing context and important details.
Also, it's difficult to trace back to some of our decisions, if we do not have a proper way to document it for the future.

## How Has This Been Tested?

This proposal is itself an RFD so we can not only improve it before moving forward but also to test the process itself.

PR #1438, opened 2020-10-05
fix: add GitHub issue as discussion place

fargozhu:
## Description

Define the RFD issue as the right place for discussion.

## Motivation and Context

We should separate the conversation about the topic that should happen in a GitHub _issue_, from the review of the document itself that should happen on a _pull request_ thread, since a _pull request_ intention is to review code changes or, like in this case, document modifications.

### DIFF
--- a/doc/rfd/rfd/0000/README.md
+++ b/doc/rfd/rfd/0000/README.md
@@ -69,7 +69,8 @@ Some of these steps may have some script support in the future to make everyone'
 
 ### Find Next RFD Number
 
-Check for the next RFD number by going through the RFDs table.
+Check for the next RFD number by going through the RFDs table and on
+[GitHub](is:pr "feat: RFD" in:title) 
 
 ### Create a Branch & Placeholder
 
--- a/doc/rfd/rfd/0000/README.md
+++ b/doc/rfd/rfd/0000/README.md
@@ -113,7 +113,7 @@ and push it remotely.
 ```
 
 Do not forget to open a pull request (maybe later we will have a bot for this) with the title
-_RDF:num_ .
+_feat: RDF num title_ .
 
 With the pull request open anyone subscribed to the repository will be notified about the new RFD
 and it can gather feedback through pull request comments.
--- a/doc/rfd/rfd/0000/README.md
+++ b/doc/rfd/rfd/0000/README.md
@@ -16,6 +16,8 @@ document known by RFD.
 1. authors: owners of the RFD must be listed with their name and email address
 (John Doe <john.doe@wonder.land>).
 2. state: one of the states discussed below.
+3. discussion: GitHub filter link that list both the pull request and the issue related with the RFD
+(i.e. https://github.com/cloudfoundry-incubator/kubecf/issues?q=%22RFD+8%22+)
 
 ## State
 
--- a/doc/rfd/rfd/0000/README.md
+++ b/doc/rfd/rfd/0000/README.md
@@ -112,8 +112,8 @@ and push it remotely.
 # git push origin rfd-0010
 ```
 
-Do not forget to open a pull request (maybe later we will have a bot for this) with the title
-_feat: RDF num title_ .
+Do not forget to open a pull request and a GitHub issue where the discussion will happen with the
+title _feat: RDF num title_ .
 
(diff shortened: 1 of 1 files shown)


doc/rfd/rfd/0000/README.md
```diff
@@ -67,7 +69,8 @@ Some of these steps may have some script support in the future to make everyone'
 
 ### Find Next RFD Number
 
-Check for the next RFD number by going through the RFDs table.
+Check for the next RFD number by going through the RFDs table and on
+[GitHub](is:pr "feat: RFD" in:title)
```
mook-as, line 73: This link appears to be broken.

mook-as (changes requested):
Hmm, is PR review inadequate for discussion? #1223 seems to have decent discussion that clarified the intent and got us onto (mostly) the same page.

For clarity (and to make it easier to understand why we made the decision in the future), please explicitly document the reasoning behind the change?

fargozhu:
> Hmm, is PR review inadequate for discussion? #1223 seems to have decent discussion that clarified the intent and got us onto (mostly) the same page.
> 
> For clarity (and to make it easier to understand why we made the decision in the future), please explicitly document the reasoning behind the change?

@mook-as I did some changes to the pull request description that it may answer to your question.

mook-as:
@fargozhu Hmm, if you mean the "We should separate…" text: that doesn't explain the reasoning behind it, as in, _why_ we should do so, just prescribes the result.  It's important for us to keep track of the reasoning behind our decisions so that when we revisit them later (because we do that a lot), we can just confirm that the previous situation still applies and skip the re-reasoning part.  This is the same reason we should document the _whys_ in our commit messages, so that when we look at it later (because there's some weird bug) we can understand the situation we were in, so we can be sure any changes we make later will not break the existing cases (or if, instead, the existing cases are no longer relevant).

fargozhu:
@mook-as I think the reason is clear: separation of concerns:

1. discussion: should happen on a GitHub issue conversation thread.
2. review: should happen on a GitHub pull request.

mook-as:
@fargozhu but I don't understand _why_ those should be separate, especially given that the documented workflow says to create the discussion at the same time as the PR?  I could see a separate discussion being useful to congeal the idea before having a RDF (roughly, python-ideas@ → PEP); but since here we're creating both at the same time, and issues and PRs are mostly the same thing in GitHub, it's unclear to me what a separate issue would be tracking.

It may be useful to expand more on what the difference in types of comments you expect are between the issue and the review.

fargozhu:
@mook-as I am closing for now this _pull request_ because I think I should collect more user comments based on experience  and that will only happen when have more people using the RFD process.

Thanks for all your comments and I am pretty sure we will come back to this topic soon ;)

mook-as:
Hmm, can we automate our way out of this instead?

1. _[No Changes]_ Find RFD number
    - Actually maybe we can skip this step.
2. _[No Changes]_ Create a Branch & Placeholder
3. _[No Changes]_ Push Branch Remotely
4. _[No Changes]_ Open a Pull Request
   - Discussion happens here
5. _[No Changes]_ Pull Request approved
6. GitHub Actions runs on [`pull_request_review`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_review):
   1. Check that the PR only modifies `/doc/rfd/…`
   2. Check that the PR was approved
   3. Update the RFD number (and maybe discussion URL to the target PR?)
   4. Update the RFD index
   5. Automatically merge

I want to avoid having pending discussions within the repo, because in that case having a canonical discussion URL will be difficult.  Also, automating things means removing the ability for people to screw up (because I know I'm likely to do that) :D

fargozhu:
@mook-as I like the process and the automation can come immediately after. 

I am just trying to figure out where the RFD index is stored? Do we keep it in a _.<index-file>_ on the RFD root or, do we perform a lookup?

mook-as:
I thought we already had https://github.com/cloudfoundry-incubator/kubecf/blob/master/doc/rfd/README.md? That was the file I was thinking of when I wrote about updating the index, at least.

fargozhu:
@mook-as I'll close this one and create a new one for the automation work.

fargozhu:
Closing. Check my previous comment.

Issue #1371, opened 2020-09-24
Eirini brain tests are failing


satadruroy:
KubeCF 2.5/Eirini/Ingress Controller
EKS/k8s 1.17 

```
Tests complete: 7 Passed, 2 Skipped, 5 Failed                                                                                          │
│   Skipped tests:                                                                                                                       │
│     017_syslog_forwarding_test.rb                                                                                                      │
│     011_nfspersi_test.rb                                                                                                               │
│   Failed tests:                                                                                                                        │
│     010_insecure_registry_test.rb with exit code 1                                                                                     │
│     004_tcprouting_test.rb with exit code 1                                                                                            │
│     005_metron_test.rb with exit code 1                                                                                                │
│     018_autoscaler_test.rb with exit code 1                                                                                            │
│     013_credhub_test.rb with exit code 1 
```

Autoscaler failures are intermittent but `insecure_registry`, `tcp_routing` and `metron` test failures were also observed on AKS with Eirini.

andreas-kupries:
@satadruroy Do you have the full test output somewhere ? Please attach it to the ticket.

The excerpt in the description essentially misses all the details needed for proper debugging.
(IIRC when a test fails the brain runner dumps the entire log for that test into the final output, to enable post-mortem analysis).

andreas-kupries:
[EBRAIN.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5300673/EBRAIN.txt)

Brain logs from the referenced build. Not the full logs from concourse, just the part dealing with the brains tests.
It has the (expected) 4, 5, and 10 failing (tcp-routing, metron, insecure-registry).

andreas-kupries:
Wrt 004/tcp_routing the main error reported is
```
+ curl --fail -s -o /dev/null tcp-route-node-env-7516.ci-aks-9fec0ee3133b908c.susecap.net
+ curl tcp.ci-aks-9fec0ee3133b908c.susecap.net:20005
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (7) Failed to connect to tcp.ci-aks-9fec0ee3133b908c.susecap.net port 20005: Connection refused
Command exited with 7
```
This points towards bad/missing setup of proper tcp routing for eirini ?

andreas-kupries:
[...]

PR #1445, opened 2020-10-05
fix: brain tests for eirini

andreas-kupries:
## Description

- disable tcp routing tests, not supported
- bring in fix for metron test (brain test release 0.0.15)

## Motivation and Context

See #1371 

## How Has This Been Tested?

Local minikube deployment with eirini and diego. Latest master head, operator `6.1.10+0.g7b1a5f2f`.

## Types of changes

## Checklist:

### DIFF
--- a/chart/assets/operations/instance_groups/brain-tests.yaml
+++ b/chart/assets/operations/instance_groups/brain-tests.yaml
@@ -40,10 +40,20 @@
         tests:
           autoscaler:
             enabled: {{ .Values.features.autoscaler.enabled }}
+{{/* DROP when eirini has proper tcp routing support
+**** See https://github.com/cloudfoundry-incubator/kubecf/issues/1371#issuecomment-702883321
+*/}}
+{{- if .Values.features.eirini.enabled }}
+          tcprouting:
+            enabled: false
+          insecure_registry:
+            enabled: false
+{{- else }}
           tcprouting:
             enabled: {{ .Values.features.routing_api.enabled }}
           insecure_registry:
             enabled: {{ .Values.features.routing_api.enabled }}
+{{- end }}
           credhub:
             enabled: {{ .Values.features.credhub.enabled }}
       release: brain-tests
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -20,7 +20,7 @@ releases:
       tag: 0.1.0
   brain-tests:
     condition: testing.brain_tests.enabled
-    version: v0.0.14
+    version: v0.0.15
   bosh-dns-aliases:
     # not needed for kubecf; functionality provided by quarks-operator
     condition: false


jandubois (approved):
I wonder why we don't disable routing-api under Eirini instead? If it doesn't work, why let the user enable it?

@f0rmiga says:

> If the gorouter misses an update from nats, it can sync with the routing-api to keep the correct state.

But I don't understand how that would work if Eirini never emits any routes to routing-api in the first place.

f0rmiga:
> I wonder why we don't disable routing-api under Eirini instead? If it doesn't work, why let the user enable it?
> 
> @f0rmiga says:
> 
> > If the gorouter misses an update from nats, it can sync with the routing-api to keep the correct state.
> 
> But I don't understand how that would work if Eirini never emits any routes to routing-api in the first place.

It does emit routes. It just doesn't emit the TCP ones.

chart/assets/operations/instance_groups/brain-tests.yaml
```diff
@@ -40,10 +40,20 @@
...
+          tcprouting:
+            enabled: false
+          insecure_registry:
+            enabled: false
+{{- else }}
           tcprouting:
             enabled: {{ .Values.features.routing_api.enabled }}
           insecure_registry:
             enabled: {{ .Values.features.routing_api.enabled }}
+{{- end }}
```
f0rmiga, line 56: I _think_ this does what you want.

```suggestion
          tcprouting:
            enabled: {{ and .Values.features.routing_api.enabled (not .Values.features.eirini.enabled) }}
          insecure_registry:
            enabled: {{ and .Values.features.routing_api.enabled (not .Values.features.eirini.enabled) }}
```

chart/assets/operations/instance_groups/brain-tests.yaml
```diff
@@ -40,10 +40,20 @@
         tests:
           autoscaler:
             enabled: {{ .Values.features.autoscaler.enabled }}
+{{/* DROP when eirini has proper tcp routing support
```
f0rmiga, line 43:
```suggestion
{{/* Enable these tests when Eirini has proper TCP routing support.
```

f0rmiga (changes requested):

jandubois:
@f0rmiga says
> > > If the gorouter misses an update from nats, it can sync with the routing-api to keep the correct state.
> > 
> > But I don't understand how that would work if Eirini never emits any routes to routing-api in the first place.
> 
> It does emit routes. It just doesn't emit the TCP ones.

Still confused: Eirini emits routes to gorouter. Does it also emit routes to the routing-api? Or how does it learn about the routes? Is there some document that explains how this works?

f0rmiga:
@jandubois Hmmm indeed it doesn't publish to the routing-api, only to nats (which is what the gorouter uses). You can check what I'm saying here:
https://github.com/cloudfoundry-incubator/eirini/blob/3bc2929aa05d437520da7df2e214edd1af0c941a/route/emitter.go#L106
https://github.com/cloudfoundry-incubator/eirini/blob/3bc2929aa05d437520da7df2e214edd1af0c941a/route/emitter.go#L28
https://github.com/cloudfoundry-incubator/eirini/blob/3bc2929aa05d437520da7df2e214edd1af0c941a/route/emitter.go#L51

I reviewed https://docs.cloudfoundry.org/concepts/cf-routing-architecture.html and if the gorouter misses a message from nats, it will need to wait for another emit event from the route emitter.

jandubois:
> indeed it doesn't publish to the routing-api, only to nats

So doesn't it make more sense to disable `routing-api` until Eirini supports it?

In which case this PR should be replaced with an entry in `chart/config/unsupported.yaml` that displays an error when the user wants to enable routing-api with Eirini.

Next question: why do we even make routing-api configurable? why would someone want to disable it when running with Diego?

f0rmiga:
@jandubois Yes, it does make sense to disable routing-api - I agree with that. Having the option to disable it was a request from IBM (I think the PR to being able to disable it came from @qu1queee a while ago).

f0rmiga:
@andreas-kupries I'll reject this PR in favour of disabling routing-api when Eirini is enabled.

jandubois:
> Having the option to disable it was a request from IBM (I think the PR to being able to disable it came from @qu1queee a while ago).

Indeed, it was #544 

The problem I'm struggling with is that I would like the default to be `enabled` under Diego, but `disabled` under Eirini, but we have no good mechanism to do that (that isn't confusing to the user).

Otherwise you have to specify a lot of additional settings to enable Eirini:

```
helm install ... --set install_stacks={sle15} --set features.eirini.enabled=true --set features.routing_api.enabled=false
```

The best I can currently think of is leaving the flag as `nil` and adding a comment:

```
features:
  routing_api:
    # default is true for Diego, but false for Eirini
    enabled: ~
```

and then set the appropriate default in `chart/templates/_features.yaml`:

```
{{- if kindIs "invalid" $.Values.features.routing_api.enabled }}
  {{- $_ := set $.Values.features.routing_api "enabled" (not $.Values.features.eirini.enabled) }}
{{- end }}
```

Of course needs a corresponding change in the schema to allow the null value too.

Not really fond of it either, but nothing else comes to mind right now (well, except forcing the user to disable it under Eirini).

jandubois:
> # default is true for Diego, but false for Eirini

Actually, the comment should explain that there is no point in enabling it under Eirini, so something like

```
# enabled by default, except under Eirini, where the routing-api is not (yet) supported
```

andreas-kupries:
I believe I understand the discussion now. Will make a new PR based on @jandubois idea. As hackish as it looks it may be the only way of doing it.

PR #1452, opened 2020-10-06
fix: set diego-cell ephemeral PVC disk size

### LINKED ISSUE #1451: Using VPCs as ephemeral disk doesn't set disk size
**Describe the bug**
When the user sets `sizing.diego_cell.ephemeral_disk.storage_class`, we do not set the disk size, ending up with a default of 10Gi.  As the default grootfs disk size is 40Gi, this falls over terribly.

**To Reproduce**
Deploy (on AKS in my case) with: `--set sizing.diego_cell.ephemeral_disk.storage_class=default`

**Expected behaviour**
KubeCF deploys successfully.

**Actual behaviour**
`diego-cell` never gets ready, failing in `bpm-pre-start-garden` with:
> The node running this cell doesn't have enough disk space. You requested 40960M but the disk is 9952M in size.

Examining the resulting PVC, it's only 10Gi in size.

**Environment**
 - KubeCF Version: 2.5.5
 - Operator version: 6.1.11

**Additional context**
We probably need to set `instance_groups.*.vm_resources.ephemeral_disk_size`, according to [the docs](https://quarks.suse.dev/docs/quarks-operator/concepts/from_bosh_to_kube/).

mook-as:
## Description
When we're using PVCs for ephemeral disks for diego-cell, we will need to set the disk size.  The default size of 10Gi is too small to hold the default (grootfs) disk size.  Calculate it based on the user-specified ephemeral disk size, so that it can adjust according to user needs.

It needs to be slightly larger than the user specified size, as it is also used to hold other internal data; as an initial estimate, make it 5Gi larger than requested.  We may be able to adjust this down later, as for initial deployment we appear to only need <300Mi extra.

## Motivation and Context
Fixes #1451

## How Has This Been Tested?
Deployed on AKS in a situation that lead to the discovery of #1451 in the first place.
Checked that we have a generated PVC of 45Gi (when the default is 40Gi).

## Types of changes

## Checklist:

  - By basing it on the existing size, we do not need to change the documentation.

### DIFF
--- a/chart/assets/operations/instance_groups/diego-cell.yaml
+++ b/chart/assets/operations/instance_groups/diego-cell.yaml
@@ -28,15 +28,7 @@
   value: true
 
 # Configure the size of the diego cell grootfs store
-# We are repurposing reserved_space_for_other_jobs_in_mb as the size of the grootfs store 
-- type: replace
-  path: /instance_groups/name=diego-cell/jobs/name=garden/properties/grootfs?/reserved_space_for_other_jobs_in_mb
-  value: {{ .Values.sizing.diego_cell.ephemeral_disk.size }}
-
-- type: replace
-  path: /instance_groups/name=diego-cell/jobs/name=rep/properties/diego?/executor/disk_capacity_mb
-  value: {{ .Values.sizing.diego_cell.ephemeral_disk.size }}
-
+{{- $ephemeral_disk_size := .Values.sizing.diego_cell.ephemeral_disk.size }}
 {{- if .Values.sizing.diego_cell.ephemeral_disk.storage_class }}
 # Use a PVC for garden data
 - type: replace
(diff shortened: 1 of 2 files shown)


viovanov (approved):

f0rmiga (changes requested):
I don't like that it `adds` to what was requested. Any chance we can reduce the amount of space it uses based on the total we pass from values.yaml?

mook-as:
@f0rmiga As in, set `/instance_groups/name=diego-cell/jobs/name=rep/properties/diego?/executor/disk_capacity_mb` to `.Values.sizing.diego_cell.ephemeral_disk.size` - some amount?

f0rmiga:
@mook-as Yes

mook-as:
Okay, to expand on things from chat:
During upgrades, we don't (and likely can't, depending on the storage provider) resize PVCs.  So this change would only help new installs, and not upgrades.  So we should instead size the groot storage smaller, and keep the user input for the PVC size.

f0rmiga (approved):

PR #1455, opened 2020-10-06
feat: chart: add labels to QuarksStatefulSets

mook-as:
## Description
Add the standard labels to the generated QuarksStatefulSets, so that we can see the version when examining a live deployment easier.  The selector are unaffected, as we want to be able to use the previous version for a bit during an upgrade.

## Motivation and Context
When examining an upgrade, it is difficult to figure out what version of KubeCF corresponds to each pod.

## How Has This Been Tested?
Deployed during an upgrade (HA). I have confirmed that:
- The StatefulSets are updated, not recreated (their creation timestamp is from before the upgrade took place).
- The existing pods are rolling as expected — `api-1` is upgraded and ready before `api-0` is upgraded.

(Actually, the upgrade never completes because `api-1` is not counted for bootstrap, so DB migration never takes place, and therefore `api-1` can never get ready.  But that's an existing issue outside this PR.)

## Types of changes

## Checklist:

### DIFF
--- a/chart/assets/operations/instance_groups.yaml
+++ b/chart/assets/operations/instance_groups.yaml
@@ -1,0 +1,18 @@
+{{/*
+ This sets labels on the generated StatefulSets to include the information about
+ the helm chart.
+*/}}
+{{- include "_config.load" $ }}
+{{- range $ig_name, $ig := .Values.jobs }}
+  {{- $included := false }}
+  {{- range $job := $ig }}
+    {{- $included = list $ $job.condition | include "_config.condition" | eq "true" | or $included }}
+  {{- end }}
+  {{- if $included }}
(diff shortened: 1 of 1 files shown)


f0rmiga:
@mook-as Can this be an annotation rather than a label?

mook-as:
@f0rmiga Sure, but the actual labels are all pretty [standard](https://helm.sh/docs/chart_best_practices/labels/#standard-labels).  It's not really a problem to move them, but it seems odd to put them under annotations unlike every other resource.

f0rmiga:
> @f0rmiga Sure, but the actual labels are all pretty [standard](https://helm.sh/docs/chart_best_practices/labels/#standard-labels). It's not really a problem to move them, but it seems odd to put them under annotations unlike every other resource.

I see. Indeed we have the rest as labels.

f0rmiga (approved):

PR #1462, opened 2020-10-07
refactor!: make database a singleton

f0rmiga:
## Description

Removed the possibility of doing HA with the PXC database provided as part of KubeCF. It's currently not feasible to maintain a reliable HA setup as part of KubeCF.

## Motivation and Context

When the active node doing writes is terminated in the middle of replication (e.g. going from SA to HA as part of an upgrade), the data is left broken.

## How Has This Been Tested?

Locally. CI.

## Types of changes

## Checklist:


chart/templates/database/db-statefulset.yaml
```diff
@@ -10,35 +10,22 @@ metadata:
...
-        {{- $replicas = 3 }}
-      {{- end }}
-      {{- if eq (mod $replicas 2) 0 }}
-        {{- include "_config.fail" "The number of database instances must be odd to avoid split-brain" }}
-      {{- end }}
-      replicas: {{ $replicas }}
+      # Database replicas are not allowed.
+      # High-availability is not currently feasible to support in k8s.
+      # Galera replication on rolling updates of the statefulset is hard to get
+      # done right and to maintain as part of KubeCF. For an HA database, prefer
```
mook-as, line 23:
```suggestion
      # right and to maintain as part of KubeCF. For an HA database, prefer
```

chart/values.yaml
```diff
@@ -102,7 +102,6 @@ sizing:
   credhub:
     instances: ~
   database:
-    instances: ~
```
mook-as, line 105: Huh, the schema around sizing is a mess. Not for this PR, though.

mook-as (changes requested):
Please edit `unsupported.yaml` to explicitly let users know that this is no longer supported.

/cc @fargozhu for release notes.

chart/templates/database/db-statefulset.yaml
```diff
@@ -10,35 +10,22 @@ metadata:
...
+      # using a specialized external one.
+      replicas: 1
       selector:
         matchLabels:
-          {{- list . "database" | include "component.labels" | nindent 10 }}
+          {{- list . "database" | include "component.selector" | nindent 10 }}
```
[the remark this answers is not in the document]
f0rmiga: Ahhh I picked from your commit... I hope this is no problem. We can merge your PR first. This PR won't get a merge conflict, I think.

f0rmiga:
@mook-as Please, check https://github.com/cloudfoundry-incubator/kubecf/pull/1462/commits/b31d2156910b2cfc25c85b80422c845e245adc79 if it's the expected message.

chart/config/unsupported.yaml
```diff
@@ -26,3 +26,10 @@ unsupported:
     You are reusing an old and incompatible configuration, either via a values.yaml file from an
     older release, or by using the --reuse-values option. Please start with a fresh configuration,
     only changing settings relative to the defaults in the current release.
+
+  sizing.database.instances: |
+    Database replicas are not allowed.
+    High-availability is not currently feasible to support in k8s.
+    Galera replication on rolling updates of the statefulset is hard to get
+    right and to maintain as part of KubeCF. For an HA database, prefer
+    using a specialized external one.
```
mook-as, line 35: Hmm, for user-facing text we should probably avoid using abbreviations.
```suggestion
  # Do not allow replicas for the database; doing this well in Kubernetes is
  # more effort than we can expend, and is better handled by external databases.
  sizing.database.instances: >
    Database replicas are not supported.
    KubeCF does not support highly available database configurations;
    please use an external database instead.
```

We don't _really_ need to say why in the error message (but putting it in a comment near the message is probably useful).  Details about Galera etc. would definitely belong more in a comment, I think.

Hmm, should probably mention [the documentation](https://kubecf.io/docs/deployment/kubernetes-deploy/#external-database).

chart/templates/database/db-statefulset.yaml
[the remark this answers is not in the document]
mook-as: No worries; as long as the code lands it's fine. It just seemed like it was unrelated :)

mook-as (changes requested):
Nitpicks around phrasing.

jandubois:
Is it safe to downgrade the database from HA to SA when upgrading kubecf? If not, then maybe the "unsupported" error should include some guidance to scale down the database *before* upgrading? (Assuming that is safe; or can that too still lead to data loss?)

f0rmiga:
> Is it safe to downgrade the database from HA to SA when upgrading kubecf? If not, then maybe the "unsupported" error should include some guidance to scale down the database _before_ upgrading? (Assuming that is safe; or can that too still lead to data loss?)

@jandubois Going from HA to SA should be safe if the active replica is the `database-0`.
@viovanov Is there a way to instruct the active-passive controller to use the `database-0`?

viovanov:
@f0rmiga afaik it will always scale to the 0 ordinal

viovanov (approved):

PR #1467, opened 2020-10-08
fix: Use default stack list to compute imagelist.txt

jandubois:
Only disable cflinuxfs3 when Eirini is enabled. Otherwise we will be missing all the default stack images.


andreas-kupries (approved):
LGTM.

PR #1469, opened 2020-10-08
fix: disable routing api under eirini, not supported

andreas-kupries:
## Description

- `routing_api.enabled` flag defaults to `nil` now.
- Feature config handling now sets proper default per chosen scheduler.
- Rejects user trying to force `eirini` + `routing_api`.
- Values schema updated (now allowing `nil` for the enabled flag)
- `image_list.rb` script skips the forbidden permutation.

## Motivation and Context

See #1371 and discussion in #1445 

## How Has This Been Tested?

Local minikube

## Types of changes

## Checklist:

### DIFF
--- a/scripts/image_list.rb
+++ b/scripts/image_list.rb
@@ -93,6 +93,9 @@ class HelmRenderer
     # Eirini will throw an error unless a compatible stack is selected
     if values['features']['eirini']['enabled']
       values['install_stacks'] = ['sle15']
+      # Chart will throw an error when trying to use both eirini and
+      # routing_api. Avoid.
+      values['features']['routing_api']['enabled'] = false
     end
     Tempfile.open(['values-', '.yaml']) do |values_file|
       values_file.write values.to_yaml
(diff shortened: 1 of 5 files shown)


scripts/image_list.rb
```diff
@@ -277,6 +277,9 @@ def images
 # Iterate over all permutations, rendering the chart to obtain all possible
 # images.
 permutations.each do |permutation|
+  # Skip forbidden combination of routing_api and eirini
+  next if permutation['routing_api'] && permutation['eirini']
```
jandubois, line 281: This isn't working: we only run 2 permutations: all enabled, and all disabled. So you are skipping the "all enabled" option, and we will not record the images required for Eirini.

There is already a clause I added earlier where we set `install_stacks` to just `sle15`. Add a line there to force `routing_api` to disabled...

If you are feeling generous, maybe move the settings below for embedded and external database into the same location (but outside the Eirini check) too, so all the special cases are close to each other in the code.

jandubois (changes requested):

scripts/image_list.rb
jandubois: > we only run 2 permutations

https://github.com/cloudfoundry-incubator/kubecf/blob/master/scripts/image_list.rb#L270-L277

andreas-kupries: Fixed in commit 6744838b5442e8280e922c904b1038da0cabe333. Rebased to get image_list.rb changed. Fixed. Collapsed history, force pushed.

jandubois (approved):

PR #1474, opened 2020-10-12
fix: Add sle15 stack to stemcell-pipeline

thardeck:
## Description
The sle15 stack was not automatically updated. The buildpack releases are handled the same way as the stack so we can reuse the buildpack pipeline to update the stack automatically in the future.

## How Has This Been Tested?
The pipeline was updated and the resulting pr branch #1473 works locally. Only the Docker registry needs to changed once because we will not use dockerhub anymore.

## Types of changes

## Checklist:


jandubois (approved):

jandubois:
@thardeck We have only approved/merged this PR; from the description above I understand that the updated pipeline as already been deployed, so there should be nothing left to do.

thardeck:
> @thardeck We have only approved/merged this PR; from the description above I understand that the updated pipeline as already been deployed, so there should be nothing left to do.

Thanks, yes, I only had to switch the development branch in Concourse back to `master`.

Issue #1410, opened 2020-09-30
Split locket into a separate pod


andrew-edgar:
**Is your feature request related to a problem? Please describe.**
Diego-api pod has both BBS and locket.  The bbs is currently only talking using 127.0.0.1 which is not standard and causes custom certs to be created for kubecf to add that as an alternative name.   

There is no reason locket and bbs should be in the same pod and would allow them to be more separate and make this issue https://github.com/cloudfoundry-incubator/kubecf/issues/25 also work better.

**Describe the solution you'd like**
Have locket a separate pod from Diego-api

**Describe alternatives you've considered**
Need to add 127.0.0.1 to the locket certs when this is not defined in cf-deployment.

andrew-edgar:
With more investigation this is worse than I thought!   Right now `locket.service.cf.internal` will ONLY return the "active" diego-api.  This is incorrect.   locket server is not active/passive.  it is active/active and should as long as the locket container is running this DNS needs to return all of the diego-apis.

This is bad because we are no longer balancing this locket server load.  in a larger env we are stuck sending all locket requests to the same pod.  Which is poor use of resources.   I think this should be up the priority list as this is a regression from how bosh works for the locket server.

jandubois:
I'm in favour of this change. It is pretty much the same reason we split `routing-api` away from the `api` instance group, so `api` could be active/active while `routing-api` is active/passive.

And as already mentioned, it helps with Eirini: we can get rid of `bbs` and `silk-controller` jobs, and have `locket` in an instance group that doesn't have `diego` in its name.

I would prefer to do this using the quarks-specific `qmove` op instead of doing all the manual deployment manifest editing we do at build time currently. That would require https://github.com/cloudfoundry-incubator/quarks-operator/issues/1122 to be completed though; otherwise `scripts/image_list.rb` is going to break.

According to the Quarks team tracker this is scheduled for the sprint at the end of the month, but given that they dealt with a lot of unscheduled work lately, I suspect this might slip a bit. I would still prefer to wait for that.

PR #1480, opened 2020-10-14
refactor: Move jobs between instance groups in $.kubecf.manifest

jandubois:
This way we don't need to extract the job definitions during build time and import them again; everything can be modified in-tree, and in the end just the modified manifest is written to the configmap.

It felt silly to create an update.serial setting for newly created instance groups, just to delete it again in the sequencing.yaml
file via ops, so the whole handling of update.serial has been simplified as well.

This will simplify implementation of #1410 as well


chart/templates/_jobs.tpl
```diff
@@ -2,26 +2,26 @@
...
 +-----------------------------------------------------------------------------------------
+| * Move some jobs from their manifest locations to other instance groups.
 | * Create $.Values.jobs and $.Values.addon_jobs from the deployment manifest.
 | * Define CC workers based on their property settings in both "jobs" and "resources" trees.
-| * Move some jobs from their manifest locations to other instance groups.
 ==========================================================================================
 */}}
 {{- define "_jobs.update" }}
+  {{- /* *** Move some jobs to new instance groups *** */}}
+  {{- range $from_ig, $ig := $.Values.move_jobs }}
```
mook-as, line 12: Given the things we've been doing, I wonder if this needs to be under `$.Values` at all (rather than something in `$.kubecf` that isn't modifiable by the user).

Probably not this PR, though.

Do please update the comment in [`chart/config/jobs.yaml`](https://github.com/cloudfoundry-incubator/kubecf/blob/7762c65756fde37a126f1f5e701bb6056c93f55a/chart/config/jobs.yaml#L6) to drop the reference to `qmove`; with this change I'm pretty sure we want to get _away_ from doing that.

chart/templates/_jobs.tpl
```diff
@@ -137,31 +137,49 @@
...
+  {{- $root := index . 0 }}
+  {{- $from_ig_name := index . 1 }}
+  {{- $to_ig_name := index . 2 }}
+  {{- $job_name := index . 3 }}
+
+  {{- /* Locate $from_ig in the manifest */}}
+  {{- $_ := include "_config.lookupManifest" (list $root "instance_groups" $from_ig_name) }}
+  {{- $from_ig := $root.kubecf.retval }}
+  {{- if not $from_ig }}
+    {{- include "_config.fail" (printf "Instance group %q does not exist" $from_ig_name) }}
```
mook-as, line 156: Would a more detailed error be useful?

```suggestion
    {{- include "_config.fail" (printf "Could not find instance group %q while moving job %q" $from_ig_name $job_name) }}
```

(This depends on if we get a useful stack trace when things go sideways.)

mook-as (approved):
This is basically fine, just some trivial changes.

Approving now, since we appear to have no in-flight clusters in CI, so we might as well get a run out of it.

chart/templates/_jobs.tpl
jandubois: I was originally hoping that you could move jobs into new instance groups during `helm install` time. I have a second commit that gets us 70%, but it adds a lot of complexity and templating logic for no real benefit, so I abandoned it.

So yes, the data could live somewhere else, but we only have a mechanism to load `config/*.yaml` files into `$.Values`, so it would require additional setup, just to hide the data from an **internal** config file, so I don't see the point.

I did remove the `qmove` reference; you are correct, we should no longer need the qcopy/qmove ops at all, but do the options by manipulating the manifest in memory, the way this PR already implements it.

chart/templates/_jobs.tpl
jandubois: We do get a full stacktrace from `_config.fail`, so it doesn't really matter. Since users are not expected (anymore) to even attempt to move jobs around, these errors also should only happen during development.

I changed this error message anyways, but left the other ones alone.

mook-as (approved):

Issue #1421, opened 2020-10-01
Implement imagePullSecrets


jandubois:
**Is your feature request related to a problem? Please describe.**

It should be possible to pull kubecf from a private registry requiring authentication. Or just a subset of images, like rootfs and buildpacks for a custom stack.

**Describe the solution you'd like**

We should support a global list of image pull secret names, say `kube.image_pull_secrets`. This would be a list of names, not of `imagePullSecret` objects (which are just maps with a single `name` element). This would allow the user to deploy with:

```
helm install ... --set kube.image_pull_secrets={foo,bar}
```

The secrets themselves will have to be created by the user *before* installing kubecf. We don't want to create the secrets as part of the kubecf helm chart because we don't want to make any assumption about how the secrets need to be specified.

These secrets will be added to the `default` service account. Since the namespace and service account already exist, it will be necessary to patch it using a helm pre-install hook. See for example  [helm patch default service account](https://stackoverflow.com/a/55959894/284266).

This also needs to be done to the service account of the Eirini namespace because the staging image will be pulled using that account.

The new `kube.image_pull_secrets` property needs to be added to the `values.schema.yaml` file as an "array of strings".

**Describe alternatives you've considered**

I originally considered defining `imagePullSecrets` for each release (and each non-bosh container image), and then setting them using the cf-operator facilities. However, `imagePullSecrets` are declared on the pod and not the container, so we would need to loop through all the jobs in an instance group to determine the list or required secrets. And for non cf-deployment based releases we only have the job name, but not the release name.

So this all gets pretty complex very quickly, but doesn't seem to provide any benefits over adding all the secrets to the service account, so I decided we should pursue the simpler alternative.

**Additional context**

cc @jbuns

jbuns:
Tying it to the Issue I created a few months back with:
https://github.com/cloudfoundry-incubator/kubecf/issues/1003

jandubois:
Implemented by #1482

PR #1482, opened 2020-10-14
feat: pull kubecf from a private registry requiring authentication

jbuns:
Allow users to define a list of imagepullsecret names allowing kubecf to pull from private registries requiring authentication.

## Description
Allow users to define a list of imagepullsecret names. Secrets would have to be created prior. These imagepullsecret names are then added to the serviceaccounts used by kubecf, allowing kubecf pods to pull from private registries requiring authentication. The serviceaccounts updated are:
```
default (namespace kubecf)
eirinix
opi
eirini-events
eirini-metrics
eirini-routing
eirini-staging-reporter
```
The `default` serviceaccount is updated with imagepullsecrets using helm/pre-install method. While the rest are updated by adding the entries to each of the corresponding template serviceaccount.

## Motivation and Context
Allows kubecf to pull from private registries requiring authentication.
Related issue: https://github.com/cloudfoundry-incubator/kubecf/issues/1421

## How Has This Been Tested?
Tested locally using kubecf version `2.5.8` and cf-operator version `6.1.14+0.g22ccdd3a` on GKE. Details about creating the docker-registry secret, updating serviceaccounts with imagepullsecrets, doing kubecf helm install, and proof of kubecf pods pulling from a private registry are here:
[test-kubecf-install-with-imagepullsecrets.txt](https://github.com/cloudfoundry-incubator/kubecf/files/5377464/test-kubecf-install-with-imagepullsecrets.txt)

## Types of changes

## Checklist:

### DIFF
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -18,6 +18,8 @@ releases:
     image:
       repository: ghcr.io/cfcontainerizationbot/kubecf-apps-dns
       tag: 0.1.0
+  # Can also use image `google/cloud-sdk` as an alternative image to provide
+  # kubectl capability
   kubecf_kubectl:
     image:
       repository: ghcr.io/cfcontainerizationbot/kubecf-kubectl
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -18,6 +18,10 @@ releases:
     image:
       repository: ghcr.io/cfcontainerizationbot/kubecf-apps-dns
       tag: 0.1.0
+  kubecf_kubectl:
+    image:
+      repository: ghcr.io/cfcontainerizationbot/kubecf-kubectl
+      tag: v1.19.2
   brain-tests:
     condition: testing.brain_tests.enabled
     version: v0.0.15
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -86,16 +86,13 @@ helm.sh/chart: {{ include "kubecf.chart" $root }}
 
 {{- /*
 ==========================================================================================
-| kubecf.imagePullSecrets (list $ $component)
-+-----------------------------------------------------------------------------------------
 | Add imagePullSecrets to service accounts.
 ==========================================================================================
 */ -}}
 {{- define "kubecf.imagePullSecrets" }}
-{{- $root := first . }}
-{{- range $secret_name := $root.Values.kube.image_pull_secrets }}
-- name: {{ $secret_name | quote }}
-{{- end }}
+  {{- range $secret_name := .Values.kube.image_pull_secrets }}
+    - name: {{ $secret_name | quote }}
+  {{- end }}
 {{- end }}
 
 {{- /*
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -85,6 +85,20 @@ helm.sh/chart: {{ include "kubecf.chart" $root }}
 {{- end }}
 
 {{- /*
+==========================================================================================
+| kubecf.imagePullSecrets (list $ $component)
++-----------------------------------------------------------------------------------------
+| Add imagePullSecrets to service accounts.
+==========================================================================================
+*/ -}}
+{{- define "kubecf.imagePullSecrets" }}
+{{- $root := first . }}
+{{- range $secret_name := $root.Values.kube.image_pull_secrets }}
+- name: {{ $secret_name | quote }}
+{{- end }}
+{{- end }}
+
+{{- /*
   Template "kubecf.dig" takes a dict and a list; it indexes the dict with each
   successive element of the list.
 
--- a/chart/templates/patch_serviceaccount.yaml
+++ b/chart/templates/patch_serviceaccount.yaml
@@ -1,5 +1,5 @@
 {{- include "_config.load" $ }}
-{{- if gt (len .Values.kube.image_pull_secrets) 0 }}
+{{- if .Values.kube.image_pull_secrets }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
@@ -7,7 +7,7 @@ metadata:
     "helm.sh/hook": pre-install
     "helm.sh/hook-weight": "-5"
     "helm.sh/hook-delete-policy": hook-succeeded
-  name: default
+  name: patch-sa
   namespace: {{ .Release.Namespace | quote }}
 rules:
 - apiGroups:
@@ -25,12 +25,12 @@ metadata:
     "helm.sh/hook": pre-install
     "helm.sh/hook-weight": "-5"
     "helm.sh/hook-delete-policy": hook-succeeded
-  name: default
+  name: patch-sa
   namespace: {{ .Release.Namespace | quote }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
-  name: default
+  name: patch-sa
 subjects:
 - kind: ServiceAccount
   name: default
@@ -50,15 +50,17 @@ spec:
       restartPolicy: Never
       containers:
       - name: sa
-        image: google/cloud-sdk
+        {{- with $image := $.Values.releases.kubecf_kubectl.image }}
+        image: {{ printf "%s:%s" $image.repository $image.tag | quote }}
+        {{- end }}
         imagePullPolicy: IfNotPresent
 # set image_pull_secrets list
-{{- $image_pull_secret_names := "" }}
+{{- $image_pull_secret_names := list }}
 {{- range $secret_name := $.Values.kube.image_pull_secrets }}
-  {{- $image_pull_secret_names = cat $image_pull_secret_names (printf "{\"name\":%s}," ($secret_name | quote)) }}
+  {{- $image_pull_secret_names = append $image_pull_secret_names (dict "name" $secret_name) }}
 {{- end }}
         command: ["/bin/sh", "-c"]
         args:
         - |
-          kubectl patch serviceaccount default -p '{"imagePullSecrets": [{{ $image_pull_secret_names | trimSuffix "," }} ] }' -n {{ .Release.Namespace }}
+          kubectl patch serviceaccount default -p '{"imagePullSecrets": {{ $image_pull_secret_names | toJson }} }' -n {{ .Release.Namespace }}
 {{- end }}
--- a/chart/templates/patch_serviceaccount.yaml
+++ b/chart/templates/patch_serviceaccount.yaml
@@ -54,11 +54,11 @@ spec:
         image: {{ printf "%s:%s" $image.repository $image.tag | quote }}
         {{- end }}
         imagePullPolicy: IfNotPresent
-# set image_pull_secrets list
-{{- $image_pull_secret_names := list }}
-{{- range $secret_name := $.Values.kube.image_pull_secrets }}
-  {{- $image_pull_secret_names = append $image_pull_secret_names (dict "name" $secret_name) }}
-{{- end }}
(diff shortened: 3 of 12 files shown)


chart/templates/patch_serviceaccount.yaml
```diff
@@ -0,0 +1,64 @@
+{{- include "_config.load" $ }}
+{{- if gt (len .Values.kube.image_pull_secrets) 0 }}
```
jandubois, line 2: This is fine, but could be simplified (because an empty list is `false`) to:

```
{{- if .Values.kube.image_pull_secrets }}
```

chart/templates/patch_serviceaccount.yaml
```diff
@@ -0,0 +1,64 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
+  name: default
```
jandubois, line 10: Maybe use a more descriptive name, like `patch-sa`? Same for the RoleBinding below.

chart/templates/patch_serviceaccount.yaml
```diff
@@ -0,0 +1,64 @@
...
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: sa
+        image: google/cloud-sdk
+        imagePullPolicy: IfNotPresent
+# set image_pull_secrets list
+{{- $image_pull_secret_names := "" }}
+{{- range $secret_name := $.Values.kube.image_pull_secrets }}
+  {{- $image_pull_secret_names = cat $image_pull_secret_names (printf "{\"name\":%s}," ($secret_name | quote)) }}
```
jandubois, line 58: See below for recommended rewrite. Still a tip if you need similar code in the future: `%q` in `printf` automatically quotes the value for you: `(printf "{%q:%q}" "name" $secret_Name)` would be clearer to read (for me, at least).

mixins/bits/templates/bits-service-account.yml
```diff
@@ -6,6 +6,12 @@ kind: ServiceAccount
 metadata:
   name: bits-service
   namespace: {{ .Release.Namespace }}
+{{- if gt (len .Values.kube.image_pull_secrets) 0 }}
+imagePullSecrets:
+  {{- range $secret_name := $.Values.kube.image_pull_secrets }}
+  - name: "{{ $secret_name }}"
```
jandubois, line 12: It doesn't matter here because kube object names cannot have embedded double-quotes, but I feel `{{ $secret_name | quote }}` is idiomatic because it will be correct even when the variable value includes quotes itself.

Also add `| uniq` in the `range` expression above.

mixins/bits/templates/bits-service-account.yml
```diff
@@ -6,6 +6,12 @@ kind: ServiceAccount
...
+  {{- end }}
+{{- end }}
```
jandubois, line 14: This block of code is repeated often enough that you should define a template in `chart/template/_helpers.tpl` and use that in all places. Look at `component.labels` in that file for an example.

I can't think of a good name right now, so I would call it `kubecf.imagePullSecrets` unless you can think of something better.

jandubois (changes requested):

chart/templates/patch_serviceaccount.yaml
```diff
@@ -0,0 +1,64 @@
...
+{{- end }}
+        command: ["/bin/sh", "-c"]
+        args:
+        - |
+          kubectl patch serviceaccount default -p '{"imagePullSecrets": [{{ $image_pull_secret_names | trimSuffix "," }} ] }' -n {{ .Release.Namespace }}
+{{- end }}
```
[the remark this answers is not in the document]
jandubois: > I also added `| uniq` ...

Please ignore; I was confused and thought the list of secrets was assembled from different sources. But since they all need to be specified during deployment, the schema already guarantees that the elements are unique.

chart/templates/patch_serviceaccount.yaml
jbuns: confirmed

chart/templates/patch_serviceaccount.yaml
jbuns: confirmed

chart/templates/patch_serviceaccount.yaml
```diff
@@ -0,0 +1,64 @@
...
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: sa
+        image: google/cloud-sdk
```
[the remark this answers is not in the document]
jbuns: confirmed

mixins/bits/templates/bits-service-account.yml
jbuns: confirmed

jbuns (commented):
Fixed.

chart/config/releases.yaml
```diff
@@ -18,6 +18,10 @@ releases:
     image:
       repository: ghcr.io/cfcontainerizationbot/kubecf-apps-dns
       tag: 0.1.0
+  kubecf_kubectl:
+    image:
+      repository: ghcr.io/cfcontainerizationbot/kubecf-kubectl
```
jandubois, line 23: I think I like leaving a comment here to point to `google/cloud-sdk` as an alternative image. Just to leave some breadcrumbs if anyone ever runs into the situation where `kubecf-kubectl` has been accidentally put into a private repo as well.

chart/templates/_helpers.tpl
```diff
@@ -84,6 +84,20 @@ app.kubernetes.io/version: {{ default $root.Chart.Version $root.Chart.AppVersion
 helm.sh/chart: {{ include "kubecf.chart" $root }}
 {{- end }}
 
+{{- /*
+==========================================================================================
+| kubecf.imagePullSecrets (list $ $component)
```
jandubois, line 89: You don't use the `$component` argument in the template, so you could just pass in `$` (not as a list) and get rid of `$root` inside as well.

chart/templates/_helpers.tpl
```diff
@@ -84,6 +84,20 @@ app.kubernetes.io/version: {{ default $root.Chart.Version $root.Chart.AppVersion
...
++-----------------------------------------------------------------------------------------
+| Add imagePullSecrets to service accounts.
+==========================================================================================
+*/ -}}
+{{- define "kubecf.imagePullSecrets" }}
+{{- $root := first . }}
```
jandubois, line 95: I wish we would properly indent templating logic everywhere.

I know the rest of this file doesn't use indentation either, so don't change anything. One day I'll rage-reindent this as a refactoring... 😄

chart/templates/patch_serviceaccount.yaml
```diff
@@ -0,0 +1,66 @@
...
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: sa
+        {{- with $image := $.Values.releases.kubecf_kubectl.image }}
+        image: {{ printf "%s:%s" $image.repository $image.tag | quote }}
+        {{- end }}
+        imagePullPolicy: IfNotPresent
+# set image_pull_secrets list
```
jandubois, line 57: I think this block should either be moved to the top of the "file" (in front of the `apiVersion` line), or indented with the rest of the YAML structure.

chart/templates/patch_serviceaccount.yaml
```diff
@@ -0,0 +1,66 @@
...
+{{- $image_pull_secret_names := list }}
+{{- range $secret_name := $.Values.kube.image_pull_secrets }}
+  {{- $image_pull_secret_names = append $image_pull_secret_names (dict "name" $secret_name) }}
+{{- end }}
+        command: ["/bin/sh", "-c"]
+        args:
+        - |
+          kubectl patch serviceaccount default -p '{"imagePullSecrets": {{ $image_pull_secret_names | toJson }} }' -n {{ .Release.Namespace }}
```
jandubois, line 65: I've just been wondering what happens if the service account already has `imagePullSecrets` defined: they will just get overwritten.

A quick search doesn't find any good way to append to a list. I found [this](https://stackoverflow.com/a/63438403/284266) on StackOverflow, but I'm not sure if this works when the list is empty initially.

I think it is fine to not do anything about this, but maybe add a comment that any existing secrets on the SA will also have to be specified again via the Helm setting.

mixins/eirini/templates/metrics-service-account.yml
```diff
@@ -7,6 +7,10 @@ kind: ServiceAccount
 metadata:
   name: eirini-metrics
   namespace: {{ .Release.Namespace }}
+{{- if .Values.kube.image_pull_secrets }}
+imagePullSecrets:
+{{- list . "" | include "kubecf.imagePullSecrets" }}
+  {{- end }}
```
jandubois, line 13: I would rather see the middle line indented than the closing `{{- end }}` 😄 (Yeah, I know this was just a typo)

jandubois (commented):
Thanks for making the changes, I think it is fine now.

I put in some more nit-picking comments; lmk if you want to address them, or if you want it merged as-is.

chart/config/releases.yaml
jbuns: confirmed

chart/templates/_helpers.tpl
jbuns: confirmed

chart/templates/_helpers.tpl
jbuns: confirmed and indented only my section

chart/templates/patch_serviceaccount.yaml
jbuns: confirmed

mixins/eirini/templates/metrics-service-account.yml
jbuns: confirmed

chart/templates/patch_serviceaccount.yaml
jbuns: If there's already an imagepullsecret within a serviceaccount, a patch just adds the rest in.

Example:
```
$ kubectl describe sa -n cfmr default
Name:                default
Namespace:           cfmr
Labels:              <none>
Annotations:         <none>
Image pull secrets:  default-dockercfg-sww5b # <~ Default imagepullsecret openshift adds for its private image registry
Mountable secrets:   default-token-fd55q
                     default-dockercfg-sww5b
Tokens:              default-token-fd55q
                     default-token-lrmfs
Events:              <none>

$ kubectl patch serviceaccount default -p '{"imagePullSecrets": [ {"name":"foo"}, {"name":"bar"} ] }' -n cfmr
serviceaccount/default patched

$ kubectl describe sa -n cfmr default
Name:                default
Namespace:           cfmr
Labels:              <none>
Annotations:         <none>
Image pull secrets:  foo (not found)
                     bar (not found)
                     default-dockercfg-sww5b
Mountable secrets:   default-token-fd55q
                     default-dockercfg-sww5b
Tokens:              default-token-fd55q
                     default-token-lrmfs
Events:              <none>
```

jbuns: confirmed and comments added

jbuns:
> Thanks for making the changes, I think it is fine now.
> 
> I put in some more nit-picking comments; lmk if you want to address them, or if you want it merged as-is.

Just went ahead and fixed them.

jandubois (approved):

Issue #1422, opened 2020-10-01
Create a tool for merging YAML files *with* comments


jandubois:
**Is your feature request related to a problem? Please describe.**

We want to be able to have separate `values.yaml` files for each component in the `mixins/` directory. Unlike the `config/*.yaml` files, the `values.yaml` files cannot be mixed in a single target directory, where they all use separate names; we need a single final `values.yaml` file for the helm chart. Obviously comments in both the main `values.yaml` file and in the ones to be merged in should be retained.

Another use case is updating YAML files via CI, like bumping the version of the Eirini image in `mixins/eirini/config/eirini.yaml` #841 

**Describe the solution you'd like**

I would like to see a tool in [kubecf-tools](https://github.com/cloudfoundry-incubator/kubecf-tools) that merges multiple YAML files, e.g.

```
yaml-merge chart/values.yaml mixins/bits/values.yaml mixins/eirini/values.yaml > output/helm/values.yaml
```

There may be existing code in [create_sample_value.rb](https://github.com/cloudfoundry-incubator/kubecf/blob/master/scripts/create_sample_values.rb) that can be reused for this project.

To have predictable merging semantics I think it is sensible to expect all input YAML files to be well-formatted, and also have all keys in sorted sequence. Additional constraints may become obvious once implementation starts.

Each comment block belongs to the key **below** it.

Instead of writing a spec I'll just try to give an example with `file1.yaml`:

```
# Some comment
aaa:
  a1:
  # here comes a3
  a3:
# Nothing to see
ccc: foo
```

and `file2.yaml`:

```
# Another comment
aaa:
  # this is a2
  a2:
bbb: "Better Business Bureau"
```

Merging:

```
$ merge-yaml file1.yaml file2.yaml
# Some comment
# Another comment
aaa:
  a1:
  # this is a2
  a2:
  # here comes a3
  a3:
bbb: "Better Business Bureau"
# Nothing to see
ccc: foo
```

Running `merge-yaml` on a single file will effectively `lint` the file (indentation is proper, keys are sorted). If it doesn't throw an error, it will just copy the input file to STDOUT.

Optional/special rules:

* A comment block at the top of the file, until the first empty line, is a file level comment and does not belong to any key (could be implemented with a special empty key that will sort to the top, but the key is suppressed on output.

* Comments starting with `##` (or some other sequence) are internal and will be omitted from the output file.

**Describe alternatives you've considered**

[...]

Issue #1232, opened 2020-08-19
Move kubecf-specific scripts from catapult


viovanov:
**Is your feature request related to a problem? Please describe.**
We now have multiple ways of deploying and configuring kubecf, and they are split between the `kubecf` and `suse/catapult` repos.

**Describe the solution you'd like**
Build and run code that exists in catapult should live in `kubecf-tools` and `kubecf`.
Scripts that can be reused and are useful for other projects should go in `kubecf-tools`, the rest should go in `kubecf`.

**Additional context**
n/a

viccuad:
Blocked by RFD-9 https://github.com/cloudfoundry-incubator/kubecf/pull/1495/

Issue #1231, opened 2020-08-19
Create example values.yaml for all the currently tested use cases


viovanov:
**Is your feature request related to a problem? Please describe.**
We're currently generating values.yaml files when we test.
This makes it harder to know what's being tested, and to exemplify which configurations are tested.

**Describe the solution you'd like**
The various configurations of kubecf should be represented by values.yaml files that are used in each test scenario.
The scenarios that we must support:

- single availability deployment with Diego, no autoscaler, internal database, no ingress
- single availability deployment with Eirini, no autoscaler, internal database, no ingress
- for the upgrade scenario - single availability deployment with Diego, no autoscaler, internal database, no ingress

**Additional context**
n/a

Issue #1018, opened 2020-06-15
Dependency definition and download


viovanov:
## Acceptance Criteria

- Ability to:
  - define a download location
  - define a version
  - verify the signature of the downloaded binary (or archive)
  - the downloads can be installed (into PATH)
  - check versions for local binaries (that we can't download)
- should be implemented in ruby
- should have unit tests
- support for Linux and OSX
- input file similar to (should be machine editable): https://github.com/cloudfoundry-incubator/kubecf/blame/master/dependencies.yaml

PR #1495, opened 2020-10-19
feat: RFD 9: Spec for deployment and automation scripts

viccuad:
## Description

The following spec is an interface specification for sharing deployment and
automation scripts of Helm charts and utility applications.

## Motivation and Context

Possible implementation for https://github.com/cloudfoundry-incubator/kubecf/issues/1012

May aid in:
https://github.com/cloudfoundry-incubator/kubecf/issues/220
https://github.com/cloudfoundry-incubator/kubecf/issues/75
https://github.com/cloudfoundry-incubator/kubecf/issues/1231
https://github.com/cloudfoundry-incubator/kubecf/issues/1232

Provides a framework (but leaves out for now):
https://github.com/cloudfoundry-incubator/kubecf/issues/1018

Relates to:
https://github.com/cloudfoundry-incubator/kubecf/issues/1422

## Types of changes

## Checklist:

### DIFF
--- a/doc/rfd/rfd/0009/README.md
+++ b/doc/rfd/rfd/0009/README.md
@@ -79,7 +79,6 @@ Those targets and files must be in the following structure of folders and execut
 ├── clobber/*
 ├── chart/*
 ├── configure/*
-│   └── patches/*
 ├── install/*
 ├── wait/*
 ├── include/*
@@ -114,22 +113,24 @@ defined behaviour:
   already decompressed in `$pwd/$our-chart/chart/`. `$our-chart` being the name
   of the chart at hand (e.g: kubecf).
 
-- `configure`: Accepts a list of alphanumerically ordered subsets of
-  `values.yaml`, inside the folder `configure/patches/`, which together with the
-  examples provided in `example-values/`, get merged to output the final
-  values.yaml.
-  Outcome is a valid `$pwd/$our-chart/values.yaml` file for the chart at hand.
+- `configure`: Generates any config values that cannot be taken from the
+  `example-values/` folder as they aren't there. Normally, this will be
+  cluster-dependent configs. E.g: values for the system_domain, ingress certs,
+  loadBalancer annotations for externalDNS, etc.
+  Outcome is a valid set of yaml files in `$pwd/$our-chart/values/` folder.
   This allows for reuse of existing example values, and for downstream tools to
   inject their own yaml subsets as needed.
 
-- `install`: Takes a `$pwd/$our-chart/values.yaml` and the chart contained in
-  `$pwd/$our-chart/chart/` and installs the Helm Chart at hand. 
+- `install`: Takes the config values in the `$pwd/$our-chart/values/` folder and
+  the chart contained in `$pwd/$our-chart/chart/` and installs the Helm Chart at
+  hand.
 
 - `wait`: Waits until deployment of the chart is up and ready.
 
-- `upgrade`: Takes a `$pwd/values.yaml` and a chart contained in
-  `$pwd/$our-chart/chart/` and upgrades the Helm Chart at hand. The upgrade must
-  work for both same version and different versions.
+- `upgrade`: Takes the config values in `$pwd/$our-chart/values/` folder and a
+  chart contained in `$pwd/$our-chart/chart/` and upgrades the Helm Chart at
+  hand. The implementation must work for both same-version upgrades and
+  different-version upgrades.
 
 - `clean`: Performs a helm uninstall.
 
@@ -150,11 +151,11 @@ Each of these targets:
   `configure` several times with the same outputs gives you the same
   `values.yaml` as result. Calling `install` with the same `values.yaml` gives
   you the same deployment (that means that `install` must clean on its own).
-- Must be isolated: implementation of `install` does not call on `configure`
+- Must be isolated: E.g; implementation of `install` does not call on `configure`
   itself.
   They may, though, call shared implementation code in the `include` folder of
   their shared root path.
-- Chart values must be only consumed from a yaml file, and not using `--set`.
+- Chart values must be only consumed from yaml files, and not using `--set`.
 - When useful, should save resulting artifacts of execution into
   `$pwd/artifacts`. E.g: `$pwd/artifacts/cats-run.log`
 
@@ -164,13 +165,13 @@ besides the expected default behaviour defined here.
 
 Apart from the targets, the following folders and files must be provided:
 
-- `example-values/`: Divided into 2 folders:
-  * `example-values/minimal/`: examples of minimal values.yaml that pass a
-    schema check. One can always be sure to have a working deployment with one
-    of them. E.g: `diego-sa-config.yaml`, `eirini-sa-config.yaml`.
-  * `example-values/subset/`: examples of subsets of values.yaml, which alone
-    don't pass a schema check. E.g: `private-registry.yaml`,
-    `tests-config.yaml`, `ha.yaml`, `ingress.yaml`.
+- `example-values/`: Contains examples of possible config values. Separated into
+  2 folders:
+  * `example-values/overlapped/`: examples which cannot be used at the same
+  time. E.g: `diego.yaml`, `eirini.yaml`.
+  * `example-values/subset/`: examples of subsets of configuration, which can be
+    used together. E.g: `private-registry.yaml`, `testsuites-config.yaml`, `ha.yaml`,
+    `ingress.yaml`.
 
 - `imagelist.txt`: Lists all needed container images, fully qualified with
     repository, org, name and tag. Useful for airgap and private registry
@@ -196,10 +197,8 @@ Targets' file structure:
     │   └── 10_chart.sh
     ├── configure/
     │   ├── patches/
-    │   │   ├── 01_kubecf_develop.yaml
-    │   │   ├── 02_kubecf_loadbalancer_svc.yaml
-    │   │   ├── 99_kubecf_config_override.yaml.disabled (made non exec)
-    │   └── 10_merge.sh
+    │   │   └── 02_kubecf_loadbalancer_svc.yaml
+    │   └── 10_create_cluster-dependent_values.sh
     ├── install/
     │   ├── 10_install_quarks.sh
     │   ├── 20_install.kubecf.sh
@@ -224,13 +223,14 @@ Targets' file structure:
     │   │   └── 01_cats.sh
     │   └── readme.md
     └── example-values/
-        ├── minimal/
-        │   ├── diego-sa.yaml
-        │   └── eirini-sa.yaml
+        ├── overlapped/
+        │   ├── diego.yaml
+        │   └── eirini.yaml
         └── subset/
             ├── ha.yaml
             ├── autoscaler.yaml
             ├── ingress.yaml
+            ├── testsuites-config.yaml
             └── private-registry.yaml
 ```
 
--- a/doc/rfd/rfd/0009/README.md
+++ b/doc/rfd/rfd/0009/README.md
@@ -163,7 +163,7 @@ It is not recommended for targets to take arguments and options on execution
 besides the expected default behaviour defined here.
 
 
-Apart from the targets, the following folders and files may be provided:
+Apart from the targets, the following folders and files must be provided:
 
 - `example-values/`: Divided into 2 folders:
   * `example-values/minimal/`: examples of minimal values.yaml that pass a
--- a/doc/rfd/rfd/0009/README.md
+++ b/doc/rfd/rfd/0009/README.md
@@ -42,7 +42,9 @@ together with their expected use:
 ├── one-offs/*
 │
 ├── include/*
-└── example-values/*
+├── example-values/*
+│
+└── spec_ver
 ```
 
 This structure as a whole is encouraged to be published together with the Helm
@@ -113,6 +115,9 @@ Apart from the targets, the following folders may be provided:
   `example-values/eirini-autoscaler-config.yaml`.
 
 - `include/`: Contains possible reusable scripts to be used by 1 or more targets.
+
+- `spec_ver`: file containing a string with the version of the spec that is being
+  satisfied.
 
 
 #### Example 1: KubeCF
--- a/doc/rfd/rfd/0009/README.md
+++ b/doc/rfd/rfd/0009/README.md
@@ -1,0 +1,152 @@
+---
+authors: Victor Cuadrado <vcuadradojuan@suse.com>
+state: draft
+discussion: N/A
+---
+
+# RFD 9 Specification for deployment and automation scripts
+
+## Context
+
+
+Dealing with a handful of loosely dependent Helm charts and building a Helm
+distribution from them without creating an all-or-nothing "super-chart" is
+difficult. One needs to make sure that deployment and automation scripts are
+interoperable between charts, repos, repos' histories, and that they can be
+reused and extended downstream as needed.
+
+At the same time one needs to create, reuse, and extend test environments across
+charts, and between upstream and downstream.
+
+The following spec is an interface specification for sharing deployment and
+automation scripts of Helm charts and utility applications.
+
+
+## Specification
+
+### Version `0.1`.
+
+
+The interface consists in a specific structure of files, folders and
+executables, together with their expected use.
+
+Each target is expected to be either:
+- An executable file, starting with a recognized target name (`upgrade*`. E.g:
+  `upgrade.sh`, or `upgrade.rb`).
+- A folder (`install/`) containing executable files starting with digits
+  (`10-do-foo`, `20-do-bar`).
+  Upon target call, all the files starting with digits will get executed in
+  alphanumeric order.
+  It is allowed for the folder to contain other files/folders besides those
+  starting with digits. Those would not be automatically executed upon target
+  call (e.g: readme files, local include folders, etc).
+  
+Deployment and automation scripts will expose the following targets:
+
+- `chart*`: Obtains the latest public release of the chart at hand, and puts it
+  already decompressed in `$pwd/$our-chart/chart/`. `$our-chart` being the name
+  of the chart at hand (e.g: kubecf).
+
+- `configure*`: Outcome is a valid `$pwd/$our-chart/values.yaml` file for the
+  chart at hand. This target accepts a list of alphanumerically ordered subsets
+  of `values.yaml`, inside the folder `configure/patches/`, which get merged to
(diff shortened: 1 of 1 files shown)


viccuad:
For future proofing, I think it would be useful to add a file `spec_ver`, containing a string with the version that the scripts implement (e.g: `0.1`). This allows us to make changes to the spec and have several charts be compliant with different spec versions in the future.

manno:
I'm missing context on this. Can you please add some explanation what this is for (like a user story)? 

I'm asking myself for example:

* which helm charts should implement this (kubecf?)
* who is the intended user of these scripts (ci, devs, users)? 
* where should these ten folders go? (root dir of git repo, hack, bin, ci?)
* how would someone use these folders?
* I'm also missing explanations for each of the folders. (clean what, wait for what?)

> The following spec is an interface specification for sharing deployment and automation scripts of Helm charts and utility applications.

This approach looks a bit too special to me. I'm concerned nobody else is doing it like this. 
It would replace "simply" calling helm, with executing a hierarchy of scripts?
Did you look into alternatives, maybe like [helmfile](https://github.com/roboll/helmfile)?

> Dealing with a handful of loosely dependent Helm charts and building a Helm
distribution from them without creating an all-or-nothing "super-chart" is
difficult. 

This makes me think it's about the different helm charts included in kubecf.
So how would kubecf install instructions look like with this "structure of folders and executables"?
Is this intended for releases or builds from git?
Do these scripts modify configuration (configure folder?), how is configuration distributed?
Since upstream projects won't implement this structure, it still has to be maintained, I think?

> One needs to make sure that deployment and automation scripts are
interoperable between charts, repos, repos' histories, and that they can be
reused and extended downstream as needed.

Does that mean the scripts should be used and modified by KubeCF and CAP?

> At the same time one needs to create, reuse, and extend test environments across
charts, and between upstream and downstream. The following spec is an interface specification for sharing deployment and
automation scripts of Helm charts and utility applications.

Does this also solve the problem of upstream and downstream using different clusters and testing different things? Would there be a "structure of folders and executables" for each of these use cases?

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,176 @@
...
+│
+└── spec_ver
+```
+
+This structure as a whole is encouraged to be published together with the Helm
+chart. This can be for example as artifacts of a chart release, or present in
+the chart repository, but we recommend that this files are bundled inside of the
+chart: this simplifies consumption downstream.
+
+While targets are allowed to take arguments and options on execution besides the
```
viovanov, line 55: Can you please define `target` before this paragraph?

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,176 @@
...
+starting with digits. Those would not be automatically executed upon target
+call (e.g: readme files, local include folders, etc).
+  
+Deployment and automation scripts will expose the following targets:
+
+- `chart`: Obtains the latest public release of the chart at hand, and puts it
+  already decompressed in `$pwd/$our-chart/chart/`. `$our-chart` being the name
+  of the chart at hand (e.g: kubecf).
+
+- `configure`: Outcome is a valid `$pwd/$our-chart/values.yaml` file for the
```
viovanov, line 73: Why mangle yaml files? The `helm` CLI supports specifying multiple values yaml files.

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,176 @@
...
+- `install`: Takes a `$pwd/$our-chart/values.yaml` and the chart contained in
+  `$pwd/$our-chart/chart/` and installs the Helm Chart at hand. 
+
+- `wait`: Will wait until deployment of the chart is up and running.
+
+- `upgrade`: Takes a `$pwd/values.yaml` and a chart contained in
+  `$pwd/$our-chart/chart/` and upgrades the Helm Chart at hand. The upgrade must
+  work for both same version and different versions.
+
+- `clean`: Removes all possible objects created by any target from the cluster.
```
viovanov, line 88: Can you split this into `clean` and `delete`?
`clean` can remove objects from the cluster, and `delete` can delete the chart.
I usually want to remove things from the cluster quite often, but rarely want to delete the chart from my local disk.

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,176 @@
...
+   Deletes `$pwd/$chart-name` at the end.
+
+- `one-offs/*`: Helper targets. Follow the same rules as the rest of targets,
+  except they may not be idempotent. E.g: `klog`, `login`, `smokes`, `cats`.
+  
+Each of these targets:
```
viovanov, line 94: This section should be higher in the doc.

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,176 @@
...
+  `values.yaml` as result. Calling `install` with the same `values.yaml` gives
+  you the same deployment (that means that `install` must clean on its own).
+- Is isolated: implementation of `install` does not call on `configure` itself.
+  They are allowed though to call shared implementation code in the `include`
+  folder of their shared root path.
+- When useful, saves resulting artifacts of execution into `$pwd/artifacts`.
+  E.g: `$pwd/artifacts/cats-run.log`
+
+
+Apart from the targets, the following folders may be provided:
```
viovanov, line 111: I think these _must_ be provided. They must be used to compose the helm install/upgrade commands.

viovanov (changes requested):
I don't think my review is finished, but would like to start discussing.
@viccuad the main concern I have is about the values yaml files, so please consider that first. 
And thanks! this is great

viccuad:
Thanks Mario for the review!

> I'm missing context on this. Can you please add some explanation what this is for (like a user story)?

Done! please tell me if they aren't clear.

> 
> I'm asking myself for example:
> 
>     * which helm charts should implement this (kubecf?)

Any that want to. I would start with kubecf.
Ideally, any that need to be tested together: kubecf, quarks, minibroker, stratos, stratos-metrics, caps-registry, ceph. The spec is too specific and too loose at the same time, but it's a starting point. It could completely change in follow-up versions. And at some point become irrelevant because Helm fixe, implements, or specifies the needed parts in the future.

>     * who is the intended user of these scripts (ci, devs, users)?

ci & devs. I think we are doing a disservice to users by not providing them with a Helmfile or example-values, though.

> 
>     * where should these ten folders go? (root dir of git repo, hack, bin, ci?)

It is not specified, one is free to put them wherever.

> 
>     * how would someone use these folders?

There would be a test-runner that would execute the targets. That test-runner would be simpler that current catapult.
Also, the folders can be consumed downstream by injecting the needed values.yaml subsets, or by injecting an odd `upgrade/99_apply-workaround.sh` when needed, that would get executed in the specified alphanumeric order.

> 
>     * I'm also missing explanations for each of the folders. (clean what, wait for what?)

Reworded.

> This approach looks a bit too special to me. I'm concerned nobody else is doing it like this.
> It would replace "simply" calling helm, with executing a hierarchy of scripts?
> Did you look into alternatives, maybe like [helmfile](https://github.com/roboll/helmfile)?

I have looked, but I haven't for example written a Helmfile myself for kubecf.

This RFD overlaps with Helmfiles in that it provides a declarative list of possible values for the chart, to pick and mix, which incidentally, form our test cases.
But the RFD intent is to also share implementations of CI/automation, shipping them as part of the release, to be consumed downstream. We could move to Helmfiles for the declarative values, and keep the rest of conventions for sharing the automation bits.

> This makes me think it's about the different helm charts included in kubecf.
> So how would kubecf install instructions look like with this "structure of folders and executables"?

For users, nothing changes.
For devs/CI/automation, we just change `kubecf/Makefile` to call the correct target, and this targets could be implemented with changes to the `kubecf/scripts/` folder. 

> Is this intended for releases or builds from git?

It is intended for both. Releases (just some files inside of the chart that nobody cares about) and daily git builds, which are now being consumed downstream for nightlies.

> Do these scripts modify configuration (configure folder?), how is configuration distributed?

Added explanation of configure target. Configuration is distributed and created by merging the `example-values/` files with `configure/patches`. I think we should be clearer here, or use Helmfiles.

> Since upstream projects won't implement this structure, it still has to be maintained, I think?

That could be the case. That's the status quo for Catapult, which implements more or less this for kubecf, stratos, stratos-metrics, cf-operator, and all their one-offs. Catapult would benefit from a rewrite and no implementing this spec.


> Does that mean the scripts should be used and modified by KubeCF and CAP?

Correct.

> Does this also solve the problem of upstream and downstream using different clusters and testing different things? Would there be a "structure of folders and executables" for each of these use cases?

Correct. The idea is to implement this spec on Catapult v2, substituting "helm chart" with "test environment". As a minimum, Catapult v2 would create test environments:
   - env of kubectl, cf, helm
   - plus a cluster deployment
   - plus needed beforeTest configuration of that cluster and env (DNS, etc)
   - plus which specific charts are being tested, which targets run, etc)
This configuration would be declarative, in a folder of example values yaml, which would create a suite of test cases for CAP or others.

doc/rfd/rfd/0009/README.md
viccuad: done! thanks

doc/rfd/rfd/0009/README.md
viccuad: That's correct, and a better approach. I think a good approach would be for `configure` target to provide a `$pwd/values/` folder with those, that the `install` and `upgrade` targets would consume then.

doc/rfd/rfd/0009/README.md
viccuad: just when you were reviewing, I tackled Mario's review and reworded `clean` and added a `clobber` target that leaves the cluster pristine. Would that be ok?

doc/rfd/rfd/0009/README.md
viccuad: do you mean the one-offs or the "each of these targets" paragraph?

doc/rfd/rfd/0009/README.md
viccuad: Correct! fixed. Also, I added imagelist.txt.

viccuad:
I have two concerns with the current draft:
- Namespace of the helm chart is not defined. We should define it either implicitly (`ns == name-of-chart`, eg: `kubecf`) or explicitly on a file. Which sounds similar to Helmfiles.
- Having all charts implement the spec means that development may become slower, if we need to bump spec version implementations across all of them.

doc/rfd/rfd/0009/README.md
viccuad: I'm still thinking on how to tackle this properly. Merging subsets of yaml files has the benefit that one can order then alphanumerically, and provide a deterministic way to do the merge.
I wonder about yaml files whose objects overlap. Do any charts have any of those? What happens when merging them? Does it matter?

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,256 @@
...
+rely on their example-values and automation.
+
+
+#### Parallel upstream developer
+
+I am a developer of kubecf chart. My chart doesn't directly depend on stratos
+chart or caps-registry chart. But sometimes they develop automation that I could reuse.
+If they implement this spec, I will have access to their automation in simple,
+pull-way manner.
+
```
prabalsharma, line 52: Maybe we should also add CAP / CAP CI user story, here.

viccuad: The "downstream" developer user story is written with that intent. Would that be fitting?

doc/rfd/rfd/0009/README.md
viccuad: Moved to a folder for all config values in https://github.com/cloudfoundry-incubator/kubecf/pull/1495/commits/3e5206b7c0cadbb2d136701b481830e789b34fe8

doc/rfd/rfd/0009/README.md
viccuad: done in https://github.com/cloudfoundry-incubator/kubecf/pull/1495/commits/6b63f4ebbc6dbe7b849434f583953ef7d7c958cb

doc/rfd/rfd/0009/README.md
viccuad: renamed `clean` to `uninstall` in https://github.com/cloudfoundry-incubator/kubecf/pull/1495/commits/76c684f28d9a79c6483c15fe125b7380658fd6e9, and we have clobber nevertheless.

mook-as:
How would the scripts (e.g. in `configure`) work? This defines no method of passing in options (e.g. to set the system domain, which presumably the chart itself can't know ahead of time, otherwise we'd just hard code it into the chart).

viccuad:
> How would the scripts (e.g. in `configure`) work? This defines no method of passing in options (e.g. to set the system domain, which presumably the chart itself can't know ahead of time, otherwise we'd just hard code it into the chart).

Currently in Catapult, and since cf-ci times, we use a configmap `cap-values` with information of the cluster external to Kube level.

Added specification for it in https://github.com/cloudfoundry-incubator/kubecf/pull/1495/commits/30fc89593119ab6ac7395af39bdca4d193d70c79.

mook-as:
> Added specification for it in [30fc895](https://github.com/cloudfoundry-incubator/kubecf/commit/30fc89593119ab6ac7395af39bdca4d193d70c79).

Does this mean the configuration is only available in the `configure` step? Also, how will it figure out _which_ config map to use (since we didn't name it in the spec)?  Should we have a spec-version in the config map too?

viccuad:
> Does this mean the configuration is only available in the `configure` step?

Correct, that's the idea, as there shouldn't be any need to read configurations on the other steps.

> Also, how will it figure out _which_ config map to use (since we didn't name it in the spec)?

Doh, I named it in the comment but not on the spec. Fixed in https://github.com/cloudfoundry-incubator/kubecf/pull/1495/commits/44c7fcdb6286bf139419f8a10ab2b5dd612a8e1d.

> Should we have a spec-version in the config map too?
Maybe we should have one, but since cluster creation is not part of the spec, and we haven't had the need to change the configmap keys in a long time, maybe it can wait for future spec bumps.

jandubois:
I've read this document several times, but it is still not clear to me how this would be used, and what benefits it will provide over what we already have.

I think it would be useful to know the requirements, the *what* you are trying to achieve; it is kind of hard to give feedback on the *how* without understand the *why* behind the decisions.

It sounds like the purpose is to simplify CAP QA, but the title is very generic "deployment and automation scripts", so maybe the scope is a lot narrower than I initially presumed. Is there more than a single consumer of the implementation of this spec?

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,266 @@
...
+- When useful, should save resulting artifacts of execution into
+  `$pwd/artifacts`. E.g: `$pwd/artifacts/cats-run.log`
+
+It is not recommended for targets to take arguments and options on execution
+besides the expected default behaviour defined here.
+  
+Deployment and automation scripts must expose the following targets, with the
+defined behaviour:
+
+- `chart`: Obtains the latest public release of the chart at hand, and puts it
```
mook-as, line 133: We no longer need to call this `chart`; it can just be `download` at this point, since nothing else depends on this being a chart.

doc/rfd/rfd/0009/README.md
```diff
@@ -0,0 +1,266 @@
...
+  already decompressed in `$pwd/$our-chart/chart/`. `$our-chart` being the name
+  of the chart at hand (e.g: kubecf). While this target is required to be
+  implemented, the intent is that test runners may decide to not call it and
+  drop themselves the chart under test in `$pwd/$our-chart/chart/`.
+
+- `configure`: Generates any config values that cannot be taken from the
```
mook-as, line 139: We should also pass in what configuration is under test. (As in, the things under `example-values/`.)

viccuad:
> I think it would be useful to know the requirements, the _what_ you are trying to achieve; it is kind of hard to give feedback on the _how_ without understand the _why_ behind the decisions.
> 
> It sounds like the purpose is to simplify CAP QA, but the title is very generic "deployment and automation scripts", so maybe the scope is a lot narrower than I initially presumed. Is there more than a single consumer of the implementation of this spec?

You are not the first, so I think I failed to make my point in the "context" section of the spec.

The purpose is to simplify and reuse any CI that we do for any chart. Those that:
- Validate a new KubeCF release, against pre-existing charts (quarks, minibroker, stratos)
- Validate a new CAP release (formed from a set of charts) against pre-existing charts (previous CAP release)
- Validate a new Stratos release, against pre-existing charts (KubeCF, minibroker)
- Etc.

The idea is to:
- Get a declarative state of the system under test. This includes the values.yaml being used, and the cluster.
- Package CI scripts in a way that they can be consumed and modified downstream.
- Provide a versioned spec to iterate on the implementation.

viccuad:
We held a meeting to talk about the spec, and we extracted some points.

Concerns (in any order):
- Over engineered. Targets are too monolithic, one cannot inject operations in the middle by using the alphanumeric order anyways as one doesn't know which number to use, nor it may be needed.
- Set of values.yaml under test are not defined, and they may overlap when merging. The test runner has the responsibility of selecting the values under test, which lies outside of the spec, and renders the spec moot.
- May not be relevant for future implementations.
- Chart-specific implementation.
- Other consumers (E.g: Harbor) may not implement the spec. In that case, where does the script bundle live and how do we consume it?
- We don't have granularity on the testsuites to run. If we were to specify each testsuite, then the targets would be product specific.

Ideas:
- Spec object is shared between CIs, instead of sharing charts. The CIs would then use the spec object (from now on the "script bundle") to fetch the corect charts under test.
- Change the `chart` target to obtain the chart under test from either local path, url, or end-user repository.
- Simplify targets away from folders and alphanumeric execultables and into 4-5 scripts: clean-clobber, chart, install, upgrade.
- Don't have example-values be subsets of config.yamls to be merged, provide only sets of configs that can be fully tested.

Outcome:

From the ideas, it seemed that we need to define the usage and interface between spec objects and CI/users as a starting point.
A spike was created for it: https://github.com/cloudfoundry-incubator/kubecf/issues/1578.

A rewrite of the spec, simplifying it with the ideas and concerns, is needed.

viccuad:
Thulio shared the other day porter.sh; adopting it may very well satisfy this RFD.

jandubois:
@viccuad I've marked this PR as "stale". Please consider closing it for the time being (but keep the branch around for now), as I don't think any work around this will be scheduled in the near future.

viccuad:
Closing for now. From the planning, it would need a rewrite anyways.

PR #1499, opened 2020-10-20
fix: Inline helm_ls and helm_delete

jandubois:
These are catapult functions that are not defined in kubecf. They are wrappers that work with both helm 2 and helm 3, but the kubecf scripts all require helm 3, so this isn't necessary.

Also don't hardcode the operator namespace, but use the proper config variable.

### DIFF
--- a/scripts/cf-operator-delete.sh
+++ b/scripts/cf-operator-delete.sh
@@ -3,8 +3,8 @@ source scripts/include/setup.sh
 
 require_tools kubectl helm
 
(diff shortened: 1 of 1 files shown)


mook-as (approved):
I see that [cf-operator-apply.sh](https://github.com/cloudfoundry-incubator/kubecf/blob/e656d95a7be5dd686adb5cd68d4f920e0b241166/scripts/cf-operator-apply.sh#L10) hard-codes the release name to `cf-operator`, so there's no point in trying to use a variable there.

LGTM.

PR #1502, opened 2020-10-21
feat: Improvements to GitHub Actions workflow

### LINKED ISSUE #1144: PR Checking with GitHub Actions

mook-as:
## Description
~This pull request will change our GitHub Actions workflow to include at least the basic tests, to the point where we might start deprecating our Concourse pipelines.~

~For now, though, I just have a design of how we might handle the build triggering and cluster allocation; I'd like some feedback before actually implementing things.   Hence, this is a draft with just markdown changes.  I'll edit the description once we agree and I've implemented it reasonably.~

I am now rescoping this to _just_ the changes described to scheduling the runs; this is already enough of a change that it's a good chunk to review.  Also, I can work on the tests independently of this PR (in my fork, manually triggering runs).

## Motivation and Context
Part of #1144 (but will no longer fix it by itself).

## How Has This Been Tested?

I have been running this in my fork; [sample run](https://github.com/mook-as/kubecf/actions/runs/338099108).

## Types of changes

## Checklist:

This has security implications in the sense of this may trigger runs on CI of potentially untrusted code.  So it's security of our infrastructure (and from there our code), not of the product (unless they manage to change our code).

### DIFF
--- a/.github/workflows/pull-request-lint.yaml
+++ b/.github/workflows/pull-request-lint.yaml
@@ -1,0 +1,38 @@
+name: Pull Request Linting
+on:
+  # Trigger the workflow when PRs pushed; we still run the workflow files from
+  # the master branch, to ensure that malicious actors cannot exfiltrate any
+  # secrets.
+  pull_request_target:
+    branches: [master]
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-20.04
+    env:
+      PINNED_TOOLS: true
+      TOOLS_DIR: ${{ github.workspace }}/tools
+    defaults:
+      run:
+        working-directory: kubecf
+
+    steps:
+      # Python setup
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      # Checkout kubecf
+      - name: Checkout KubeCF
+        uses: actions/checkout@v2
+        with:
+          path: kubecf
+          fetch-depth: 0
+          submodules: recursive
+
+      # Install tools
+      - run: make tools-install
+
+      # Check lint
+      - run: make lint
--- a/.github/workflows/pull-request-lint.yaml
+++ b/.github/workflows/pull-request-lint.yaml
@@ -30,6 +30,7 @@ jobs:
           path: kubecf
           fetch-depth: 0
           submodules: recursive
+          persist-credentials: false
 
       # Install tools
       - run: make tools-install
--- a/.github/workflows/pull-request-lint.yaml
+++ b/.github/workflows/pull-request-lint.yaml
@@ -2,7 +2,9 @@ name: Pull Request Linting
 on:
   # Trigger the workflow when PRs pushed; we still run the workflow files from
   # the master branch, to ensure that malicious actors cannot exfiltrate any
-  # secrets.
+  # secrets.  It is important that we don't expose any secrets to the checked-
+  # out code, both in the environment as well as on disk (i.e. we must run
+  # actions/checkout with persist-credentials: false).
   pull_request_target:
     branches: [master]
 
--- a/.github/workflows/pull-request-schedule.yaml
+++ b/.github/workflows/pull-request-schedule.yaml
@@ -1,0 +1,103 @@
+name: Pull Request Scheduler
+on:
+  schedule:
+  - cron: '*/15 * * * *'
+  workflow_dispatch: {}
+
+env:
+  BACKEND: gke
+  CLUSTERS_PER_PR: "2"
+  MAX_CLUSTERS: "5"
+  OWNER: ${{ github.repository_owner }}
+  LABEL_QUEUE: pr-test-queue
+  LABEL_TRIGGER: pr-test-trigger
+
+jobs:
+  determine-capacity:
+    name: Determine Capacity
+    runs-on: ubuntu-latest
+    outputs:
+      # The number of pull requests we can start testing now.
+      capacity: ${{ steps.calculate-capacity.outputs.capacity }}
+    env:
+      GKE_PROJECT: ${{ secrets.GKE_PROJECT }}
+    steps:
+    - name: Checkout KubeCF
+      # This is only needed to find the GitHub script in the find-jobs step.
+      uses: actions/checkout@v2
+    - name: Checkout catapult
+      uses: actions/checkout@v2
+      with:
+        repository: SUSE/catapult
+        path: catapult
+
+    # set GKE secret. TBD: use vault instead
+    - name: Setup Cloud Provider Credentials
+      run: |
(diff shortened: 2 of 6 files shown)


mook-as:
FWIW, I would probably also move the linting to its own file, since that doesn't need to allocate a cluster; this would mean we get quicker feedback on linting errors, but still run the full tests once we have a cluster.  That will need to also be after PR approval, though.

The `pull-request-schedule.yaml` step will probably need to written as a script rather than actions YAML, since it will need to be able to make GitHub API calls — as far as I can tell, sorting issue by last-label-apply-time doesn't really have a good API.

jandubois:
> That will need to also be after PR approval, though.

Why can't we run just linting before PR approval? At least then we won't start reviewing until all linting errors are fixed. 😄

mook-as:
> > That will need to also be after PR approval, though.
> 
> Why can't we run just linting before PR approval? At least then we won't start reviewing until all linting errors are fixed. 😄

Because the linting script would come from the PR, and it could touch `GITHUB_TOKEN`, like everything else, right? Hmm. At least, I assumed the token would be available; now that I think about it though, it might be fine (because the workflow definition is still from `master`, and I don't see us exposing that for any reason…)

viccuad:
The build triggering and cluster automation sounds perfect, gives us the blocking pool that we never implemented in Concourse.

.github/workflows/pull-request-schedule.yaml
```diff
@@ -0,0 +1,103 @@
...
+      with:
+        repository: SUSE/catapult
+        path: catapult
+
+    # set GKE secret. TBD: use vault instead
+    - name: Setup Cloud Provider Credentials
+      run: |
+        json_file="$(mktemp)"
+        cat > "${json_file}" <<< "${GKE_CRED_JSON}"
+        echo "GKE_CRED_JSON=${json_file}" >> "${GITHUB_ENV}"
```
viccuad, line 39: Does this respect newline characters correctly?

.github/workflows/pull-request-lint.yaml
```diff
@@ -0,0 +1,39 @@
+name: Pull Request Linting
+on:
+  # Trigger the workflow when PRs pushed; we still run the workflow files from
+  # the master branch, to ensure that malicious actors cannot exfiltrate any
+  # secrets.
```
viccuad, line 5: One could still attack some vulnerability of the linters themselves, to run arbitrary code on PR push. But I suppose that's a small enough attack surface.

.github/workflows/pull-request-schedule/trigger-builds.js
```diff
@@ -0,0 +1,65 @@
+module.exports = async ({ github, context }) => {
+    // PullRequestQuery is a GraphQL query to find pull requests with the proper
+    // label
+    const PullRequestQuery = `
+        query($owner: String!, $repo: String!, $label: String!) {
+            repository(owner: $owner, name: $repo) {
+                labels(first: 10, query: $label) {
+                    nodes {
+                        name
+                        pullRequests(states: OPEN, last: 10) {
```
viccuad, line 10: Does this `last: 10` in the query for OPEN pullRequests means to obtain only the last 10 PRs? Or is it a pagination thing or the like?

viccuad (dismissed):
Looks good here! approving, but note that I just learnt how to use map, reduce, filter, etc on js. Cool API polling and queue implementation with the GH labels, states and age. Yet, I hope we don't grow the amount of Javascript :sweat_smile: .

Things that are outside the scope of this PR:

- In pull-request-ci.yaml, I think there may be misconfigurations because we are not setting `BACKEND` or `CLUSTER_NAME`. If they aren't set, it will go with the default  `BACKEND=kind`, which will set `CLUSTER_NAME=kind` and create a _buildkind_ . I doubt we want this because when we create the cluster we need to set `BACKEND=gke` which will give us a _buildgke_ to operate on the rest of targets. Maybe we could add them to the global `env` in pull-request-ci.yaml?

- The strategy matrix perhaps needs to be expanded and maybe reworked into 2 lanes (diego & eirini, as diego has testsuites that only apply to it) and then add an experimental flag for things that can continue on error. Without it, we may fail in brains and sits.

.github/workflows/pull-request-schedule.yaml
mook-as: There are no new line characters involved: `GKE_CRED_JSON` (as exported to `GITHUB_ENV`) is the path to the temporary file we created.  (There's a new line created by `echo` but I don't think that's not what you meant.)

.github/workflows/pull-request-schedule/trigger-builds.js
mook-as: You're right, this means we only get 10 PRs (with the label); that's obviously wrong (I forgot to change the testing values I used).  I should double check what order this is in (by testing with a different label), and see if picking the _first_ PRs would make sense; since we will eventually remove the labels (when the PRs get scheduled) that should work fine.

.github/workflows/pull-request-lint.yaml
mook-as: Yes, but the linter runs don't get to have any secrets (since we don't need to create clusters to test them, and we check out the source with `persist-credentials: false`).  The issue with the workflow file is that _it_ can have access to secrets (e.g. `github.token`), which it may choose to export to scripts.

…I think; it'll probably be a good idea to double check the environment in that case.  I'll try to do that.  Either way it sounds like I need to expand the comment to say ↑.

.github/workflows/pull-request-schedule/trigger-builds.js
mook-as: Fixed in a40333443489f8407f4b019b8e1b25669dcc0250

.github/workflows/pull-request-lint.yaml
mook-as: Added more comments in fcb37225b6f4f4e2c9dc674b038ddf60abbd3580

.github/workflows/pull-request-schedule.yaml
viccuad: doh true, it's a file. That answers my question.

viccuad (approved):

Issue #1248, opened 2020-08-21
Set GIT_PAGER=cat inside CI pipelines


jandubois:
For some reason `git` seems to believe to have a semi-functional terminal available when running under concourse.

That way it can hang indefinitely, waiting for input (example from running `httplint` and finding an error):

```
URLS should not start with 'http://' (with excepted domains listed in .http_exceptions)
WARNING: terminal is not fully functional
-  (press RETURN)
```

We don't want to affect developer use of the scripts, so adding `GIT_PAGER=cat` to CI should be the best way to avoid this problem in general, and not just for this particular script.

andreas-kupries:
Is this about the CI definitions under `.github/workflows` ? (Looks to be github actions ?)

jandubois:
> (Looks to be github actions ?)

This was about the concourse pipeline, but it could also apply to Github Actions.

But otherwise, yes, this is about `git diff` or `git grep` using the pager, waiting for the user to press a key before continuing. Under CI it will just hang until the runner times out.

andreas-kupries:
Given our work on switching to GHA I would say we should make this GHA and leave Concourse behind.

PR #1506, opened 2020-10-22
fix: Disable interactive paging for git commands.

andreas-kupries:
## Description

Added a `GIT_PAGER: cat` setting to the GHA definitions.

## Motivation and Context

See #1248 

## How Has This Been Tested?

This is untested as of right now. It is unclear how to test it at all.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/conventional-commits-lint.yml
+++ b/.github/workflows/conventional-commits-lint.yml
@@ -5,6 +5,10 @@ on:
       - opened
       - edited
       - synchronize
+
+env:
+  GIT_PAGER: cat
+
 jobs:
   main:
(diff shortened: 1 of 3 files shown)


mook-as (changes requested):
Please use the [workflow-level `env`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#env) instead, so we don't have repeat them for each step?

andreas-kupries:
Moved setting to workflow level with commit b979c75ab4c9f6eb2207372edb179f3b27f596fc

mook-as (approved):

PR #1514, opened 2020-10-26

cfcontainerizationbot:

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -103,9 +103,9 @@ binaries:
 # Additional files we need to download; see http_file() invocation in WORKSPACE.
 external_files:
   cf_operator:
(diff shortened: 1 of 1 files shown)


dependencies.yaml
```diff
@@ -103,9 +103,9 @@ binaries:
 # Additional files we need to download; see http_file() invocation in WORKSPACE.
 external_files:
   cf_operator:
-    sha256: 9d4f8219a204b7f0b90170c4289b0df80e6da350f8941f047bc6e873ba22ad4c
+    sha256:
```
jandubois, line 106: Why is there no sha256 anymore? @manno do you know?

jandubois (changes requested):

PR #1516, opened 2020-10-26
fix: Remove blobstore settings from eirini-helm ops file

jandubois:
## Description

These settings are not needed, and they will break when kubecf is configured to use an external blobstore because then the singleton-blobstore service doesn't even exist.

I suspect that more settings in this file are redundant, but want to have a minimal change to just fix Eirini with external
blobstore deployment.

## How Has This Been Tested?

I've only tested that Eirini continues to work on minikube. I have not tried to deploy with an external blobstore, so there may still be other undiscovered issues.


andreas-kupries (approved):
Looks ok.

PR #1519, opened 2020-10-26
fix: deployment-name should be label not annotation

univ0298:
## Description
The deployment-name is supposed to be a label on a secret, not an annotation. Without this change the multiple cluster feature doesn't work

## Motivation and Context
See above

## How Has This Been Tested?

Tested using cf-operator 6.1.7 this progressed the deployment past a blocking issue

## Types of changes

## Checklist:

### DIFF
--- a/chart/templates/multiple_cluster_mode.yaml
+++ b/chart/templates/multiple_cluster_mode.yaml
@@ -91,8 +91,9 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ .Values.features.multiple_cluster_mode.provider_link_service.nats.secret_name }} 
+  labels:
+    quarks.cloudfoundry.org/deployment-name: "{{ include "kubecf.deployment-name" . }}" 
   annotations:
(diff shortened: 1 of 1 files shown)


mook-as (approved):
Sorry for the delay.

I can confirm that:
- the deployment name is already set on other objects (pods generated by quarks-operator) as a label, without multi-cluster mode being turned on
- there are existing selectors (i.e. working on labels) looking at the deployment name
- after the change, all of the deployment name things are labels and not annotations.

jandubois:
I've tried twice now to get this PR through CI, but both times it failed during smoke testing with Diego (it worked fine with Eirini):

https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/smoke-tests-diego-pr/builds/229
https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/smoke-tests-diego-pr/builds/230

Just triggered the Diego pipeline one more time, but it certainly feels weird that it fails twice in a row in the same tests.

jandubois:
> Just triggered the Diego pipeline one more time, but it certainly feels weird that it fails twice in a row in the same tests.

Looks like 3rd time's a charm...

PR #1542, opened 2020-11-02
fix: downgrade sle15 rootfs because the latest image does not exist

jandubois:
The 24.81 rootfs has only been build with the previous 27.8 stemcell. This issue will self-correct once there is another update to the rootfs, which happens pretty frequently.


jbuns (approved):
LGTM

PR #1543, opened 2020-11-03
Fix '/instance_groups/name=api' not found when multiple_cluster_mode.control_plane is disabled

JimmyMa:
## Description
  The multiple-cluster-mode-operations removes '/instance_groups/name=api' when multiple_cluster_mode.control_plane is disabled, so any other operations after multiple-cluster-mode-operations gets error if they try to operate '/instance_groups/name=api'.

  The multiple-cluster-mode-operations must be the botton of chart/templates/bosh_deployment.yaml

## How Has This Been Tested?
Tested in a shard cluster with below values:

```
features:
  # To support multi-clusters, deploy diego-cell separately please set control_plane is false  and cell_segment is true
  multiple_cluster_mode:
    control_plane:
      enabled: false
    cell_segment:
      enabled: true
```

## Types of changes

## Checklist:

### DIFF
--- a/chart/templates/bosh_deployment.yaml
+++ b/chart/templates/bosh_deployment.yaml
@@ -78,12 +78,6 @@ spec:
     type: configmap
 {{- end }}
 
-# To support multi-clusters
-{{- if or .Values.features.multiple_cluster_mode.control_plane.enabled .Values.features.multiple_cluster_mode.cell_segment.enabled }}
-  - name: multiple-cluster-mode-operations
-    type: configmap
-{{- end }}
-
 # Custom operations
 {{- range $_, $ops := .Values.operations.custom }}
   - name: {{ $ops | quote }}
@@ -95,3 +89,9 @@ spec:
 {{- end }}
   - name: user-provided-properties
     type: configmap
+
+# To support multi-clusters, NOTE: this section must be the bottom of this file
+{{- if or .Values.features.multiple_cluster_mode.control_plane.enabled .Values.features.multiple_cluster_mode.cell_segment.enabled }}
+  - name: multiple-cluster-mode-operations
+    type: configmap
+{{- end }}


jandubois:
I don't understand how this would create any errors; the multi-cluster ops are already very close to the bottom.

The only ops applied even later are the customs ops and property definitions provided by the user via `values.yaml`. They are intentionally the last ops so that they can be used to override anything that the built-in ops have produced.

If the user provides customs ops that affect an instance group that doesn't even exist in the deployment, then that is a config error, and not a problem in kubecf itself.

So from that perspective I disagree that the error should be hidden, especially if the price for this is reducing the flexibility of what can be overridden in custom ops and properties.

Can you provide an example of the issue you are dealing with?

JimmyMa:
@jandubois hello Jan, below two files define some properties for job api and  cc-worker:

https://github.com/cloudfoundry-incubator/kubecf/blob/master/chart/config/api.yaml#L1-L8 
https://github.com/cloudfoundry-incubator/kubecf/blob/master/chart/config/jobs.yaml#L28-L40

Then the generated user-provided-properties configmap has below content, and that caused the '/instance_groups/name=api' not found error when deploy a cell segment which has only dieog-cell job.

```
kubectl -n kubecf get configmap user-provided-properties -o yaml
apiVersion: v1
data:
  ops: |-
    - type: replace
      path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties?/build
      value: "2.6.1-dirty"
    - type: replace
      path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties?/cc/jobs/local/number_of_workers
      value: 2
    - type: replace
      path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties?/description
      value: "CloudFoundry KubeCF"
    - type: replace
      path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties?/name
      value: "KubeCF"
    - type: replace
      path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties?/support_address
      value: "https://kubecf.io/"
    - type: replace
      path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties?/version
      value: 2
    - type: replace
      path: /instance_groups/name=cc-worker/jobs/name=cloud_controller_worker/properties?/cc/jobs/generic/number_of_workers
      value: 1
```

mook-as:
I can confirm that I can see the error (though I didn't actually deploy); this basically boils down to having two parallel systems here (the ops files done at the operator, and the configs stuff at helm templating time).

I believe the correct fix here would look more like #1480 (judging by the PR title; I hadn't actually looked at the changes yet).

jandubois:
#1615 is supposed to fix the issue while still allowing user supplied ops and properties to override any settings generated by the chart.

jandubois:
@JimmyMa #1615 has been merged to `master`, so please test, and re-open if it doesn't fix your issue!

PR #1550, opened 2020-11-04

cfcontainerizationbot:


jandubois (changes requested):
```
Error: failed to download "https://s3.amazonaws.com/cf-operators/release/helm-charts/cf-operator-v7.0.0-pre.0.tgz" (hint: running `helm repo update` may help)
```

cc/ @manno

jandubois:
Looks like the `7.0-pre` tag was not supposed to be a proper release...

PR #1559, opened 2020-11-04
fix: chart: Axe tcp-router-public if unavailable

mook-as:
## Description
If routing-api is not enabled, then remove the tcp-router-public service (when not using ingress).  Otherwise we end up with a service that will never point at anything.

## Motivation and Context
Discovered while working on #1144.
We will never finish waiting for services to be ready if we have a service that can never be ready.

## How Has This Been Tested?
Tested locally (and in the WIP CI stuff); I can actually finish waiting now.

## Types of changes

## Checklist:

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -229,7 +229,7 @@ deploy_args: &deploy_args
   # Setup dns
   tcp_router_ip=$(kubectl get svc -n scf tcp-router-public -o json | jq -r .status.loadBalancer.ingress[].ip | head -n 1)
   # tcp-router svc may not exist, if so, use an rfc5737 example IP address
-  tcp_router_ip=${$tcp_router_ip:-192.0.2.0}
+  tcp_router_ip="${tcp_router_ip:-"192.0.2.0"}"
   public_router_ip=$(kubectl get svc -n scf router-public -o json | jq -r .status.loadBalancer.ingress[].ip | head -n 1)
   # ssh-proxy service is named differently in diego and eirini
   ssh_proxy_svc=$(kubectl get svc -n scf | grep ssh-proxy | awk '{print $1}' | head -n 1)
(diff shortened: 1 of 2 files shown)


andreas-kupries (dismissed):
LGTM. Brain tests has the expected failures for tcp-routing and insecure registry (both of which require tcp-routing). Smoke passed.

viccuad:
Hm. Concourse CI fails on the gcloud DNS transaction, as there's no tcp-router-ip for the service:
https://concourse.suse.dev/builds/40662#L5fabb78f:615
https://github.com/cloudfoundry-incubator/kubecf/blob/aec0b2e01b68e54f54a62b756457a3aafbfae88f/.concourse/pipeline.yaml.gomplate#L240-L241

Concourse CI doesn't use Catapult nor external-dns for the creation of GKE clusters.

To not overcomplicate things for now, I suppose we could just give a valid but bogus value to `$tcp_router_ip` on the pipeline, when the service is not present. The correct way would be to either use external-dns properly (moving the pipeline code to Catapult) or read `features.routing_api` which is not there until later.

andreas-kupries (dismissed):
It LGTM.
Although I would have broken the change over two lines instead of nesting, i.e.
```
<original line>
tcp_router_ip="${tcp_router_ip:-FIXED_IP}"
```
But that is a nit, and can be argued as style (Of course I would argue it is readability/clarity :wink: ).

viccuad:
Just realized that for testing this PR, either we fly the pipeline from this branch, or deploy a copy of the pipeline with config including:
```
github_status: true
branches: # Repository branches to track
- marky/service-no-tcp
pr_resources:
- pr
pr_base_branch: marky/service-no-tcp # filter PRs depending on the branch they target. `~` for all branches
```

Tried the second one, but flying failed with:
`error: yaml: line 427: did not find expected alphabetic or numeric character`. Looking into it I think that the pr resource doesn't support `/` chars. Woould be weird. Will continue tomorrow.

viccuad:
I have flown the Concourse changes to the kubecf pipeline, the PR should pass now.

viccuad:
Reflown changes to pipeline, other PRs went through with them, so the are validated (e.g: https://github.com/cloudfoundry-incubator/kubecf/pull/1593). yes, that it's reckless..

thardeck (approved):

Issue #1047, opened 2020-06-19
Use the vault service for the kubecf pipelines


viovanov:
## Acceptance Criteria

- all pipelines for the kubecf ci use the vault service as a source of secrets.

viovanov:
Blocked by #1045

viovanov:
This ticket is about using the vault action in the kubecf github actions pipeline

andreas-kupries:
To fill things in, we are talking about

  - `.github/workflows/conventional-commits-lint.yml`
  - `.github/workflows/pull-request-ci.yaml`
  - `.github/workflows/pull-request-queue.yaml`

in this repository. The secrets currently used by these flows are

  - `secrets.GITHUB_TOKEN` (alt: `secrets.github_token` ?! (both forms are used))
  - `secrets.GH_BOT_TOKEN`
  - `secrets.GKE_CRED_JSON`

For the first two we have multiple possibilities, all in `secret/data/kubecf`:

  - `github.cfcontainerizationbot-labeler-personal-access-token`
  - `github-access-token`
  - `github-limited-token`

@viovanov, @mook-as Do you know the relevant mapping ?

The last item on the other hand is trivial, it maps to `secret/data/cloud-providers`, key `gke.GKE_CRED_JSON`.

andreas-kupries:
Note, further: The vault action will need the VAULT_TOKEN (same as we did for pxc-docker).
I do not have the ability to add a GHA secret for that to this repo.

mook-as:
`secrets.GITHUB_TOKEN` is the same as `github.token`, which is supplied by GitHub Actions and we don't have to do anything about it.

I don't know about the other ones; as we can't dig it out of the GitHub secrets UI, we'll need to look at their use to fix things.  I suspect the one in `pull-request-queue.yaml`, for example, should map to `cfcontainerizationbot-labeler-personal-access-token` (because it adds a label to the PR).

andreas-kupries:
Thank you, that was exactly what I was looking for.

PR #1567, opened 2020-11-06
fix: switch flows over to usage of suse vault for secrets.

andreas-kupries:
## Description

- Added actions to the flow accessing the suse vault (volt).
- Adapted secret access in other actions to match.

## Motivation and Context

See #1047

## How Has This Been Tested?

It has not, not yet.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -186,7 +186,7 @@ jobs:
         if: env.BACKEND == 'gke'
         run: |
           json_file="$(mktemp)"
-          echo "${{ steps.vault.outputs.GKE_CRED_JSON }}" > "${json_file}"
+          cat > "${json_file}" <<< "${GKE_CRED_JSON}"
           echo "GKE_CRED_JSON=${json_file}" >> "${GITHUB_ENV}"
         timeout-minutes: 1
 
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -189,6 +189,8 @@ jobs:
           cat > "${json_file}" <<< "${GKE_CRED_JSON}"
           echo "GKE_CRED_JSON=${json_file}" >> "${GITHUB_ENV}"
         timeout-minutes: 1
+        env:
+          GKE_CRED_JSON: ${{ steps.vault.outputs.GKE_CRED_JSON }}
 
       # Cache catapult's common tools
       - name: cache.catapult-common-tools
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -177,7 +177,7 @@ jobs:
         with:
           exportEnv: false
           url: https://volt.cap.explore.suse.dev/
-          token: ${{ secrets.VAULT_TOKEN }}
+          token: ${{ secrets.VAULT_TOKEN_CLOUD_PROVIDERS }}
           secrets: |
             secret/data/cloud-providers gke.GKE_CRED_JSON | GKE_CRED_JSON
 
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -171,15 +171,23 @@ jobs:
           echo "SSH_AGENT_PID=${SSH_AGENT_PID}" >> "${GITHUB_ENV}"
         timeout-minutes: 1
 
-      # set GKE secret. TBD: use vault instead
+      - name: Query vault for GKE secrets
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          exportEnv: false
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/cloud-providers gke.GKE_CRED_JSON | GKE_CRED_JSON
+
+      # set GKE secret.
       - name: set GKE_CRED_JSON
         if: env.BACKEND == 'gke'
         run: |
           json_file="$(mktemp)"
-          echo "$GKE_CRED_JSON" > "${json_file}"
+          echo "${{ steps.vault.outputs.GKE_CRED_JSON }}" > "${json_file}"
           echo "GKE_CRED_JSON=${json_file}" >> "${GITHUB_ENV}"
-        env:
-          GKE_CRED_JSON: ${{ secrets.GKE_CRED_JSON }}
         timeout-minutes: 1
 
       # Cache catapult's common tools
--- a/.github/workflows/pull-request-queue.yaml
+++ b/.github/workflows/pull-request-queue.yaml
@@ -14,9 +14,19 @@ jobs:
     name: enqueue
     runs-on: ubuntu-20.04
     steps:
+      - name: Query vault
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          exportEnv: false
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/kubecf "github.cfcontainerizationbot-labeler-personal-access-token" | GH_BOT_TOKEN
+
       - uses: actions-ecosystem/action-add-labels@9141732
         with:
           # We need a bot token because the regular GH Actions token is not allowed to trigger
           # secondary GH Actions events.
-          github_token: ${{ secrets.GH_BOT_TOKEN }}
+          github_token: ${{ steps.vault.outputs.GH_BOT_TOKEN }}
           labels: pr-test-queue
--- a/.github/workflows/pull-request-queue.yaml
+++ b/.github/workflows/pull-request-queue.yaml
@@ -22,7 +22,7 @@ jobs:
           url: https://volt.cap.explore.suse.dev/
           token: ${{ secrets.VAULT_TOKEN }}
           secrets: |
-            secret/data/kubecf "github.cfcontainerizationbot-labeler-personal-access-token" | GH_BOT_TOKEN
+            secret/data/kubecf "github-access-token" | GH_BOT_TOKEN
 
       - uses: actions-ecosystem/action-add-labels@9141732
         with:


.github/workflows/pull-request-ci.yaml
```diff
@@ -199,14 +199,21 @@ jobs:
...
 
-      # set GKE secret. TBD: use vault instead
+      - name: Query vault for GKE secrets
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/cloud-providers "gke.GKE_CRED_JSON" | GKE_CRED_JSON
```
mook-as, line 209: We could possibly set `exportEnv` to `false`, and then use the output explicitly in the next step, to ensure that we don't accidentally leak the contents later.  Not that it matters that much, since people can read the file… but it helps to avoid accidentally printing it out when dumping the environment.

.github/workflows/pull-request-ci.yaml
```diff
@@ -199,14 +199,21 @@ jobs:
           echo "::set-env name=SSH_AUTH_SOCK::${SSH_AUTH_SOCK}"
           echo "::set-env name=SSH_AGENT_PID::${SSH_AGENT_PID}"
 
-      # set GKE secret. TBD: use vault instead
+      - name: Query vault for GKE secrets
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
```
mook-as, line 207: Hmm, there's a `github` auth method; if that works, we won't need to set a token (that we'd need to rotate), since `github.token` is automatic.  Not sure _who_ that token identifies, though; so if it doesn't use the right one out of the box there's no point (i.e. setting a token for a user and manually setting it as a secret has no point).

.github/workflows/pull-request-queue.yaml
```diff
@@ -14,9 +14,18 @@ jobs:
...
     runs-on: ubuntu-20.04
     steps:
+      - name: Query vault
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/kubecf "github.cfcontainerizationbot-labeler-personal-access-token" | GH_BOT_TOKEN
```
mook-as, line 24: Since we refer to it by `steps.…` we can set `exportEnv` to `false` too.

.github/workflows/pull-request-ci.yaml
andreas-kupries: This is the same token @viovanov set up for pxc-docker. And no, this is not github token. I remember, I thought that it was GH token and that was then the first issue the GHA flow for pxc-docker had.

.github/workflows/pull-request-queue.yaml
andreas-kupries: See commit f3c7c8d8411a3c291ca677ec5be23283aa41c778

.github/workflows/pull-request-ci.yaml
andreas-kupries: See commit f3c7c8d8411a3c291ca677ec5be23283aa41c778

.github/workflows/pull-request-ci.yaml
viccuad: if it's a github token, it would suffice for it to be a read:org GH token for an account that is on the corresponding team, (maybe `suse-platform-team` GH team?).

.github/workflows/pull-request-ci.yaml
```diff
@@ -199,14 +199,22 @@ jobs:
...
-      # set GKE secret. TBD: use vault instead
+      - name: Query vault for GKE secrets
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          exportEnv: false
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/cloud-providers "gke.GKE_CRED_JSON" | GKE_CRED_JSON
```
mook-as, line 210: Hmm, when trying this out on elsewhere, it seems like I have to _not_ quote the key:
```suggestion
            secret/data/cloud-providers gke.GKE_CRED_JSON | GKE_CRED_JSON
```
YMMV, I have no idea what I'm doing.

andreas-kupries: Weird. It might be something with JSONata (that is the jq-like library/package used to process the key value).
If your key contains `-` for example (dash, minus) the quoting is required to prevent mishandling of this char as operator instead of part of key.

That said, is your `elsewhere` another GH action, or is it the vault cli ?

mook-as: The `elsewhere` is #1592.

andreas-kupries: Accepted suggestion. Lets see how that goes.

.github/workflows/pull-request-ci.yaml
mook-as: Tried it over in #1592; it turns out trying to use the Actions token results in a HTTP 500 from Vault instead (probably because it had problems resolving that token to an actual person).  So that's not going to work.  Marking this as resolved.

.github/workflows/pull-request-ci.yaml
```diff
@@ -199,14 +199,22 @@ jobs:
...
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/cloud-providers gke.GKE_CRED_JSON | GKE_CRED_JSON
+
+      # set GKE secret.
       - name: set GKE_CRED_JSON
         run: |
           json_file="$(mktemp)"
-          echo "$GKE_CRED_JSON" > "${json_file}"
+          echo "${{ steps.vault.outputs.GKE_CRED_JSON }}" > "${json_file}"
```
mook-as, line 216: Hmm. Technically this means the token will be in the process command line here, right? (Though I guess `echo` is _probably_ a built in and doesn't have a real command line?)  Do we have to worry about it showing up in the bash history? (This probably isn't interactive, so… probably not?)

Generally using the environment is better.

.github/workflows/pull-request-ci.yaml
```diff
@@ -199,14 +199,22 @@ jobs:
...
           echo "::set-env name=SSH_AGENT_PID::${SSH_AGENT_PID}"
 
-      # set GKE secret. TBD: use vault instead
+      - name: Query vault for GKE secrets
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          exportEnv: false
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
```
viccuad, line 208: We could also pin the CA cert of volt.cap.explore.suse.dev, using `caCertificate` (see https://github.com/marketplace/actions/vault-secrets), instead of using the trust chain of RootCert, IntermediateCert, VoltCert. But maybe it's more hassle than anything.

.github/workflows/pull-request-ci.yaml
viccuad: I agree that `echo` without `set -x` is ok, is a built-in of the shell.

I wonder how GH Actions redacts things, e.g:
https://github.com/cloudfoundry-incubator/kubecf/runs/1216923853?check_suite_focus=true#step:10:3.
Looking at https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs, it seems that the secret `steps.vault.outputs.GKE_CRED_JSON` would get redacted in the worker and never shared with GHA.

> Generally using the environment is better.

I don't personally agree, as env can get leaked to subprocesses and utilities you call (as normally you don't unset them), and env usually gets logged or dumped somewhere. I try to always use files, in memory with tmpfs, mktemp (what we are doing here), named fifos, or by mounting the files into whatever corresponds. If you were to have access to the file, you would have access to the env anyways. But at least you don't usually log the file contents somewhere.

.github/workflows/pull-request-ci.yaml
```diff
@@ -199,14 +199,22 @@ jobs:
...
+
+      # set GKE secret.
       - name: set GKE_CRED_JSON
         run: |
           json_file="$(mktemp)"
-          echo "$GKE_CRED_JSON" > "${json_file}"
+          echo "${{ steps.vault.outputs.GKE_CRED_JSON }}" > "${json_file}"
           echo "::set-env name=GKE_CRED_JSON::${json_file}"
-        env:
-          GKE_CRED_JSON: ${{ secrets.GKE_CRED_JSON }}
```
viccuad, line 218: Here we are not setting the env var `GKE_CRED_JSON` with  `GKE_CRED_JSON: ${{ secrets.GKE_CRED_JSON }}` anymore.

I think we need to set `GKE_CRED_JSON` to the path of `"${json_file}"`, as Catapult expects an absolute path on it, pointing to a file containing the gke json secret.

Something like:
```suggestion
        env:
          GKE_CRED_JSON: $(readlink -f ${json_file})

```

viccuad:
The conflicts with master come from merging https://github.com/cloudfoundry-incubator/kubecf/pull/1533 (I'm not fixing them yet to not mark all the review comments as outdated).

.github/workflows/pull-request-ci.yaml
andreas-kupries: Will fix.

.github/workflows/pull-request-ci.yaml
andreas-kupries: PR is rebased. Left this as-is for now.

.github/workflows/pull-request-ci.yaml
andreas-kupries: PR rebased, conflicts resolved. Not doing anything with certs.

.github/workflows/pull-request-ci.yaml
andreas-kupries: PR rebased, conflicts resolved. Now I can do the fix here.

andreas-kupries: After resolving conflicts the configuration looks like this:
```
        run: |
          json_file="$(mktemp)"
          echo "${{ steps.vault.outputs.GKE_CRED_JSON }}" > "${json_file}"
          echo "GKE_CRED_JSON=${json_file}" >> "${GITHUB_ENV}"
```

This is the new way of setting the GHA environment. Note that `json_file` is a path generated by `mktemp`.
That is already an absolute path.
I do not see anything to be done here anymore.

.github/workflows/pull-request-ci.yaml
```diff
@@ -171,15 +171,23 @@ jobs:
...
         timeout-minutes: 1
 
-      # set GKE secret. TBD: use vault instead
+      - name: Query vault for GKE secrets
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          exportEnv: false
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
```
mook-as, line 180: Huh, I thought `secrets.VAULT_TOKEN` didn't have the correct access, and we needed `secrets.VAULT_TOKEN_CLOUD_PROVIDERS` for that; see the scheduler workflow.

https://github.com/cloudfoundry-incubator/kubecf/blob/7762c65756fde37a126f1f5e701bb6056c93f55a/.github/workflows/pull-request-schedule.yaml#L40-L46

.github/workflows/pull-request-queue.yaml
```diff
@@ -14,9 +14,19 @@ jobs:
...
     steps:
+      - name: Query vault
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          exportEnv: false
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/kubecf "github.cfcontainerizationbot-labeler-personal-access-token" | GH_BOT_TOKEN
```
mook-as, line 25: I believe we can use `github-access-token` instead (i.e. @suse-cf-ci-bot) instead of this one (which would be @cfcontainerizationbot).

There's also`github-limited-token` but I'm not sure what the differences in permission are; might be useful to try and see if that is enought to set labels.

mook-as (changes requested):

.github/workflows/pull-request-ci.yaml
andreas-kupries: Addressed with commit abcd4410dac0c7a0970ad5ea69f6b01bbf05a048

.github/workflows/pull-request-queue.yaml
andreas-kupries: Addressed with commit abcd4410dac0c7a0970ad5ea69f6b01bbf05a048

.github/workflows/pull-request-queue.yaml
```diff
@@ -14,9 +14,19 @@ jobs:
...
     steps:
+      - name: Query vault
+        id: vault
+        uses: hashicorp/vault-action@v2.0.1
+        with:
+          exportEnv: false
+          url: https://volt.cap.explore.suse.dev/
+          token: ${{ secrets.VAULT_TOKEN }}
+          secrets: |
+            secret/data/kubecf "github-access-token" | GH_BOT_TOKEN
```
mook-as, line 25: Hmm, does this work? I though we needed to use backticks for quoting here (because JSONata)?

mook-as (commented):
LGTM, but please let me know once we've ran this somewhere; I _think_ we need different quoting.

.github/workflows/pull-request-ci.yaml
```diff
@@ -171,15 +171,23 @@ jobs:
...
+          secrets: |
+            secret/data/cloud-providers gke.GKE_CRED_JSON | GKE_CRED_JSON
+
+      # set GKE secret.
       - name: set GKE_CRED_JSON
         if: env.BACKEND == 'gke'
         run: |
           json_file="$(mktemp)"
-          echo "$GKE_CRED_JSON" > "${json_file}"
+          echo "${{ steps.vault.outputs.GKE_CRED_JSON }}" > "${json_file}"
```
mook-as, line 189: Okay, I actually tried to deploy this and this _definitely_ doesn't work.
The vault output contains the quote character (`"`), and because the `${{}}` is interpreted by GitHub, the shell command ends up being:
```bash
echo "{"blah": "foo", …}"
```
Of course, this means that the actual command run is (replacing the quoting):
```bash
echo '{blah: foo, …}'
```
That's no longer valid JSON, so it falls over in later steps.

mook-as (changes requested):

.github/workflows/pull-request-ci.yaml
andreas-kupries: So, if I understand correct, I should remove the outer quotes ?
IOW

> echo ${{ steps.vault.outputs.GKE_CRED_JSON }} > "${json_file}"

On the other side, I see

> cat > "${json_file}" <<< "${GKE_CRED_JSON}"

in the `pull-request-schedule.yaml` of oyurs (IIRC).
Hm, not sure why that works, as you have `exportEnv: false` in the step before, retrieving the vault data.

mook-as: No, because if you removed the outer quotes the whole thing gets shell expansion.
`pull-request-schedule.yaml` has an associated `env:` block _for that step only_ which sets the variable.  In the quoted line, `"${GKE_CRED_JSON}"` is expanded by the shell, but only once; so the value (complete with quotes) is passed as-is as standard input.  As it is right now in this PR, `${{ … }}` is completed by the GitHub workflow (rather than the shell it spawns), so the shell is not aware of that and only sees the expanded values.

.github/workflows/pull-request-queue.yaml
andreas-kupries: In some previous work with Vlad the issue with the dash/minus characters in the field name was fixed by double-quotes. so I used them here also.

.github/workflows/pull-request-ci.yaml
andreas-kupries: @mook-as I now have a local commit going with your form (`cat ...`).

How can I test this ?

Could I
- make a fork of the repo, then
- merge the branch into the master of the fork,

and that would trigger GHA, but on the fork ?

mook-as: Yep, that should work; remember to apply the `pr-test-queue` label (or `pr-test-trigger` if you wan to skip the queuing step) to the PR to get it picked up.  Please ping me on Slack / 🚀 if you need assistance.

andreas-kupries: Will do. Right now meeting...

andreas-kupries: I missed the `has an associated env: block for that step only which sets` on my first look. Maybe gh page was not updated. That explains things and I have now put this into my spec too.

.github/workflows/pull-request-queue.yaml
mook-as: Seems to work in actual deployment, so 👍

mook-as (approved):
Seems to work fine in an actual test run.

PR #1573, opened 2020-11-09
fix: copy githubstatus and github-pr-resource images to ghcr

jandubois:
Just to avoid the pipeline from stalling after hitting the dockerhub bandwidth limit.

Moving them to my personal Github account it temporary until we figure out which org we can use for stable names.

The pipeline has already been deployed.

### DIFF
--- a/.concourse/pipeline.yaml.gomplate
+++ b/.concourse/pipeline.yaml.gomplate
@@ -29,14 +29,16 @@ resource_types:
 - name: pull-request
   type: docker-image
   source:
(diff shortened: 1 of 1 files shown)


mook-as (approved):
We still need to keep tickets to make them live outside the personal package space (or dump concourse completely and use GitHub Actions only).

viccuad:
Cloding, superseded by https://github.com/cloudfoundry-incubator/kubecf/pull/1585

PR #1575, opened 2020-11-09
fix: autoscaler: Drop control script

mook-as:
## Description
This drops the BOSH-style control script for the autoscaler golang apiserver, and uses BPM to directly launch the binary.

I suspect I may be missing features in that control script that the IBM folks are using; however GitHub doesn't seem to like me. @aqan213 — would you have some time to take a look?

## Motivation and Context
I suspect that the control script (having been designed for monit) uses a PID file, which doesn't work well when the container restarts.  I'm seeing issues on some deployments (on CI) where that pod gets stuck at an unready state (crashing repeatedly); deleting the pod and letting the deployment spin up a new one works around the issue.

## How Has This Been Tested?
This was cherry-picked out of my branch related to GitHub Actions work for #1144; I haven't had issues related to the autoscaler not being ready since I made the change.

## Types of changes

## Checklist:

### DIFF
--- a/chart/assets/operations/instance_groups/app-autoscaler.yaml
+++ b/chart/assets/operations/instance_groups/app-autoscaler.yaml
@@ -923,9 +923,10 @@
     bpm:
       processes:
       - name: golangapiserver
-        executable: /var/vcap/jobs/golangapiserver/bin/apiserver_ctl
+        executable: /var/vcap/packages/golangapiserver/api
         args:
-        - start
+        - -c
+        - /var/vcap/jobs/golangapiserver/config/apiserver.yml
     run:
       healthcheck:
         golangapiserver:


andreas-kupries:
@mook-as The ctl script supports pre/post start/stop hooks.

mook-as:
@andreas-kupries Yes, but as far as I can tell we don't actually use that capability.

andreas-kupries:
Yes. Thus the question to @aqan213 if IBM does, right ?

aqan213:
@mook-as @andreas-kupries  Since autoscaler does not support bpm for now(it supports monit). We need to migrate to bpm(refer to https://bosh.io/docs/bpm/transitioning/#converting-_ctl-scripts-to-bpmyml). We actually use the start command only in kubecf and can modify it just like Mark does.
```
- name: golangapiserver
  executable: /var/vcap/packages/golangapiserver/api
  args:
  - -c
  - /var/vcap/jobs/golangapiserver/config/apiserver.yml
```

mook-as:
Thanks for the answer!

Okay, sounds like it should be okay to merge this then? r?@andreas-kupries then :)

andreas-kupries (approved):
:+1:

PR #1576, opened 2020-11-09
fix: chart: bump resource limits

mook-as:
## Description
Bump the memory limit for the cloud controller nginx, as well as the blobstore nginx.  As we're only using those limits to prevent runaway memory consumption, this should be safe.  I did not touch the requests.

## Motivation and Context
My CI runs were failing because the containers were being OOMKilled in the middle of the run (which was only discovered upon examining the logs dumped at the end of a run).

## How Has This Been Tested?
Ran CATS &c runs on GitHub Actions; the tests have been passing since the changes.

## Types of changes

## Checklist:

### DIFF
--- a/chart/config/resources.yaml
+++ b/chart/config/resources.yaml
@@ -118,7 +118,7 @@ resources:
   api:
     cloud_controller_ng:
       cloud_controller_ng: {memory: {limit: 2048, request: 1024}}
-      nginx: 192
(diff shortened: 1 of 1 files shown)


jandubois (commented):
Looks good to me; I would approve, but CI is currently dead, so it won't do any good.

viccuad (approved):

Issue #1557, opened 2020-11-04
CI is broken because of Dockerhub bandwidth limits


jandubois:
The last few PR builds failed to report status back to Github:

```
failed to fetch digest for image 'resource/github-status:release': 429 Too Many Requests
does the image exist?
```

We should move all images used by Concourse to the Github registry instead.  The above image is from https://github.com/colstrom/concourse-github-status

But we are still pulling many more images from Dockerhub, at the very least:

```
cloudfoundry-incubator/kubecf
resource/github-status
splatform/base-ci
splatform/catapult
splatform/fissile-stemcell-sle
splatform/github-pr-resource
```

viccuad:
`splatform/catapult` is being built on concourse.suse.de, from https://github.com/SUSE/cloudfoundry/blob/master/ci/pipelines/catapult/pipeline.yml. This may be the case for all the `splatform` org images.

jandubois:
I've moved the the 2 images used by the pipeline to `ghcr.io/jandubois`:

```
resource/github-status
splatform/github-pr-resource
```

Especially the `github-pr-resource` is downloaded way more often than I expected.

Anyways, we need to create a Github organization that can become the owner of all these images; they should not be stored under a personal account.

The name should not mention `concourse` but be generic enough to include also all the images for bosh releases, for stemcells, and for the quarks operator.

viccuad:
@jandubois honest thanks for moving the images; that should have been me, and I was blocking CI.

@viovanov could we create the GH org now? or should we reuse cloudfoundry-incubator as [quarks does](https://github.com/orgs/cloudfoundry-incubator/packages)? In the second case, we would need credentials for it.

Once we have a target org, we can change the target of [suse-buildpacks-ci](https://github.com/cloudfoundry-incubator/kubecf/tree/master/.concourse/suse-buildpacks-ci) to also publish there.

viovanov:
@viccuad please use the incubator org

jandubois:
I've added some comments on #1581 about images for autoscaler and eirini.

viccuad:
Prereequisites:
- [x] ghcr.io/cloudfoundry-incubator on vault and concourse

For all images to migrate:
1. They need to be pushed to ghcr.io/cloudfoundry-incubator
2. They need to be made public manually on ghcr.io
3. We need to consume them in the helm chart

Possible incomplete list of images to migrate:
- [ ] splatform/github-pr-resource
      resource/github-status
[...]

PR #1581, opened 2020-11-11
fix: use images from Github Container Registry

jandubois:
This PR uses the images on ghcr.io built with the latest stemcell.

The following images still reside on Dockerhub:

```
docker.io/cfcontainerization/coredns
```

Ref #1557

### DIFF
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -53,11 +53,11 @@ releases:
     condition: false
   postgres:
     condition: features.autoscaler.enabled
-    version: "39"
+    version: "42"
   sync-integration-tests:
     # XXX SITS only makes sense when using Diego; add error check somewhere?
     condition: testing.sync_integration_tests.enabled
-    version: v0.0.3
+    version: v0.0.2
 
 unsupported:
   ? releases.defaults
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -42,7 +42,7 @@ releases:
   database:
     # database (previously known as pxc) is not a BOSH release.
     image:
-      repository: docker.io/cfcontainerization/pxc
+      repository: ghcr.io/cloudfoundry-incubator/pxc
       tag: 0.9.11
   metrics-discovery:
     # metrics-agent add-on (new in cf-deployment 13.0) is not used by kubecf
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -7,10 +7,10 @@
 releases:
   # The defaults for all releases, where we do not otherwise override them.
   '$defaults':
-    url: docker.io/cfcontainerization
+    url: ghcr.io/cloudfoundry-incubator
     stemcell:
       os: SLE_15_SP1
-      version: 27.5-7.0.0_374.gb8e8e6af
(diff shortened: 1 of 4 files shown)


jandubois:
There are 2 images in a different org that probably should be moved to `cloudfoundry-incubator`:

```
ghcr.io/cfcontainerizationbot/kubecf-apps-dns
ghcr.io/cfcontainerizationbot/kubecf-kubectl
```

jandubois:
I found the current `postgres` version does not exist for autoscaler. There are newer releases, so I tried this patch:

```
--- a/chart/config/releases.yaml
+++ b/chart/config/releases.yaml
@@ -53,7 +53,7 @@ releases:
     condition: false
   postgres:
     condition: features.autoscaler.enabled
-    version: "39"
+    version: "42"
   sync-integration-tests:
     # XXX SITS only makes sense when using Diego; add error check somewhere?
     condition: testing.sync_integration_tests.enabled
```

But then `asdatabase-0` pod never got ready. This blocks init containers in other autoscaler pods, so I don't know what the status of the autoscaler images is.

mook-as:
The autoscaler issue was a hard-coded version in a path in the readiness probe; pushed a commit to make that work.  Also, bump the SITS version _down_ since v0.0.3 doesn't seem to be on GHCR (will need to address that in a bit).

viccuad:
Full KubeCF installations from this PR should now work. Ready for review. 

Comitted the change for eirinix images, created from the updated eirinix Concourse pipeline. Some images where missing from ghcr.io, as they predate the pipeline change on publishing, so I copied those tagged images manually to  ghcr.io/cloudfoundry-incubator.

Tested by deploying diego, eirini, ha & autoscaler successfully. Only missing image was `docker.io/cfcontainerization/pxc:0.9.11`

Which I manually copied to `ghcr.io/cloudfoundry-incubator/pxc:0.9.11` meanwhile we wait for resolution of https://github.com/SUSE/cloudfoundry/pull/201.

chart/config/releases.yaml
```diff
@@ -42,7 +42,7 @@ releases:
   database:
     # database (previously known as pxc) is not a BOSH release.
     image:
-      repository: docker.io/cfcontainerization/pxc
+      repository: ghcr.io/cloudfoundry-incubator/pxc
```
viccuad, line 45: Is this the correct way to change the repository of this image? or can one drop the repository key and use `releases.$defaults`?

viccuad:
Instead of opening a new PR for consuming the pxc image, I have edited the title and body of the PR and expanded the PR scope to all images, not only bosh-release images.

My intent was to have only 1 PR with all the new images, so it's easy to pass through CI.

chart/config/releases.yaml
mook-as: Yes, because the helm template doesn't respect `releases.$defaults`:
https://github.com/cloudfoundry-incubator/kubecf/blob/a336ea4b70b9c940a0c82d2f8d0abc44b4807937/chart/templates/database/db-statefulset.yaml#L49-L53

mook-as (approved):
👍 LGTM

viovanov (approved):

Issue #1148, opened 2020-07-21
Internet-less suite for GH Actions


viovanov:
## Acceptance Criteria

- ability to control trafic in AKS (to be done in catapult)
- ability to run the internetless CATS suite from kubecf (move code from catapult)
- enable the internetless CATS suite

Blocked by #1144

viccuad:
Setting an airgap via networkPolicy in catapult GKE is possible with https://github.com/SUSE/catapult/pull/317. 
Configuration of cats for internetless is still happening at https://github.com/SUSE/catapult/blob/0f85c1d66e7612a81f4c20442c598052e886705f/modules/tests/kubecf-test.sh#L127-L147, and its what should be ported to kubecf/scripts/*.

Issue #1147, opened 2020-07-21
Upgrade tests in GH Actions


viovanov:
## Acceptance Criteria

- PR checking should test upgrade scenarios for KubeCF

Blocked by #1144

viccuad:
From retros, I remember we needed:

- Upgrade from previous latest minor version to release under test.
- Upgrade from previoust latest major version (starting with 2.X) to release under test.
- Upgrade from release under test to release under test.
- Upgrade from release under test to future, non-existent yet, release.

mook-as:
This should be doable now, because the bulk of the GH Actions work is in.

To test things, (in your GitHub fork) label PRs with the `pr-test-trigger` label and the CI action should pick it up.

PR #1599, opened 2020-11-18
fix: GitHub Actions improvements

mook-as:
## Description
Various small CI changes to make running tests more stable; see commits for details.

### Cloud Provider Changes
- Skip AKS steps on GKE, etc.
- Bump on GKE cluster size, and use non-preempt for CI.
  - They should be cleaned up correctly by the [scheduler workflow](https://github.com/cloudfoundry-incubator/kubecf/blob/master/.github/workflows/pull-request-schedule.yaml), so there's less pressure to keep things cheap in case they get orphaned
- Set up the GKE config directory (the way it is set it catapult), so that `kubectl` can automatically refresh the access token if a build takes a long time (e.g. running CATS).

### Workflow Changes
- The actions need to run against the PR branch, still, so that it checks out the correct source to test against.
- Move the config values generation to a shell script, because it was rather unwieldy
  - Various changes to bring it up to date
- Add timeouts to most steps, so we don't get stuck runs taking too long that would never succeed.
- Fix interaction with how we determine if we're running Eirini
- Wait for DNS to be set up before running tests, as that can take a long time if there are too many entries.
- Retry smoke tests if it fails; it appears that some clusters can fail the first run, for unknown reasons.
- Don't try to get logs if everything succeeded; that can take ~20 minutes (due to the size), which is better spent running tests for the next PR.

## Motivation and Context
Part of #1144 in trying to move our CI to GitHub Actions.

This PR is **not** sufficient for parity; we will still need to look at:
- [ ] brain tests (this blocks closing #1144)
- [ ] upgrade tests (#1147)
- [ ] internetless (#1148)
- [ ] ccdb rotation

## How Has This Been Tested?
Been running on my GitHub fork for a few days (manually triggered); it seems to have gotten to a point where most runs succeed now.  The last test run (of the whole thing, complete with internal PR) was https://github.com/mook-as/kubecf/actions/runs/369237207.

## Types of changes

## Checklist:

  - Need to ensure that people don't run CI until the code has been reviewed.

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -1,7 +1,7 @@
 name: Pull Request Tests
 on:
   # Trigger the workflow when PRs are labeled
-  pull_request_target:
+  pull_request:
     branches: [master]
     types: [labeled]
 
@@ -40,6 +40,7 @@ jobs:
           path: kubecf
           fetch-depth: 0
           submodules: recursive
+          persist-credentials: false
 
       # Python setup
       - uses: actions/setup-python@v2
@@ -109,6 +110,7 @@ jobs:
         with:
           path: kubecf
           submodules: recursive
+          persist-credentials: false
         timeout-minutes: 1
 
       # We need catapult to create Kubernetes clusters
@@ -117,6 +119,7 @@ jobs:
         with:
           repository: SUSE/catapult
           path: catapult
+          persist-credentials: false
         timeout-minutes: 1
 
       # Python setup
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -335,7 +335,8 @@ jobs:
 
       # Fetch logs for debugging
       - name: Fetch logs
-        if: always()
+        id: fetch-logs
+        if: failure()
         run: |
           gcloud auth activate-service-account --key-file=${GKE_CRED_JSON}
           # Running klog.sh twice will grab logs from both namespaces
@@ -348,7 +349,7 @@ jobs:
       - name: Upload logs
         # Even if the fetch logs step failed, we may have _some_ logs from the
         # operator part; try to upload anyway.
-        if: always()
+        if: always() && steps.fetch-logs.outcome != 'skipped'
         uses: actions/upload-artifact@v2
         with:
           name: ${{ matrix.backend }}-klog.tgz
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request-ci.yaml
```diff
@@ -196,7 +196,7 @@ jobs:
           DEBUG_MODE: true
           GKE_NODE_COUNT: 1
           GKE_PREEMPTIBLE: false
-          GKE_INSTANCE_TYPE: n2d-highcpu-16
+          GKE_INSTANCE_TYPE: n2d-highcpu-32
```
viccuad, line 202: Note that on Concourse right now we are using 16, we may want to hold this change:

https://github.com/cloudfoundry-incubator/kubecf/blob/1986be55ff3957b94373eb25d934a9c2c80665d9/.concourse/pipeline.yaml.gomplate#L186

.github/workflows/pull-request-ci.yaml
```diff
@@ -284,6 +284,13 @@ jobs:
 
       # Run smoke tests
       - run: make smoke-tests
+        id: smoke-tests
+        timeout-minutes: 10
+        continue-on-error: true
+
+      - name: Re-rune smoke tests
```
viccuad, line 294: I'm not completely convinced of this, but feel free to mark as resolved. If `make kubecf-wait` is not enough, we should either fix that or kubecf. Just out of curiosity, how many times have you seen the need for re-running smokes on the runs?

viccuad (dismissed):
Yay! Amazing PR with so many good changes! Approving.
I would like to move the scripts inside the workflow into kubecf/.github or kubecf/scripts, but that can wait.

.github/workflows/pull-request-ci.yaml
mook-as: Yep; at a minimum, we need more than 16 gigs of RAM (or rather, at least switch to `n2d`, which gives a full 16G rather than `n1` which gives slightly less), due purely due to our requests (it wasn't enough to start CATS).

I only bumped it to 32 because riding _that_ close to the edge seemed bad.

I'm happy to drop the change here to just `n2d`, if that helps; they're all in separate commits.

.github/workflows/pull-request-ci.yaml
mook-as: I'm not convinced either, but then I noticed similar failures in concourse runs.  Of the last 5 runs (10 clusters, because diego + eirini), it might have fixed one of those?

I'm okay with dropping it.

mook-as:
Rebased to drop the two changes mentioned, we can do them separately if desired.

andreas-kupries (approved):
LGTM. And the `Re-rune` nit is gone with the dropped commits.

PR #1605, opened 2020-11-19
fix: don't define Eirini cluster role when deploying Diego

### LINKED ISSUE #1582: Eirini cluster role is created also with Diego
**Describe the bug**
https://github.com/cloudfoundry-incubator/kubecf/blob/master/mixins/eirini/templates/eirini-cluster-role.yaml#L1 is applied with both Diego and Eirini.

**To Reproduce**
Install KubeCF with Diego, the Eirini cluster role binding is present. Try to install another deployment of kubeCF in another namespace, with the same operator, fails with:
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRoleBinding "eirini-cluster-rolebinding" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "foo": current value is "kubecf"
```

**Expected behavior**
Eirini cluster role binding shouldn't be present on Diego installation

**Environment**
 - CF Version
 - Component/Plugin Name
 - Others

**Additional context**
Prevents me to deploy multiple KubeCF in one cluster

jandubois:
There is no need to define the Eirini cluster role and role binding when we are not using them.

Also prefixes the cluster roles and rolebindings with the release namespace name to allow installation of multiple instances within the same cluster.

Fixes #1582

### DIFF
--- a/mixins/eirini/templates/eirini-cluster-role.yaml
+++ b/mixins/eirini/templates/eirini-cluster-role.yaml
@@ -1,10 +1,12 @@
+{{- include "_config.load" $ }}
+{{- if .Values.features.eirini.enabled }}
 ---
 # This cluster role should not be required
 # https://github.com/cloudfoundry-incubator/eirini/issues/110
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: eirini-cluster-role
+  name: "{{ .Release.Namespace }}-eirini-cluster-role"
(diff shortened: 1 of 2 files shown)


mook-as (changes requested):
👍 but please fix the commit message (… when _not_ deploying Eirini)

mook-as (approved):

PR #1609, opened 2020-11-19

cfcontainerizationbot:

### DIFF
--- a/chart/templates/apps_dns.yaml
+++ b/chart/templates/apps_dns.yaml
@@ -136,7 +136,7 @@ spec:
         command: [/resolvwriter]
         args:
         - --upstream-dns-host
-        - {{ printf "bosh-dns.%s.svc" .Release.Namespace | quote }}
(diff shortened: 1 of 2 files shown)


mook-as:
v7 has a breaking change re: name of the `bosh-dns` service; we'll need to fix it in [apps-dns](https://github.com/cloudfoundry-incubator/kubecf/blob/e42dd0b38af6c88aa7c5d0ca9c939e87c62affe4/chart/templates/apps_dns.yaml#L139).

jandubois:
> we'll need to fix it

Are you going to take care of it @mook-as ? By pushing another commit to this PR?

mook-as:
Not immediately (I just happened to read the quarks-operator PR and thought that would be important here); if I _do_ end up looking at this later, I'll make a commit.  But if anybody else wants to pick it up that's fine too :)

jandubois (dismissed):

viovanov (dismissed):

jandubois (dismissed):

jandubois (approved):

PR #1610, opened 2020-11-20
feat: CI: GHA: brains, internetless, and CCDB rotation

mook-as:
## Description
This adds some of the missing parts of the existing concourse-based pipeline to GitHub Actions:
- Runs the brains tests
- Runs the CCDB rotation tests
- Runs the internetless tests (#1148)

## Motivation and Context
This should be the last part of #1144

## How Has This Been Tested?
Deployed & ran on my fork on KubeCF.

## Types of changes

## Checklist:

## Additional information:
- The network policy stuff for internetless tests is from catapult.
- So is the technique for patching the runtime properties for the CATS QuarksJob to run just the internetless suite.
  - I changed it a bit so the property patching is separate from what properties to patch; we could perhaps later change the Eirini config generation to use normal + a step to patch the test suites (so that it is more visible).
- I'm splitting out the experimental flags to separate flags, so that it is easier to adjust them individually if needed.

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches: [master]
     types: [labeled]
+  workflow_dispatch: {}
 
 env:
   GIT_PAGER: cat
@@ -14,13 +15,16 @@ jobs:
   # If it doesn't have it, the flow stops. If it does, we remove it (approval is good for one run only).
   dequeue:
     name: Dequeue
-    if: contains(github.event.pull_request.labels.*.name, 'pr-test-trigger')
+    if: >
+      github.event_name == 'workflow_dispatch'
+      || contains(github.event.pull_request.labels.*.name, 'pr-test-trigger')
     runs-on: ubuntu-20.04
     steps:
       - uses: actions-ecosystem/action-remove-labels@556e306
         with:
           github_token: ${{ secrets.github_token }}
           labels: pr-test-trigger
+        if: github.event_name == 'pull_request'
 
   build:
     name: Build
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -291,7 +291,8 @@ jobs:
         timeout-minutes: 10
 
       # Run brain tests
-      # - run: make brain-tests
+      - run: make brain-tests
+        timeout-minutes: 25
 
       # Run sits tests
       - run: make sync-integration-tests
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -103,8 +103,10 @@ jobs:
       matrix:
         include:
           - backend: diego
+            internetless: true
             ccdb-rotate: true
           - backend: eirini
+            internetless: false
             ccdb-rotate: false
       fail-fast: false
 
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request-ci.yaml
```diff
@@ -306,6 +306,27 @@ jobs:
...
+        run: |
+          source ${GITHUB_WORKSPACE}/catapult/build${CLUSTER_NAME}/.envrc
+          testing/ccdb_key_rotation/rotate-ccdb-keys-test.sh
+
+          echo "::group::Wait for KubeCF to be ready again"
+          make kubecf-wait
+          echo "::endgroup::"
+        env:
+          CHART: ${{ github.workspace }}/kubecf-bundle/kubecf_release.tgz
+        continue-on-error: ${{ !matrix.ccdb-rotate }}
```
viccuad, line 360: From what I can read, `continue-on-error` marks the step as green, even if it fails, which is dangerous.
I wonder if this line is needed if `fail-fast: false` is set.

Perhaps we want to set `fail-fast: true` and:
```
run: foo
if: experimental == true
```

But I may be misunderstanding things.
(same for the other instances of `continue-on-error` and the re-do of the experimental jobs in the PR).

viccuad (changes requested):
Nice! and so fast :).

I have one concern, which is the new _pull-request-ci/isolate-network.sh_.

I wonder if it wouldn't be better to use Catapult's `make airgap-up` and `make airgap-down`, which is implemented for some k8s providers, and errors for those not implemented yet. Given that some providers need special settings (kind, minikube) or may not have support for networking plugins (CaaSP), so they need host/ssh access to change iptables, plugins, etc.
The Catapult targets are also used for airgap installs, hence why the policy includes the default ns too, as the docker registry is there when performing airgapped installs. We could simplify this.

If we don't call Catapult's `make airgap-{up,down}`, perhaps we should:
- Add another policy to airgap `$QUARKS_NAMESPACE`, and possibly `default` ns.
- Add a simple test to check that it is airgapped, since we could be missing the networking plugin for example.
- Add it as a kubecf make target, in _kubecf/scripts/_. It concerns me that we are duplicating efforts in GHA scripts and _kubecf/scripts_ as the first are not easily consumed downstream. While I expect _pull-request-ci/override-cats-properties.sh_ is to be moved in the future to _kubecf/scripts_ to comply with an rfd-9/10, _pull-request-ci/isolate-network.sh_ doesn't need to be, so why not have it directly in _kubecf/scripts/_.

mook-as (commented):
Ah, I didn't realize `make airgap-{up,down}` existed, because I was looking at the catapult code which [didn't use it](https://github.com/SUSE/catapult/blob/a1f1ad3941a64fcd91824b6eff7471ced24998a0/modules/tests/kubecf-test.sh#L61).

Hmm, actually, looking at the implementation, `make airgap-up` is [only implemented for CaaSP](https://github.com/SUSE/catapult/blob/a1f1ad3941a64fcd91824b6eff7471ced24998a0/modules/experimental/airgap_up.sh#L7), so that wouldn't even work.

viccuad:
> Ah, I didn't realize `make airgap-{up,down}` existed, because I was looking at the catapult code which [didn't use it](https://github.com/SUSE/catapult/blob/a1f1ad3941a64fcd91824b6eff7471ced24998a0/modules/tests/kubecf-test.sh#L61).
> 
> Hmm, actually, looking at the implementation, `make airgap-up` is [only implemented for CaaSP](https://github.com/SUSE/catapult/blob/a1f1ad3941a64fcd91824b6eff7471ced24998a0/modules/experimental/airgap_up.sh#L7), so that wouldn't even work.

D'oh, true, the code for `make airgap-{up,down}` is on https://github.com/SUSE/catapult/pull/317.
(note that what you pointed to is `make module-experimental-airgap-up`, and old implementation which is backend dependent but is not in the backends module…).

I would argue that the point still stands, though.

.github/workflows/pull-request-ci.yaml
viccuad: I see, thanks for the detailed explanation :).
It's just that `continue-on-error` always marks the step as green on UI, so we need to aware of that.

viccuad (dismissed):
Approving the PR as we talked on Slack. We can change the internetless part and consume catapult's `make airgap-up` if and once that lands :).

viccuad (approved):

PR #1611, opened 2020-11-20
feat: CI: GHA: Add upgrade tests

viccuad:
## Description

This PR adds an `upgrade-tests` job, analogous to the `tests` job. In it we test, for both Diego and Eirini:

0. Install of KubeCF General Availability (latest public release)
1. Upgrade from latest GA minor version (`kubecf-bundle-ga`) to release under test (`kubecf-bundle`).
2. Upgrade from release under test (`kubecf-bundle`) to release under test (`kubecf-bundle`).
3. Upgrade from release under test (`kubecf-bundle`) to future, non-existent yet, release (`kubecf-bundle-next`).

Notes:
- We aren't testing "Upgrade from latest GA major version to release under test". Currently, this is the same as "1." as we started supporting upgrades between major versions with 2.X, so we don't need to test upgrades from 1.X to 2.X, and we have not bumped the major version yet to 3.x .
- There is repetition between `upgrade-tests`, `build`, and `tests` job. For the GHA boilerplate code this may be ok. I have made an effort in deduplicating the shellscripts that are around, and moving them to either catapult or scripts/. GHA doesn't support yaml anchors yet nor a way to not repeat steps. If we want to remove _all_ repetition, we should run the yaml through templating, but then we should have lint and tests for it.
- Creation of `kubecf-bundle-next` is by doing a bersion bump, no change to k8s objects, DB schema, etc. This should improve in the future.
- We are not enabling yet either HA (needs a bump on instance type) nor Autoscaler (needs https://github.com/cloudfoundry-incubator/kubecf/pull/1632).

## Motivation and Context

Closes https://github.com/cloudfoundry-incubator/kubecf/issues/1147.

## How Has This Been Tested?

Tested in runs at: https://github.com/viccuad/kubecf/pull/2
See: https://github.com/viccuad/kubecf/actions?query=workflow%3A%22Pull+Request+Tests%22

## Types of changes

## Checklist:


viccuad:
@mook-as requested your review even if it is a draft, because of the repetition problem. What do we want to do?
I vote for repeating GHA boilerplate code, and move scripts out of `run` actions; at the end we should have only make targets in `run` actions for it to be acceptable.

mook-as (commented):
Looking at the concourse pipelines now, I'm not even sure anchors were a great fit either.  Since GitHub Actions doesn't support that anyway, I agree that extracting scripts out and leaving the rest in is probably the best.

I _could_ see, as a future step, doing templating… probably with a lint-style check on PRs to make sure re-running the templating has no differences (i.e. that the checked-in YAML files are not out of sync from the template).  Not sure I'd want `ytt` for this though; maybe after we've done #1422?

viccuad:
I will go with extracting scripts then.
For the record, I _really_ don't like templating it, and if we do it, yes, we should have lintings or tests for it. I'm on the opinion that the pipelines should have 0 logic, which should go to catapult or kubecf/scripts. Extracting scripts makes me more than happy.

viccuad:
Rebased, reworded PR comment, ready for review :).

viccuad:
Sigh, it seems that diego upgrades failed on a smoke post upgrade, and eirini failed on deployment.
For a successfull diego upgrades run, one can see: https://github.com/viccuad/kubecf/runs/1463690679?check_suite_focus=true

scripts/cf-operator-wait.sh
```diff
@@ -19,13 +19,11 @@ for crd in "${crds[@]}" ; do
...
-deployments=(
-    cf-operator-quarks
-    cf-operator-quarks-job
-    cf-operator-quarks-secret
-    cf-operator-quarks-statefulset
-)
-for deployment in "${deployments[@]}" ; do
+deployments=$(kubectl get deployments --namespace cf-operator \
+                       --no-headers -o custom-columns=":metadata.name" \
+                  | tr '\r\n' ' ');
```
mook-as, line 24:
```suggestion
mapfile -t deployments < <(kubectl get deployments --namespace cf-operator --output=name)
```

And then drop the `deployment/` prefix when we use it (since that's included in `--output=name`).

scripts/cf-operator-wait.sh
```diff
@@ -19,13 +19,11 @@ for crd in "${crds[@]}" ; do
...
+
+for deployment in ${deployments} ; do
     RETRIES=60 DELAY=5 retry kubectl wait --for=condition=Available \
         --namespace=cf-operator --timeout=600s "deployment/${deployment}"
```
mook-as, line 28:
```suggestion
for deployment in "${deployments[@]}" ; do
    RETRIES=60 DELAY=5 retry kubectl wait --for=condition=Available \
        --namespace=cf-operator --timeout=600s "${deployment}"
```

scripts/kubecf-upgrade.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+source scripts/include/setup.sh
+
+require_tools helm kubectl
+
+helm upgrade kubecf "${CHART}" \
+     --namespace "${KUBECF_NS}" --values "${VALUES}" "$@"
```
mook-as, line 7: Doesn't `kubecf-apply.sh` already do `helm upgrade --install`? Why do we need a new script?

scripts/cf-operator-upgrade.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+source scripts/include/setup.sh
+
+require_tools cf_operator_url kubectl helm
+
+if [ -z "${CHART:-}" ]; then
+    CHART="$(cf_operator_url)"
+fi
+
+helm upgrade cf-operator \
```
mook-as, line 10: Can we not change `cf-operator-apply.sh` to do `helm upgrade --install` instead of having a new script?

scripts/kubecf-wait.sh
```diff
@@ -63,3 +63,25 @@ mapfile -t endpoints < <(get_resource endpoints)
...
     RETRIES=180 DELAY=10 retry wait_for_endpoint "${endpoint}"
 done
+
+green "Waiting for all services to have associated external IPs"
+wait_for_service_external_ips() {
+    selector=(
+        'app.kubernetes.io/component=router'
+        '!quarks.cloudfoundry.org/instance-group-name'
+    )
+    jsonpath='{.items[].status.loadBalancer.ingress[].ip}'
```
mook-as, line 73: Wouldn't this only make sense for LB configurations? (It wouldn't make sense for minikube, for example.)

mook-as (changes requested):
I wonder if we could not have a completely independent pipeline… but I definitely agree that templating is _not_ the way to go, just from our experiences using it in concourse.

.github/workflows/pull-request-schedule.yaml
```diff
@@ -6,7 +6,7 @@ on:
 
 env:
   BACKEND: gke
-  CLUSTERS_PER_PR: "2"
+  CLUSTERS_PER_PR: "4"
```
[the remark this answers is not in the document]
mook-as: Ah, I misread, there's also eirini-upgrades.

.github/workflows/pull-request-ci.yaml
```diff
@@ -428,13 +402,454 @@ jobs:
...
+        id: latest_version
+        uses: abatilo/release-info-action@v1.3.0
+        with:
+          owner: cloudfoundry-incubator
+          repo: kubecf
+
+      # Download latest General Availability chart bundle
+      - id: download-bundle-ga
+        name: Download bundle-ga
+        uses: dsaltares/fetch-gh-release-asset@master
```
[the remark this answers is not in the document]
viccuad: Well, this is embarrassing. Pinned in https://github.com/cloudfoundry-incubator/kubecf/pull/1611/commits/d22a63f15801f1e11a975705921ee12b62390097

.github/workflows/pull-request-ci.yaml
```diff
@@ -428,13 +402,454 @@ jobs:
...
+          mkdir kubecf-bundle-next
+          tar xf kubecf-bundle-next.tgz -C kubecf-bundle-next
+        working-directory: ${{ github.workspace }}
+
+      # Upload generated bundle-next folder
+      - name: Upload bundle-next
+        uses: actions/upload-artifact@v2
+        with:
+          name: kubecf-bundle-next.tgz
+          path: ${{ github.workspace }}/kubecf-bundle-next/
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/cloudfoundry-incubator/kubecf/pull/1611/commits/3630d30a19ff7463aabf1e3c458de70e65010eff

.github/workflows/pull-request-ci.yaml
```diff
@@ -428,13 +402,454 @@ jobs:
...
+          repository: SUSE/catapult
+          path: catapult
+          persist-credentials: false
+        timeout-minutes: 1
+
+      # Python setup
+      - uses: actions/setup-python@v2
+        with:
+          # Force 3.8 because google-cloud-sdk is incompatible with 3.9
+          # https://issuetracker.google.com/issues/170125513
```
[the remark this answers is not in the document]
viccuad: Postponing it for now then.

.github/workflows/pull-request-ci.yaml
```diff
@@ -428,13 +402,454 @@ jobs:
...
+    env:
+      # For KubeCF
+      PINNED_TOOLS: true
+      TOOLS_DIR: ${{ github.workspace }}/tools
+      # FEATURE_AUTOSCALER: "true" # FIXME
+      # For catapult
+      # Note that credentials-related parts are in the individual steps
+      BACKEND: gke
+      DOWNLOAD_CATAPULT_DEPS: "false"
+      CLUSTER_NAME: ci-${{ github.run_id }}-${{ matrix.backend }}
```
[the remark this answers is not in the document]
viccuad: Fixed on the merge from master.

.github/workflows/pull-request-ci.yaml
```diff
@@ -428,13 +402,454 @@ jobs:
...
+      # Get resource info for debugging
+      - name: Get Resource Info
         if: always()
         run: |
           set +o errexit
-          # Remove any terraform locks, because we don't care if the previous run
-          # has finished successfully (e.g. on cancel).
-          find "build${CLUSTER_NAME}" -name '.terraform.*.lock*' -delete
-          make clean
+          gcloud auth activate-service-account --key-file=${GKE_CRED_JSON}
```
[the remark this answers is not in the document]
viccuad: Removed all of them in, and it works. See: https://github.com/cloudfoundry-incubator/kubecf/pull/1611/commits/b1dccf338617d807dd9b3baf377fa851c42f2509

I left it untouched when copying from the `tests` job, and to be honest I don't see the need for it. If the workspace is persistent, and we already used Catapult to authenticate and do `make k8s`, why can't we just do the `source .envrc` and operate with the login saved there? The same applies to all `kubectl`, `helm`, etc operations.

That way we abstract the public cloud creds (be it AWS, GKE, AKS,..) and simplify things. Which is the idea of using Catapult to create the buildfolder, its envrc, and the cluster.

scripts/cf-operator-wait.sh
viccuad: Fixed in https://github.com/cloudfoundry-incubator/kubecf/pull/1611/commits/d47783193a7553735981823540830fb0aa61d8f8
Committed it myself, sorry (I rebased master at the beginning, and later did a merge from master instead, cherry-picking my commit for this..).

scripts/cf-operator-wait.sh
viccuad: Fixed, see previous commit.

scripts/kubecf-upgrade.sh
viccuad: My intent was to move away from doing `--set` anywhere. But yes, this should be fixed in `{cf-operator,kubecf}-apply.sh` instead of replicating targets.
Reverted. See:
https://github.com/cloudfoundry-incubator/kubecf/pull/1611/commits/38af78b4963f5f70f3c842e59d3a26c18425c7db
https://github.com/cloudfoundry-incubator/kubecf/pull/1611/commits/706daa62fb6698798d1b82fee5af842d0fd90533

scripts/cf-operator-upgrade.sh
viccuad: Reverted, see previous comment

scripts/kubecf-wait.sh
viccuad: True. I gave it a try, but I suppose this check will end up just being completely skipped in the other cases for now.

viccuad: Decided to just move the wait-for-ip code into a script in pull-request-ci/, for now.

viccuad:
Ready for review.
With current master, upgrades seem to fail from GA to chart-under-test.
I have pinned charts beforehand, and it worked.
Here is the latest run (still ongoing) in my fork: https://github.com/viccuad/kubecf/pull/2/checks?check_run_id=1499733138

.github/workflows/pull-request-ci.yaml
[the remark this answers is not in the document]
mook-as: Filed https://github.com/SUSE/catapult/issues/327 to keep track on the catapult side.

.github/workflows/pull-request-ci.yaml
```diff
@@ -80,6 +80,66 @@ jobs:
           name: kubecf-bundle.tgz
           path: ${{ github.workspace }}/bundle
 
+  build-next:
```
mook-as, line 83: I would have just made `build` use a matrix instead, but sure. (We can explicitly set `VERSION` to empty string to keep the existing behaviour.)

.github/workflows/pull-request-ci.yaml
```diff
@@ -291,28 +351,7 @@ jobs:
...
-          for host in api login; do
-            fqdn="${host}.${system_domain}"
-            printf "Waiting for %s to have IP address %s..." "${fqdn}" "${ip}"
-            while [[ "$(dig +short "${fqdn}")" != "${ip}" ]]; do
-              printf "."
-              sleep 10
-            done
-            printf "Done.\n"
-          done
+        run: exec "${GITHUB_WORKSPACE}/kubecf/.github/workflows/pull-request-ci/wait-for-ip.sh"
```
mook-as, line 354: ❤️

mook-as (approved):
👍 Nothing terrible enough to block merging.

I had been [experimenting](https://github.com/mook-as/kubecf/actions/runs/392265931) (based on this branch) to use matrix to split between normal & upgrade; didn't love the results, though, so just merging this is fine.

.github/workflows/pull-request-ci.yaml
viccuad: True, that's a better approach. I take note for following PRs.

viccuad:
Retriggered CI, it passses. Merging. Thanks for the reviews!

PR #1613, opened 2020-11-20
fix: CI: Actions: Fix PR triggering

mook-as:
## Description
The automated CI runs were not quite working correctly:
- The CI run was removing the wrong label
- The CI run was proceeding even with no labels set
- The scheduler wasn't triggering correctly, because the GitHub Actions provided token cannot be used to trigger further runs.  Workaround is to use a token from the vault.

## Motivation and Context
CI triggering was busted.

## How Has This Been Tested?
Ran on my fork, manually running the scheduler and seeing that:
- `pr-test-queue` was removed
- `pr-test-trigger` was added
- The CI run actually happened
- `pr-test-trigger` was removed again.

## Types of changes

## Checklist:

## Notes

As a precaution, I'm only using the token from the vault where necessary; that's to avoid chewing up the API request quota, since the GitHub Actions token should be ephemeral (and hopefully gets a different quota; at least it would be a different user).

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -20,11 +20,12 @@ jobs:
       - uses: actions-ecosystem/action-remove-labels@556e306
         with:
           github_token: ${{ secrets.github_token }}
-          labels: pr-test-queue
+          labels: pr-test-trigger
 
   build:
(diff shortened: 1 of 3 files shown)


andreas-kupries (approved):
:+1:

PR #1615, opened 2020-11-20
fix: avoid generating ops that reference instance groups that don't exist

jandubois:
Right now these are just property settings for `api` and `cc-worker` that are set in `chart/config/api.yaml` and `chart/config/jobs.yaml`, but this commit pre-emptively handles all groups that are removed in `chart/templates/multiple_cluster_mode.yaml`.

Ideally this would all be handled by conditions defined in `chart/config/jobs.yaml`, but the required logic becomes too complex, so special-casing this issue seems warranted.

This commit also adds an op for `brain-tests`, which seems to be missing because of an oversight.

This PR should supersede #1543

### DIFF
--- a/chart/templates/_config.tpl
+++ b/chart/templates/_config.tpl
@@ -72,6 +72,7 @@
     {{- include "_jobs.update" . }}
     {{- include "_resources.update" . }}
     {{- include "_database.update" . }}{{/* database/_database.tpl */}}
+    {{- include "_multicluster.update" . }}
 
     {{- range $condition, $message := $.Values.unsupported }}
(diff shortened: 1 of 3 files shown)


mook-as (approved):
I keep wondering where you're removing the old code until I realize this _isn't_ moving the multi cluster mode stuff into the new system :D

We probably want to do that at some point?

jandubois:
> We probably want to do that at some point?

Maybe; I was hinting at that with

> Ideally this would all be handled by conditions defined in chart/config/jobs.yaml, but the required logic becomes too complex, so special-casing this issue seems warranted.

PR #1625, opened 2020-11-25
feat: CI: change instance type to standard-8

mook-as:
## Description
This gives us ~double the memory, but half of the CPU (at a bit over half the price).  We appear to be memory-constrained; we're getting kubelet-level OOMs.

This bumps the memory on the cluster to 32GB (instead of the previous 16GB); to control costs, I have also reduced the CPU from 16 to 8.

## Motivation and Context
🐱 runs were failing on GitHub Actions; hijacking a running test cluster show that the node had events that looked suspicious:
> `System OOM encountered, victim process: log-cache, pid: 123378`

## How Has This Been Tested?
Ran more 🐱 tests on my fork.  Double checked to make sure the duration of the steps were not too far off.  The only major difference is `make kubecf-wait` now takes ~18 minutes (out of a limit of 30).  In general though this _is_ slower: previously the pipeline took ~90 minutes, it now takes ~120.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -206,7 +206,7 @@ jobs:
           DEBUG_MODE: true
           GKE_NODE_COUNT: 1
           GKE_PREEMPTIBLE: false
-          GKE_INSTANCE_TYPE: n2d-highcpu-16
+          GKE_INSTANCE_TYPE: n2d-standard-8
           EXTRA_LABELS: '{"ci": "${{ github.run_id }}"}'
         timeout-minutes: 20 # This should take ~10 minutes.
 


viccuad (approved):
I think this is ok.
WRT to upgrade tests, it takes around 1h 30 mins to perform the 3 consecutive upgrades, and given that they are also needed for merging and are run in parallel with normal tests, sounds good. The kubecf-wait should not have much effect (looks like an increase of ~3 mins).

viccuad:
I suppose the standard-8 instance may not be enough for HA tests, though.

mook-as:
Yeah, this won't be enough for HA testing (we'd probably need more nodes per cluster for that); but then none of this was ever sized for HA anyway.

PR #1630, opened 2020-11-26
fix: CI: Add owner name to cluster name

mook-as:
## Description
This attaches the (sanitized) owner name to the cluster name; this does _not_ remove the existing owner label on the cluster.

## Motivation and Context
This is necessary because we don't get any information other than the cluster name with looking through DNS record sets; so we need something there to be able to identify entries that are safe to remove.

## How Has This Been Tested?
Deployed locally a few times (via GitHub Actions, of course).

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -80,10 +80,24 @@ jobs:
           name: kubecf-bundle.tgz
           path: ${{ github.workspace }}/bundle
 
+  build-info:
+    name: Get Build Information
+    runs-on: ubuntu-latest
+    outputs:
+      owner: ${{ steps.get-owner-name.outputs.owner }}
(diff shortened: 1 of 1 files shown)


viccuad (approved):
Approving.
I think that setting the github.repository_owner on the cluster name, and making GKE_CLUSTER_NAME be the same as CLUSTER_NAME is the correct thing to do, so this should go in.

But I'm not convinced that we should rely on CLUSTER_NAME to delete DNS entries, to be fair.

mook-as:
I would also prefer not relying on the cluster name for DNS; however, as far as I can tell we don't have any other way to tie a DNS entry to an already-deleted cluster.  It should (but doesn't) have any way to attach tags to the record.

jandubois (approved):

PR #1638, opened 2020-12-01
fix: remove sample_values.yaml

### LINKED ISSUE #1626: Get rid of sample_values.yaml
Now that all "hidden" values have been moved to internal config files, I don't think the reasoning from #711 still applies.

The `sample_values.yaml` is now redundant and anyways hard to discover, so I suggest we stop including it.

jandubois:
Fixes #1626


mook-as (commented):
👍

mook-as (approved):

PR #1648, opened 2020-12-07
fix: GHA: CI: Checkout PRs code on CI jobs and fix lint errors on wait-for-ip.sh and 

viccuad:
## Description

Fix shellcheck lint errors in `pull-request-ci/wait-for-ip.sh `.
These errors where not detected by the PR tests:
  https://github.com/cloudfoundry-incubator/kubecf/runs/1499748939
  https://concourse.suse.dev/builds/42248
nor the [master run](https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/lint-master/builds/234). Both of them just call `make lint`.
They are detected by [GH Actions runs](https://github.com/viccuad/kubecf/runs/1511161957?check_suite_focus=true) on my fork, though. This is a bit puzzling.

In the process of fixing it, I realized that I couldn't get GH Actions to consume the fix, as it always checks outs Master from what it looks like.

There seems to not be any problem on checking out PR's branch on the pull-request-ci.yaml workflow, hence https://github.com/cloudfoundry-incubator/kubecf/commit/edec61515d71544d79f0ea5f0487e48d63b6ec3d. Note that I didn't change the lint workflow.

Note that because of `pull_request_target`, the changes in behavior for GHA may not get applied in the test of this PR.
https://github.com/cloudfoundry-incubator/kubecf/blob/4dd38e7123a0a04d4f6f1129b0257d49546b7d48/.github/workflows/pull-request-lint.yaml#L8 
`pull_request_target`  (instead of `pull_request`) makes GHA check out master, and not the branch code. This is needed for security reasons, as lint tests are run without approval for all PRs, hence they need to run from code already in master.

## Motivation and Context

I want to iterate on GHA, but I can't because of the linting errors. And since GHA is checking out master instead of HEAD, I can't fix it in the same branch.

## How Has This Been Tested?

Here: https://github.com/viccuad/kubecf/runs/1511162182?check_suite_focus=true
I used `pull_request` there and no `pull_request_target`. But that only meant that it run 2 lint jobs, one for master and one for the PR's branch, so I couldn't continue past linting even if 1 of those jobs succeeded.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/pull-request-ci.yaml
+++ b/.github/workflows/pull-request-ci.yaml
@@ -42,6 +42,7 @@ jobs:
       - name: Checkout KubeCF
         uses: actions/checkout@v2
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
           path: kubecf
           fetch-depth: 0
           submodules: recursive
--- a/.github/workflows/pull-request-ci/wait-for-ip.sh
+++ b/.github/workflows/pull-request-ci/wait-for-ip.sh
@@ -11,8 +11,8 @@ selector=(
     )
 jsonpath='{.items[].status.loadBalancer.ingress[].ip}'
 filter=(
-      --namespace=kubecf
-      --selector="$(IFS=, ; echo "${selector[*]}")"
+      '--namespace=kubecf'
+      '--selector='"$(IFS=, ; echo "${selector[*]}")"
     )
 ip="$(kubectl get service "${filter[@]}" --output="jsonpath=${jsonpath}")"
 system_domain="$(gomplate --context ".=${GITHUB_WORKSPACE}/kubecf-values.yaml" --in '{{ .system_domain }}')"


viovanov (approved):

viccuad:
Linting on Concourse passes: https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/lint-pr/builds/367
Linting on GHA fails, because it checks out the master branch when linting, for security purposes: https://github.com/cloudfoundry-incubator/kubecf/pull/1648/checks?check_run_id=1511283719

.github/workflows/pull-request-ci.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
       - name: Checkout KubeCF
         uses: actions/checkout@v2
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
mook-as, line 45: I believe this is unnecessary; when triggered by the PR, it already defaults to the head of the PR.  And when manually triggered, this will just error out.  (This is why this job is triggered on labelling — so that it will have a correct `HEAD`; if we just did it in the scheduler it wouldn't be associated with the correct branch.)

The problem is that the lint job isn't using the correct commit, not the CI job.

mook-as (changes requested):
The lint fixes look fine.

We probably need to figure something else out for the lint job — I _believe_ (but this needs to be tested again!) that we can actually make the `lint` job trigger on any PR; that job doesn't require any secrets (to spin up clusters, etc.), so it should be safe to spawn from PRs.

viccuad:
Superseded by https://github.com/cloudfoundry-incubator/kubecf/pull/1652.

.github/workflows/pull-request-ci.yaml
viccuad: > And when manually triggered, this will just error out.

Well, that's definitely not good. But on my tests I didn't see that behaviour, tbh.

The intent was to always check the HEAD of the branch on CI and lint workflows. Yet later I removed ` ref: ${{ github.event.pull_request.head.sha }}`  from the lint workflow because of security reasons:
*one could open a PR changing the lint workflow to consume the GH secrets and print them, and it would get run automatically*.

PR #1654, opened 2020-12-08
fix: update cf-operator-delete logic

jandubois:
The cluster roles and role bindings are prefixed with `$CF_OPERATOR_RELEASE`.

Also, the names have changed, and there are 2 additional roles now.


mook-as (approved):
Spot checked `job`; it doesn't have a `-cluster` suffix, so this seems correct.

PR #1682, opened 2021-01-12
feat: Use cflinuxfs3 as default stack

thardeck:
to make sure the latest upstream buildpacks can be used by default.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -4,7 +4,7 @@
 # List of stacks to install; the first one will be used as the default.
 # A stack is a prebuilt root file system that supports a specific
 # operating system with a corresponding set of buildpacks.
-install_stacks: [sle15, cflinuxfs3]
+install_stacks: [cflinuxfs3, sle15]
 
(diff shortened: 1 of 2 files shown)


viccuad (approved):
I suppose this will make Eirini fail with default install_stacks. On CI, we select the `sle15` stack if deploying Eirini, so it should succeed there.

thardeck:
> I suppose this will make Eirini fail with default install_stacks. On CI, we select the `sle15` stack if deploying Eirini, so it should succeed there.
Eirini seems to have separate configuration options which I did not change.

troytop:
If the platform admin did not specify the stacks (and their order) in their values, upgrading to this version would change the default stack (IIUC). This would be an unexpected change for the end users and probably not be immediately noticed by the admin. 

Better to leave the defaults as they are and make a release not to _manually_ change the default stack in values.yaml. That way, admins know that the change is happening.

viccuad:
@troytop thank you. I missed that in the review.

@viovanov ping, to include "Manually change default stack to cflinuxfs3" in release notes.

PR #1684, opened 2021-01-13
feat: Update cf-operator dependency.

cfcontainerizationbot:


viovanov (approved):

viovanov (changes requested):
This needs another bump

viovanov (approved):

PR #1697, opened 2021-03-04
Split out locket from diego-api as it should be independant

### LINKED ISSUE #1410: Split locket into a separate pod

andrew-edgar:
Locket should be independant from diego-api.  Colocating them makes it less resilient and causes other instances to be incorrectly dependant on diego-api.

## Description
Moving the locket from diego-api to it's own pod.

## Motivation and Context
It solves issues with diego-api being only able to talk to the single locket co-located.  It also allows for better separation of the two independant processes.

Fixes #1410

## How Has This Been Tested?
Deployed in IBM hybrid testing environments.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

Can you point me to the docs I would need to update and I can add those to the PR.

### DIFF
--- a/chart/assets/operations/instance_groups/locket.yaml
+++ b/chart/assets/operations/instance_groups/locket.yaml
@@ -1,0 +1,43 @@
+# Set instance count to 2; the default inherited from diego-api is 2
+- type: replace
+  path: /instance_groups/name=locket/instances
+  value: 2
+
+# Disable tuning /proc/sys kernel parameters as locket is running on containers.
+- type: replace
+  path: /instance_groups/name=locket/jobs/name=locket/properties/set_kernel_parameters?
+  value: false
+
+# Add quarks properties for locket.
+- type: replace
+  path: /instance_groups/name=locket/jobs/name=locket/properties/quarks?
+  value:
+    ports:
+    - name: locket
+      protocol: TCP
+      internal: 8891
+    run:
+      healthcheck:
+        locket:
+          readiness:
+            exec:
+              command:
+                - "/usr/bin/nc"
+                - "-z"
+                - "127.0.0.1"
+                - "8891"
+      #         - /var/vcap/packages/cfdot/bin/cfdot
+      #         - locks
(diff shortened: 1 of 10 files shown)


chart/assets/operations/instance_groups/locket.yaml
```diff
@@ -0,0 +1,43 @@
+# Set instance count to 2; the default inherited from diego-api is 2
```
jandubois, line 1: This is not necessary; the `_jobs.move` logic will do this automatically: https://github.com/cloudfoundry-incubator/kubecf/blob/9efc4b3/chart/templates/_jobs.tpl#L178

More generally, the new instance group also needs to be hooked up to the "sizing" mechanism in https://github.com/cloudfoundry-incubator/kubecf/blob/master/chart/assets/operations/sizing.yaml

So there should also be an entry for `sizing.log_cache.instances: ~` in `values.yaml`.

Finally, I think that with this change `diego-api` is no longer needed when running with Eirini, so please update the various places where `diego-cell` is disabled and tread `diego-api` similarly. Or tell me why it is still needed even after locket has been moved out.

Maybe also add `Fixes #1410` to the PR description to make sure the issue gets closed when this PR gets merged?

jandubois (changes requested):

chart/assets/operations/instance_groups/locket.yaml
andrew-edgar: ok working on this thanks.

jandubois (approved):

PR #1698, opened 2021-03-05
fix: master into release-2.7

jandubois:
Changes are bumps to sle15 and suse-java-buildpack, as well as a bump to quarks-operator-7.2.1


satadruroy (approved):
LGTM.

PR #1699, opened 2021-03-05
Release drafter

manno:


jandubois:
Hi @manno, I want to give this a try, but I have a couple of questions about how it works:

* Does it rewrite the draft releases notes from scratch every time a PR is merged?
* So can I add labels to already merged PRs and the draft will get updated to reflect that?
* Can that be triggered by adding the label, or does it require merging another PR?
* Will the release notes body be updated from the template each time the template is updated (e.g. bumping Quarks should update not just `dependencies.yaml`, but also `.github/release-drafter.yaml`?
* How do you make manual changes to the release notes? Do you have to do them in one sitting, and then publish them, to prevent them from getting overwritten if anybody merges another PR?

I've looked at all the docs, but this is still not clear to me...

jandubois (approved):
Ok, let's give it a try; what is the worst that can happen? 😈

I do wonder though if `$RESOLVED_VERSION` in the body of the template is going to work with the tag being set in the workflow, but we'll see...

jandubois:
@manno I've merged the PR now to see how it works, but if you can, please still answer the questions I asked above...

manno:
>    Does it rewrite the draft releases notes from scratch every time a PR is merged?

The workflow is triggered for every push to master
 
>    So can I add labels to already merged PRs and the draft will get updated to reflect that?

Yes.

>   Can that be triggered by adding the label, or does it require merging another PR?

You would have to trigger the workflow for label changes on merged PRs, not sure if that's possible.

>   Will the release notes body be updated from the template each time the template is updated (e.g. bumping Quarks should update not just dependencies.yaml, but also .github/release-drafter.yaml?

The release-drafter.yaml is just the config for the run.

> How do you make manual changes to the release notes? Do you have to do them in one sitting, and then publish them, to prevent them from getting overwritten if anybody merges another PR?

Once you are ready to release you edit the draft. If another push to master triggers the workflow it might overwrite your edits.
I think you can have multiple drafts, so just change the version.

PR #1720, opened 2021-03-26
Bump cf-deployment to 16.9.0

jandubois:
Reverts the backported CAPI fixes from 1.101.0 because we now bump capi-release to 1.107.0
Remove UAA patch that has been applied upstream.
Patches CAPI to not call the `/tasks` endpoint in Eirini (it is not implemented in Eirini-1.8).
Adds patches to all CAPI jobs to unset `KUBERNETES_SERVICES_HOST`, which it uses to detect cf4k8s.


bosh/releases/pre_render_scripts/scheduler/cloud_controller_clock/jobs/patch-cc.sh
```diff
@@ -17,144 +17,36 @@ patch --verbose "${target}" <<'EOT'
...
---- app/controllers/v3/app_manifests_controller.rb
-+++ app/controllers/v3/app_manifests_controller.rb
-@@ -70,7 +70,7 @@ class AppManifestsController < ApplicationController
-   def validate_content_type!
-     if !request_content_type_is_yaml?
-       logger.error("Context-type isn't yaml: #{request.content_type}")
--      invalid_request!('Content-Type must be yaml')
-+      bad_request!('Content-Type must be yaml')
+cat <<'EOT' > /var/vcap/all-releases/jobs-src/capi/cloud_controller_ng/opi-task.patch
+diff --git lib/cloud_controller/opi/task_client.rb lib/cloud_controller/opi/task_client.rb
```
mook-as, line 28: It would be nice if we can get a git commit message or some comments describing why we want this patch. (I understand that it's because our Eirini doesn't support the `/tasks` endpoint; but the comment should describe why just removing it is safe.)

chart/assets/operations/instance_groups/cc-worker.yaml
```diff
@@ -9,8 +9,3 @@
         command: [/usr/bin/pgrep, --full, cc-worker-cloud_controller_worker]
   {{- end }}
 {{- end }}
-
-
-{{- range $path, $bytes := .Files.Glob "assets/operations/pre_render_scripts/cc-worker_*" }}
```
mook-as, line 14: Hmm, don't we have a [cc-worker patch](https://github.com/cloudfoundry-incubator/kubecf/blob/7c852467748e367b1ec7349747c3e5e4f691e001/bosh/releases/pre_render_scripts/cc-worker/cloud_controller_worker/jobs/patch_cloud_controller_worker_erb.sh) still? Why do we no longer need to add it explicitly?

chart/assets/operations/instance_groups/scheduler.yaml
```diff
@@ -88,10 +88,6 @@
 - type: remove
   path: /instance_groups/name=scheduler/jobs/name=cc_deployment_updater
 
-{{- if not .Values.features.eirini.enabled }}
```
mook-as, line 91: We still have a scheduler patch, and you're running it… I assume this means we now use the scheduler even in the Eirini case? I don't see any matching changes in this PR though?

mook-as (changes requested):
Mostly just questions because I have lost track of things.

chart/assets/operations/instance_groups/scheduler.yaml
jandubois: The code that removes the `/tasks` functionality from the `opi/task_client.rb` is for `cloud_controller_clock`, which is part of `scheduler`.

The other `scheduler` patch is removing the `KUBERNETES_SERVICE_HOST` environment variable, and it seems fine for the Eirini branch too, so I removed the conditional instead of hard-coding patch names in the scheduler template.

bosh/releases/pre_render_scripts/scheduler/cloud_controller_clock/jobs/patch-cc.sh
jandubois: I've added comments to the commit message. Will force-push amended commits.

chart/assets/operations/instance_groups/cc-worker.yaml
jandubois: Thanks for catching that; we do indeed still need this. I've amended the "revert" commit to except this change from the revert.

mook-as: Odd that we still have a single line change here, but that's good enough :)

mook-as (approved):
LGTM

PR #1731, opened 2021-05-25
feat: update cf-operator dependency.

cfcontainerizationbot:
Increment quarks-operator version in def.bzl file.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -103,9 +103,9 @@ binaries:
 # Additional files we need to download; see http_file() invocation in WORKSPACE.
 external_files:
   cf_operator:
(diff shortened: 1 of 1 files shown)


jandubois (approved):

jandubois (approved):

jandubois:
Looks like Github didn't process the status from re-running `deploy-eirini`: https://concourse.suse.dev/teams/main/pipelines/kubecf/jobs/deploy-eirini-pr/builds/10

Which is weird, given it shows `smoke-tests-eirini` and `cf-acceptance-tests-eirini` as passed, which is only possible when `deploy-eirini` has already passed as well.

Going to override Github and merge now.

PR #1733, opened 2021-06-02
Bump catapult to latest version

jandubois:
Relevant changes:
* use `wait_for_kubecf` is more robust and doesn't return early
* don't deploy metrics-server chart if metrics-server is already running.


jandubois:
This change has already been deployed and tested with #1731

PR #1734, opened 2021-06-02
Revert #1725 (i.e. re-instate #1697: split out locket into its own instance group)

jandubois:
Change #1697 was already merged to `master`, but I backed it out in #1725 because it looked like it might be responsible for the frequent upgrade-test failures on CI (even though it obviously passed CI itself before it was merged).

I think now that all the upgrade-test failures were due to the recently fixed catapult issues (and/or to running out of loadbalancers/firewall resources on GKE). This PR is just going to re-apply the original changes that were already merged before.

### DIFF
--- a/chart/assets/operations/instance_groups/locket.yaml
+++ b/chart/assets/operations/instance_groups/locket.yaml
@@ -1,0 +1,43 @@
+# Set instance count to 2; the default inherited from diego-api is 2
+- type: replace
+  path: /instance_groups/name=locket/instances
+  value: 2
+
+# Disable tuning /proc/sys kernel parameters as locket is running on containers.
+- type: replace
+  path: /instance_groups/name=locket/jobs/name=locket/properties/set_kernel_parameters?
+  value: false
+
+# Add quarks properties for locket.
+- type: replace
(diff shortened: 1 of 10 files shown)


chart/assets/operations/instance_groups/locket.yaml
```diff
@@ -0,0 +1,38 @@
...
+    run:
+      healthcheck:
+        locket:
+          readiness:
+            exec:
+              command:
+                - "/usr/bin/nc"
+                - "-z"
+                - "127.0.0.1"
+                - "8891"
```
mook-as, line 23: Should probably use a TCP probe instead, but I don't think that's covered under this PR.

mook-as (approved):
Looks familiar.

chart/assets/operations/instance_groups/locket.yaml
jandubois: > Should probably use a TCP probe instead

@andrew-edgar Can you take a look at this?

I'll try to get this PR through CI as-is (because it is a revert of a revert), and will merge if it passes, so any changes should be in a separate PR.

